PR #361, opened 2026-03-22
feat(ui): close resource and Helm drawers by clicking outside

SaviorVirus:
Add a transparent backdrop overlay behind the resource detail drawer and Helm release drawer so users can dismiss them by clicking any empty area, instead of having to click the same row again.

Made-with: Cursor

## Description

Currently, to close the resource detail drawer or Helm release drawer, users had to click the same row that opened it (or the X button). This PR adds a transparent backdrop overlay behind both drawers so clicking anywhere on the empty area to the left dismisses them — a more intuitive UX pattern.

Changes:
- `packages/k8s-ui/src/components/workload/ResourceDetailDrawer.tsx` — backdrop is hidden when the drawer is expanded to full-screen WorkloadView to prevent accidental dismissal
- `web/src/components/helm/HelmReleaseDrawer.tsx` — same backdrop added

## Type of change

- [ ] Documentation update

## How has this been tested?

- [x] Tested locally with a remote cluster (EKS)
- Opened a resource in the Resources view → clicked empty area to the left of the drawer → it closed
- Opened a Helm release → clicked empty area to the left → it closed
- Expanded resource drawer to full WorkloadView → confirmed backdrop is not shown (no accidental close)
- Confirmed X button and row click still work as before

## Checklist

## Related issues

N/A

### DIFF
--- a/packages/k8s-ui/src/components/workload/ResourceDetailDrawer.tsx
+++ b/packages/k8s-ui/src/components/workload/ResourceDetailDrawer.tsx
@@ -113,6 +113,19 @@ export function ResourceDetailDrawer({ resource, onClose, onNavigate, initialTab
   const headerHeight = headerHeightProp ?? 49
 
   return (
+    <>
+      {/* Backdrop — only shown in drawer mode (not expanded), closes on click */}
+      {!expanded && (
+        <div
+          className={clsx(
+            'fixed inset-0 z-30 transition-opacity duration-300',
+            isOpen ? 'opacity-100' : 'opacity-0 pointer-events-none',
+          )}
+          style={{ top: headerHeight }}
+          onClick={onClose}
+        />
+      )}
+
     <div
       className={clsx(
         'fixed right-0 bg-theme-surface border-l border-theme-border flex flex-col shadow-2xl z-40',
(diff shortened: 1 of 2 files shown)


nadaverell (commented):
Thanks for the PR! 

I think this makes sense on the topology view where clicking empty space feels like it should dismiss the drawer, but not so much on the resource and helm views. There it was intentional that users can click through different rows to switch the drawer content.

In those, we have 3 close mechanisms - X button, Escape, clicking the same row.

If we want click-outside-to-close specifically for the topology graph, that could be handled at the graph canvas level rather than with a global backdrop.

nadaverell:
Hey @SaviorVirus, friendly ping — any thoughts on scoping click-outside to the topology canvas?

nadaverell:
Thanks for this, @SaviorVirus, and sorry it sat so long.

We actually went the other way on this one. The drawer is intentionally non-modal — while it's open you can click other rows in the list to switch between resources without closing it. A backdrop overlay would intercept those clicks, so click-outside-to-close trades away the quick row-to-row switching. It's a real tradeoff and we're leaning toward keeping the switching.

The drawer's also been refactored a fair bit since, so this would need a rebase regardless. Going to close it — appreciate the contribution either way. 🙏

PR #456, opened 2026-04-11
Fix pod terminal default shell (fixes #452)

### LINKED ISSUE #452: Change the default terminal when attaching to a pod
## Problem statement

When attaching to a pod terminal it appears the default is a very basic shell.  This seems to be a very limiting feature that does not support up arrow, tab auto completion, display of current pwd (working directory) upon attachment.

## Proposed solution

Can the pod terminal attachment be updated to support something more bash driven that has enhanced functionality like tab auto completion, up arrow, display of current pwd (working directory)?

## Alternatives considered

Comparing to Lens or Headlamp they use a slightly more feature full shell attachment program.

## Additional context

Debugging live pods and performing commands is slightly inconvenient.

nadaverell:
## Summary

Pod exec was hardcoded to `/bin/sh`, which on most minimal images is BusyBox `ash` or `dash` — neither ships readline, so users saw no up-arrow history, no tab completion, and no working-directory in the prompt. Bash was typically present on the same images but never tried. Closes #452.

The default is now a nested `sh -c` script that detects the best available shell via `command -v` and then `exec`s into it:

```sh
export TERM=xterm-256color
if command -v bash >/dev/null 2>&1; then exec bash -il
elif command -v ash >/dev/null 2>&1; then exec ash
else exec sh
fi
```

- **`bash -il`** — interactive login so `/etc/bash.bashrc` and `/etc/profile` get sourced, which is how images pick up a `\w`-containing `PS1` (the third symptom from the issue).
- **Detection before exec** — a naive `exec bash || exec ash` cascade doesn't work: POSIX requires non-interactive shells to exit immediately when `exec` fails to find the target, so the first missing shell kills the session before the fallback runs. `command -v` confirms existence first.
- **`TERM=xterm-256color`** — exported up front so colours and cursor movement work in images that don't set `TERM` themselves.

Also adds a `--pod-shell-default` CLI flag as a power-user escape hatch for cases where someone wants a different default (e.g. `zsh`). Wired through `internal/app/bootstrap` alongside the existing `--disable-exec` pattern. Normal users never need to touch it.

The existing `?shell=<path>` query-param override is preserved for the explicit case. Pods with no shell at all still surface the existing `shell_not_found` error path unchanged.

## What's not in scope

- **Windows pod support** (PowerShell fallback) — Radar doesn't detect pod OS today; separate follow-up.
- **UI shell picker** — neither Lens nor Headlamp expose one; the CLI flag leaves a clean seam if demand materialises.

## Verification

Unit tests: table-driven `TestDefaultExecCommand` covers the override × fallback matrix, plus a tripwire `TestDefaultShellScript` documenting the exact script content so future edits can't silently drift.

Live-tested against a real EKS cluster across five distinct scenarios:

| Scenario | Image | Expected | `$0` | `TERM` | Prompt rendered |
|---|---|---|---|---|---|
| Bash golden path | `debian:bookworm-slim` | `bash -il` | `bash` | `xterm-256color` | `root@test-bash-452:/#` |
| Ash fallback | `nginx:alpine` | `ash` | `ash` | `xterm-256color` | `/ #` |
| sh-only `else` branch | debian+bash removed | `sh` (dash) | `sh` | `xterm-256color` | `#` |
| `--pod-shell-default` override | `nginx:alpine` | user banner + shell | — | `xterm-256color` | `RADAR_OVERRIDE_HIT_72` + `/ #` |
| Multi-container switch | podinfo (alpine) + istio-proxy (ubuntu) | per-container | `ash` → `bash` | `xterm-256color` | `~ $` → `istio-proxy@…:/$` |

An earlier draft of this fix used `exec bash \|\| exec ash \|\| exec sh`; the live alpine test immediately caught that this dies with exit 127 on the first missing shell (POSIX exec-failure-exits-non-interactive-shell). The `command -v`-first form above is what the tests and live runs now validate.

### DIFF
--- a/cmd/desktop/main.go
+++ b/cmd/desktop/main.go
@@ -39,6 +39,7 @@ func main() {
 	fakeInCluster := flag.Bool("fake-in-cluster", false, "Simulate in-cluster mode for testing")
 	disableHelmWrite := flag.Bool("disable-helm-write", false, "Simulate restricted Helm permissions")
 	disableExec := flag.Bool("disable-exec", false, "Simulate restricted exec permissions")
+	podShellDefault := flag.String("pod-shell-default", "", "Override the default pod exec shell command (runs as 'sh -c <value>'; empty = built-in bash -il → ash → sh cascade)")
 	timelineStorage := flag.String("timeline-storage", fileCfg.TimelineStorageOr("memory"), "Timeline storage backend: memory or sqlite")
 	timelineDBPath := flag.String("timeline-db", fileCfg.TimelineDBPath, "Path to timeline database file (default: ~/.radar/timeline.db)")
 	prometheusURL := flag.String("prometheus-url", fileCfg.PrometheusURL, "Manual Prometheus/VictoriaMetrics URL (skips auto-discovery)")
@@ -83,6 +84,7 @@ func main() {
 		FakeInCluster:    *fakeInCluster,
 		DisableHelmWrite: *disableHelmWrite,
 		DisableExec:      *disableExec,
+		PodShellDefault:  *podShellDefault,
 		TimelineStorage:  *timelineStorage,
 		TimelineDBPath:   *timelineDBPath,
 		PrometheusURL:    *prometheusURL,
--- a/cmd/explorer/main.go
+++ b/cmd/explorer/main.go
@@ -44,6 +44,7 @@ func main() {
 	disableHelmWrite := flag.Bool("disable-helm-write", false, "Simulate restricted Helm permissions (disables install/upgrade/rollback/uninstall)")
 	disableExec := flag.Bool("disable-exec", false, "Simulate restricted exec permissions (disables terminal, debug shell)")
 	disableLocalTerminal := flag.Bool("disable-local-terminal", false, "Disable local terminal feature")
+	podShellDefault := flag.String("pod-shell-default", "", "Override the default pod exec shell command (runs as 'sh -c <value>'; empty = built-in bash -il → ash → sh cascade)")
 	// Timeline storage options
 	timelineStorage := flag.String("timeline-storage", fileCfg.TimelineStorageOr("memory"), "Timeline storage backend: memory or sqlite")
 	timelineDBPath := flag.String("timeline-db", fileCfg.TimelineDBPath, "Path to timeline database file (default: ~/.radar/timeline.db)")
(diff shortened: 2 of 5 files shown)


7adityaraj:
hello @nadaverell 

I have been observing issue with pod shell exec from version 1.5.0 and 1.5.1 but its fine 1.4.8 and lower, most probably due to this PR. What happens is -  shell try to load but it doesn't  show anything. For some pods i get warning msgs as below but not for all pods.

```
2026/04/21 17:28:54 Exec session exec-2 started (kube-system/coredns-674b8bbfcf-mlffv)
2026/04/21 17:28:54 [exec] WARNING: error looks shell-related but isShellNotFoundError did not match — kubelet error text may have drifted; update isShellNotFoundError patterns if this recurs. errMsg="command terminated with exit code 127"
```

```
└─ $ echo $SHELL
/bin/bash
```

nadaverell:
> hello @nadaverell
> 
> I have been observing issue with pod shell exec from version 1.5.0 and 1.5.1 but its fine 1.4.8 and lower, most probably due to this PR. What happens is - shell try to load but it doesn't show anything. For some pods i get warning msgs as below but not for all pods.
> 
> ```
> 2026/04/21 17:28:54 Exec session exec-2 started (kube-system/coredns-674b8bbfcf-mlffv)
> 2026/04/21 17:28:54 [exec] WARNING: error looks shell-related but isShellNotFoundError did not match — kubelet error text may have drifted; update isShellNotFoundError patterns if this recurs. errMsg="command terminated with exit code 127"
> ```
> 
> ```
> └─ $ echo $SHELL
> /bin/bash
> ```

thanks for reporting. will test if there's regression ASAP

nadaverell:
> hello @nadaverell
> 
> I have been observing issue with pod shell exec from version 1.5.0 and 1.5.1 but its fine 1.4.8 and lower, most probably due to this PR. What happens is - shell try to load but it doesn't show anything. For some pods i get warning msgs as below but not for all pods.
> 
> ```
> 2026/04/21 17:28:54 Exec session exec-2 started (kube-system/coredns-674b8bbfcf-mlffv)
> 2026/04/21 17:28:54 [exec] WARNING: error looks shell-related but isShellNotFoundError did not match — kubelet error text may have drifted; update isShellNotFoundError patterns if this recurs. errMsg="command terminated with exit code 127"
> ```
> 
> ```
> └─ $ echo $SHELL
> /bin/bash
> ```

tweaked error detection - shuold show a more friendly message now when shell not found

PR #490, opened 2026-04-19
chore(deps): bump dompurify override to 3.4.0

nadaverell:
## Summary

Closes Dependabot alert #50 (dompurify ADD_TAGS bypasses FORBID_TAGS via short-circuit evaluation, moderate severity).

- Bumps `dompurify` override in root `package.json`: `^3.3.2` → `3.4.0`
- Regenerates `package-lock.json`

## Why an exact pin instead of caret range

Tried `^3.4.0` first — npm's override resolver refused to upgrade past monaco-editor's pinned `3.2.7` peer constraint when using a range. Using an exact `"3.4.0"` forces the override correctly. We can relax this later if monaco bumps.

## Risk

Low. dompurify is a transitive dep via `@monaco-editor/react` → `monaco-editor`; we don't call DOMPurify directly. Monaco uses it for sanitizing hover/completion HTML, which doesn't expose the ADD_TAGS attack surface.

## Why the lockfile diff is huge

npm 11's resolver restructured ordering during regeneration. `lockfileVersion: 3` unchanged. The functional change is isolated to the dompurify node — `npm audit` reports 0 vulnerabilities after the change.

## Test plan

- [x] `npm audit` → 0 vulnerabilities
- [x] `npm run tsc` in `web/` → clean
- [x] `npm run build` in `web/` → clean
- [ ] CI passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,21 +10,21 @@
       ]
     },
     "node_modules/@emnapi/core": {
-      "version": "1.9.1",
-      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.9.1.tgz",
-      "integrity": "sha512-mukuNALVsoix/w1BJwFzwXBN/dHeejQtuVzcDsfOEsdpCumXb/E9j8w11h5S54tT1xhifGfbbSm/ICrObRb3KA==",
+      "version": "1.9.2",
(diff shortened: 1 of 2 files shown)


nadaverell:
Consolidated into #489 — combining CodeQL enablement + dompurify bump into a single chore PR.

PR #497, opened 2026-04-21
feat(auth): chunk session cookies so large OIDC identities don't exceed the 4KB cookie limit

### LINKED ISSUE #527: Add chunks for large cookies
## Problem statement

When using OIDC authentication with Keycloak (or other IdPs that include large claims in the ID token), the session cookie can exceed the 4KB browser cookie size limit. This causes the browser to silently reject the cookie, resulting in authentication failure after the OIDC callback redirect. The user successfully logs in but then sees 401 Unauthorized errors because the session cookie is not stored by the browser.

**Example from production:**
- Keycloak returns ID tokens with `sid`, `session_state`, and other claims
- Base64-encoded session payload exceeds 4KB
- Browser rejects the `radar_session` cookie
- All subsequent API requests return 401

## Proposed solution

Add automatic cookie chunking support that transparently splits large session cookies into multiple smaller cookies when the payload exceeds the safe limit (3800 bytes, leaving headroom for cookie attributes).

**Implementation approach:**

1. **Cookie creation**: When the cookie value exceeds 3800 bytes:
   - If ID token is present, drop it and retry (ID token is only needed for RP-Initiated Logout and can be omitted)
   - If still too large, split the value into chunks of ~3700 bytes each
   - Create chunk cookies named `radar_session_chunk_0`, `radar_session_chunk_1`, etc.
   - Create a meta-cookie `radar_session_chunks` containing the chunk count
   
2. **Cookie parsing**: 
   - Check for the main cookie first (for backward compatibility)
   - If not present, look for `_chunks` meta-cookie
   - Reassemble chunks in order
   - Verify HMAC signature on the reassembled value

3. **Secure flag handling**:
   - Set `Secure` flag based on `X-Forwarded-Proto` header
   - Use `__Secure-` prefix for secure cookies
   - Apply to both main cookie and chunk cookies

4. **Fallback behavior**:
   - Attempt to drop ID token before chunking (most tokens still fit without ID token)
   - Log warnings when chunking is necessary
   - Last resort: truncate the cookie (with warning) for extremely large cases

## Alternatives considered

1. **Server-side session storage with cookie ID only**
   - Store session data on server, use random ID in cookie
   - **Rejected because**: Requires shared session storage (Redis/DB) which adds operational complexity for single-replica deployments

2. **Compress cookie value before storing**
   - Use gzip/zstd compression on payload
   - **Rejected because**: Base64 + compression is unpredictable, may not reduce size enough

3. **Use `localStorage` + JWT instead of cookies**
   - Store session token in localStorage, send in Authorization header
   - **Rejected because**: Requires significant frontend changes, less secure (XSS vulnerability)

4. **Reduce claims in ID token**
   - Configure Keycloak to exclude large claims
   - **Rejected because**: Not under our control; IdP configuration varies per deployment

5. **Split session into multiple independent cookies**
   - Store username in one cookie, groups in another, etc.
   - **Rejected because**: Complex to manage, increases signature overhead

## Additional context

**Browser cookie limits:**
- Chrome: 4096 bytes per cookie (RFC 6265 requires browsers to support at least 4096)
- Firefox: 4097 bytes
- Safari: 4096 bytes
- We use 3800 bytes as safe limit to leave room for cookie name and attributes

**Current session cookie structure:**
radar_session=eyJ1IjoiYWxpY2UiLCJnIjpbImRldnMiXSwiZSI6MTc0N...base64.HMAC-signature

text

**Chunked cookie example:**
radar_session_chunks=3
radar_session_chunk_0=eyJ1IjoiYWxpY2UiLCJnIjpbImRldnMiXSwiZSI6MTc0NTI...
radar_session_chunk_1=...continuation
radar_session_chunk_2=...continuation

text

**Impacted files:**
- `pkg/auth/cookie.go` - main cookie implementation
- `pkg/auth/auth.go` - exports cookie functions
- `internal/auth/middleware.go` - authenticates requests
- `internal/auth/oidc.go` - sets cookie after OIDC login

**Testing done:**
- Unit tests for chunking logic
- Manual testing with Keycloak
- Verified cookie size with browser dev tools

pull request https://github.com/skyhook-io/radar/pull/497

jollyekb:
## Summary

When Radar runs in OIDC mode, a user's session is stored in a signed `radar_session` cookie. Identities with large OIDC tokens or many group memberships (common with Keycloak/LDAP-backed IdPs) can push that cookie past the browser's ~4 KB per-cookie limit, at which point the browser silently drops it — the user completes login but every subsequent request comes back 401.

This adds automatic cookie chunking so large sessions are split across multiple cookies and transparently reassembled, removing the 4 KB ceiling as a login failure mode.

Fixes #527.

## What changed

All session-cookie handling lives in `pkg/auth/cookie.go`. `CreateSessionCookie` now returns `[]*http.Cookie` and picks one of two representations:

- **Single cookie** — when the signed value fits under the per-cookie limit (3800 bytes). If a large OIDC ID token is the only thing pushing it over, the ID token is dropped first (it's only needed for RP-Initiated Logout and degrades gracefully to `client_id`).
- **Chunked** — when it still doesn't fit, the value is split across `radar_session_chunk_0..N` cookies plus a `radar_session_chunks` meta-cookie holding the count.

Parsing (`ParseSessionCookie`) prefers the single main cookie; otherwise it reassembles the chunks (bounded by a max chunk count) and verifies the HMAC over the **whole reassembled value**. Because the signature covers the full value, a dropped, reordered, or tampered chunk fails verification rather than yielding a partial session.

Key invariants:

- **One representation at a time.** Issuing a single cookie clears the chunk meta-cookie; issuing chunks expires the main cookie. A session can never be shadowed by a stale cookie of the other shape (e.g. an old single cookie winning over a fresh chunked one after a group change).
- **Logout clears everything.** Clearing a session expires the main cookie, the meta-cookie, and every chunk index — so a chunked session can't survive logout by being reassembled on the next request.
- **Secure handling.** Cookies are marked `Secure` based on the connection (`r.TLS` / `X-Forwarded-Proto`); the OIDC callback always marks them `Secure`. Clearing cookies mirror the `Secure` attribute so they reliably evict the originals.
- **Bounded by design.** Reassembly is capped at a fixed chunk count, so a forged meta-cookie can't drive an unbounded read. An identity too large to fit even when chunked is refused with a `413` and an actionable message instead of issuing an unusable session that would loop login → 401.

`internal/auth` (middleware, OIDC callback) and the proxy-mode logout in `internal/server` are updated to set/clear the returned cookie set.

## Testing

- `go build ./...`, `go vet`, `go test ./pkg/auth/... ./internal/auth/... ./internal/server/...` — all pass.
- Unit coverage in `pkg/auth/cookie_test.go`: single/chunked round-trip, dropped/tampered-chunk rejection, chunk-count cap, representation-switch cleanup, logout clearing, and oversized-session refusal.
- Validated end-to-end against a real OIDC authorization-code flow with a 350-group identity (a 5-chunk session): the callback issues `Secure` chunk cookies, reassembly authenticates with all 350 groups intact, a tampered chunk is rejected with 401, and logout clears the main + meta + all chunk cookies. Also exercised the same cookie path live in proxy auth mode.

## Notes

- A very large group set can exceed a reverse proxy's request-header budget (16 chunks ≈ a ~59 KB `Cookie` header) before it reaches Radar; the `413` refusal path is the clean fallback. Worth a short note in the auth docs (follow-up).

---

> [!NOTE]
> **Medium Risk**
> Touches core authentication cookie issuance and parsing for every request; behavior is well bounded (chunk cap, HMAC, refusal path) but regressions could break login or logout for OIDC and proxy modes.
> 
> **Overview**
> Large OIDC sessions (many groups / big ID tokens) no longer fail silently when a single `radar_session` cookie would exceed the browser ~4 KB limit. **`pkg/auth/cookie.go`** now returns **multiple cookies**: one signed cookie when the payload fits, otherwise numbered `radar_session_chunk_*` plus a `radar_session_chunks` meta-cookie, with HMAC verified over the **reassembled** value. Oversized identities beyond a capped chunk count get **clear-only** cookies; the OIDC callback detects that via `sessionIssued` and responds with **413** instead of redirecting into a login loop.
> 
> **Call sites** (middleware sliding TTL / proxy header auth, OIDC login & logout, proxy `handleLogout`) **set or clear every cookie** in the returned slice. **`ClearSessionCookie`** now takes the request, mirrors **Secure** on expirations, and wipes chunk cookies so chunked sessions cannot survive logout.
> 
> **Tests** cover chunk round-trip, tampering, representation switches, logout clearing, and refusal paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a65446518ed8b7bc61ed8e134136d66de47ee92f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
Thanks for this, @jollyekb — the chunked-cookie approach is the right call for the oversized-OIDC-token problem (#527), and the core design (split the signed value across `_chunk_N` cookies + a `_chunks` meta-cookie, HMAC over the reassembled blob) is sound. I rebased the branch onto current `main` and pushed a couple of commits on top to get it merge-ready rather than leave it waiting. Summary of what changed and why:

**Correctness / security**
- **Cookie name** was `radar_seion` — restored to `radar_session` (the typo would have invalidated every existing session on deploy).
- **Logout** (`ClearSessionCookie`) only cleared the main cookie, so a chunked session survived logout (the chunks reassembled on the next request). It now takes the request and expires every chunk + the meta-cookie.
- **Representation switches** could keep a stale session: issuing chunked cookies now expires the main cookie, and issuing a single cookie clears the chunk meta-cookie (parse only follows the chunk path when the meta is present, so dropping it neutralizes any stale chunks).
- **OIDC `Secure`**: the callback derived `Secure` from `X-Forwarded-Proto`, but `/auth/callback` is auth-exempt so middleware never re-secures it — a downgrade behind TLS terminators that don't set that header. Forced `Secure` for OIDC to match the middleware invariant.
- Bounded chunk reassembly (`maxCookieChunks`) so a forged `_chunks` meta-cookie can't drive an unbounded read loop, and a warning when a session needs more chunks than the parser accepts.

**Tests**
- The signature change to `[]*http.Cookie` had broken the existing auth test packages (they no longer compiled). Fixed the call sites and added coverage for the chunked path: round-trip, dropped/tampered-chunk rejection, count cap, main-cookie precedence, logout-clears-chunks, and both representation transitions.

This went through a cross-model security review (the HMAC/forgery, reorder/drop, and normal-logout paths all came back clean). `go build`, `go vet`, and the `pkg/auth` + `internal/auth` test packages all pass.

One thing we deliberately did **not** add: the `__Secure-` cookie name prefix mentioned in the original issue. The `Secure` flag covers the actual 4 KB-rejection bug; the prefix is a separate hardening step we can do as a follow-up. Appreciate the contribution!

nadaverell:
Thanks @cursor — both addressed in `078f7fca`:

1. **Over-max chunks** — `createChunkedCookies` now refuses to issue a chunk set above the parser cap (returns only clearing cookies + logs an error), so the browser never gets cookies the server would then reject. Added `TestCreateSessionCookie_RefusesOversizedSession`.
2. **Clear cookies omit Secure** — `expireCookie` now mirrors the `Secure` attribute, and `ClearSessionCookie` derives it from the request, so a clear reliably evicts a Secure session cookie. Logout also now clears every possible chunk index (not just the advertised count) to avoid stale-chunk residue after a shrink.

nadaverell (approved):

PR #569, opened 2026-04-29
feat(cluster-card): cloudMode capability + adapt ClusterHealthCard for embedded use

nadaverell:
## Summary

Adapts the OSS cluster-health card so it stays useful when Radar is embedded in Radar Cloud, fixes a broken platform-logo path that only manifested in Cloud, and cleans up the `in-cluster` bootstrap-name leak that's been showing through to users.

## What changes

**`/api/capabilities` learns `cloudMode`.** Sourced from the existing `RADAR_CLOUD_MODE=true` env that the helm chart already sets when `cloud.enabled=true`. Single source of truth — no new prop on the radar-app library API surface.

**ClusterHealthCard branches on `cloudMode`:**
- Hides the local-MCP indicator. The OSS framing (`localhost:9280/mcp`, no auth, click-to-set-up dialog) is wrong in Cloud where the MCP endpoint is `api.radarhq.io/mcp`, PAT-authed, and org-wide. Cloud surfaces this from the hub Home dashboard instead — see the paired hub-web PR for the new MCPTile.
- Hides the H2 cluster name. In Cloud, the HubClusterSwitcher in the top bar is the canonical cluster identity; repeating it in the card body just made the card feel introduced rather than described.

**`in-cluster` headline fallback (independent of cloudMode).** When Radar runs in-cluster with no kubeconfig context, the bootstrap default sets `cluster.name = "in-cluster"`. The card was rendering that string verbatim as the H2. Now falls back to the platform label so the headline reads "Google Kubernetes Engine" instead.

**Platform icons bundled as Vite source assets.** GKE / EKS / AKS icons used to live at `web/public/icons/*` and were referenced as absolute `/icons/...` paths. That 404s in Cloud because `public/` isn't part of the published `@skyhook-io/radar-app` npm package — only `src/` ships. Moved into `web/src/assets/platform-icons/` and imported as Vite assets so they bundle into consumers.

## Companion PR

`radar-hub-web`: new `MCPTile` on the Home dashboard (the new surface for MCP discovery in Cloud). Pairs with the suppressed per-cluster card here.

## Version

Bumps `@skyhook-io/radar-app` to `0.2.2` for hub-web consumption.

## Verification

- `make tsc` clean
- `go build ./...` clean
- `go test ./internal/k8s/ ./internal/server/` clean
- Visual-test of the standalone OSS path: `cluster-view-baseline-old-radar-app.png` (still shows old behavior — expected, since the visual-test stack doesn't set `RADAR_CLOUD_MODE=true` and uses the `--cloud-url` cloud-dialer path against a real kubeconfig context, so neither the cloudMode branches nor the `in-cluster` fallback activates locally). The fixes target the production helm-deployed in-cluster path.

---

> [!NOTE]
> **Medium Risk**
> Touches Cloud-mode detection used for auth hardening and settings behavior, and changes the capabilities API shape consumed by multiple frontends. Mostly additive and backward-compatible, but mis-detection could affect Cloud/OSS security posture and UI behavior.
> 
> **Overview**
> Adds a centralized `internal/cloud.Mode()` (ParseBool-based) + startup logging for `RADAR_CLOUD_MODE`, and switches `main`, auth middleware, and server settings logic to use it to avoid silent Cloud/OSS mode mismatches.
> 
> Extends `/api/capabilities` with a `deployment.mode` value (`local`/`in-cluster`/`cloud`) and updates the TS types + capabilities context defaults so newer frontends remain compatible with older backends.
> 
> Updates `ClusterHealthCard` to render appropriately when embedded in Cloud: bundle platform icons as Vite assets, hide redundant cluster headline and local MCP card in `cloud` mode, and avoid showing the `in-cluster` sentinel name by falling back to the platform label.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 66da65597af063657be566c3588c6c24b2f4711f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
## Review-feedback follow-ups (commit d87befa)

Addresses findings from the comprehensive review:

- **Cloud-mode env-var hardening**: New `internal/cloud/mode.go` is the single source of truth — `strconv.ParseBool` so common typos (`True` / `1` / `TRUE`) parse correctly instead of silently degrading Cloud → OSS mode (which would have leaked per-user theme/pinnedKinds into the shared in-cluster `settings.json`). Logs resolved value at startup; one-time WARNING on unparseable. Replaces the three independent `os.Getenv("RADAR_CLOUD_MODE") == "true"` reads in `server.go`/`auth/middleware.go`/`main.go`.
- **Orphan icon cleanup**: Deleted `web/public/icons/` entirely. The 3 platform icons moved to `web/src/assets/platform-icons/` in the prior commit; the 3 unused `cloud-provider-*.png` files (never referenced anywhere) went too.
- **Comment polish**: `capabilities.go` field comment now explains intent (suppress chrome the hub renders) not mechanism. `headlineName` block in ClusterHealthCard describes both fallback branches now that there are two derivation steps. Cross-repo refs (HubClusterSwitcher, HomePage.tsx) softened to concept-level pointers.

### Deferred follow-up (not in this PR)

The type-design review flagged `cloudMode: bool` as a leaky abstraction (rated 2/5 encapsulation) and recommended a `deployment.mode: 'oss' | 'oss-in-cluster' | 'cloud'` enum that would also absorb the `cluster.name === 'in-cluster'` heuristic. Valid critique — but it's a substantive design call that deserves its own PR + review cycle. Filed mentally as a follow-up; the bool ships the current cluster-card work without overcommitting to a shape.

Tests + build clean: `go build ./... && go test ./internal/k8s/ ./internal/server/ ./internal/auth/ ./internal/cloud/`.

nadaverell:
## Type-design refactor: `cloudMode: bool` → `deployment.mode` enum (commit 45e58ad)

Took the type-design review's recommendation. The bool was leaky and the parallel `cluster.name === 'in-cluster'` string-sniff was the same concept fragmented elsewhere — both gone, replaced by a single closed enum:

```typescript
type DeploymentMode = 'local' | 'in-cluster' | 'cloud'
interface Deployment { mode: DeploymentMode }
```

- `local`: Radar binary on a developer's machine with kubeconfig (most OSS use)
- `in-cluster`: Radar pod observing its own cluster, no kubeconfig
- `cloud`: Radar pod tunneled to Radar Cloud's hub (`RADAR_CLOUD_MODE=true`)

Backend computes the mode in `handleCapabilities` from sources we already had: `cloud.Mode()` for the cloud branch, `k8s.GetKubeconfigSummary().Mode == "in-cluster"` for the bootstrap sentinel detection. Cloud beats in-cluster (Cloud is technically in-cluster + tunnel, but the user-visible behavior is the cloud-tunnel half).

ClusterHealthCard's frontend logic gets cleaner:
- `cloudMode` → `deployment.mode === 'cloud'`
- `cluster.name === 'in-cluster'` string-sniff → `deployment.mode !== 'local'` (either in-cluster or cloud — both lack a meaningful kubeconfig context for the headline)

Verified end-to-end against the local stack with `RADAR_CLOUD_MODE=true`:

```
$ curl /api/capabilities
{"deployment":{"mode":"cloud"},...}
```

Cluster view: GKE platform logo renders, no H2 cluster name (HubClusterSwitcher carries it), no MCP card. Screenshot: `v6-cluster-deployment-mode-cloud.png`.

Backwards-compat: `cloudMode` field is removed entirely. `@skyhook-io/radar-app@0.2.2` is the only published version that ever exposed it (current branch); hub-web doesn't read the field directly (only the embedded RadarApp does). Clean cut — no deprecation cycle needed.

Build + tests: `go build ./... && go test ./internal/k8s/ ./internal/server/ ./internal/auth/ ./internal/cloud/` clean. `make tsc` clean.

PR #572, opened 2026-04-29
fix(resources): stop "Updated Xs" timer resetting on no-op refetches

eliran-ops:
## Summary

Two related bugs in the Resources view header (SKY-820):

- **16**: The "Updated Xs" timer briefly reset to "<1s" when reopening filters, suggesting a real refetch was triggered for a UI-only action.
- **10**: Age values in the resource list felt like they shifted between refreshes, with no fixed reference to verify against.

## Root cause (bug 16)

The effect that drives `lastUpdated` was keyed on React Query's `dataUpdatedAt`, which bumps on every successful fetch — even when the response is byte-identical and structural sharing returns the same `resources` reference. Window focus, mount, and any sibling subscriber issuing the same queryKey can all trigger such no-op refetches, each one resetting the visible timer.

## Fix

Single file, +65/-12:

1. **Gate the `lastUpdated` effect on a real cache mutation**: `if (resources === lastDataRef.current) return`. React Query's structural sharing returns the same reference when responses are byte-identical, so reference equality is the cheapest correct way to detect "real change vs no-op".

2. **Switch the toolbar label to coarse buckets** (`just now` → `Xm` → `Xh` → `Xd`). The previous behavior was per-second `formatAge` — distracting in the periphery without aiding any user decision (a label going `5s` → `6s` → `7s` carries zero actionable info). Buckets cross at minute / hour / day boundaries, so the label updates roughly once per minute under an hour, once per hour up to a day, etc. — and a single `setTimeout` to the next boundary replaces the 1Hz `setInterval`.

3. **Isolate the label in its own `LastUpdatedLabel` component** so re-renders don't cascade into the parent's virtualized table.

4. **Bug 10 — Age column tooltip**: wrap the relative-age cell in a Tooltip showing the absolute `creationTimestamp.toLocaleString()`, so users have a stable reference when relative ages feel inconsistent. Also handles the falsy-`creationTimestamp` case explicitly.

## UX rationale for buckets

Continuous per-second updates are flagged as distracting / cognitively expensive across UX research (Smashing's [Real-Time Dashboards](https://www.smashingmagazine.com/2025/09/ux-strategies-real-time-dashboards/), accessibility guidance via WCAG / web.dev). Consumer products (Slack, Linear, GitHub notifications) all converged on coarse buckets. Grafana's own Clock Panel documents "if second ticking annoys you, change the time format" as the escape hatch — strong signal the per-second tick is a known irritation. For the toolbar staleness indicator, users care about "is this fresh / a few minutes old / stale" — bucket granularity, not seconds.

## Test plan

- [x] `npm test` (k8s-ui) — 211 passing
- [x] Visual: header shows "Updated just now" on fresh data, stable across 15s sampling; no flicker; no "<1s" reset on window-focus / refetch events
- [x] Visual: Age cell tooltip shows absolute creation timestamp on hover

### DIFF
--- a/packages/k8s-ui/src/components/resources/ResourcesView.tsx
+++ b/packages/k8s-ui/src/components/resources/ResourcesView.tsx
@@ -1752,6 +1752,51 @@ function getInitialFiltersFromURL() {
 // Sort state type
 type SortDirection = 'asc' | 'desc' | null
 
+// Coarse "just now / Xm / Xh / Xd" buckets — finer-grained updates
+// add motion in the periphery without aiding any user decision.
+function formatLastUpdatedBucket(elapsedMs: number): string {
+  const elapsedSec = Math.max(0, Math.floor(elapsedMs / 1000))
+  if (elapsedSec < 60) return 'just now'
+  const minutes = Math.floor(elapsedSec / 60)
+  if (minutes < 60) return `${minutes}m`
+  const hours = Math.floor(minutes / 60)
+  if (hours < 24) return `${hours}h`
+  return `${Math.floor(hours / 24)}d`
+}
+
+// ms until the displayed bucket would change.
+function msToNextBucket(elapsedMs: number): number {
+  const elapsed = Math.max(0, elapsedMs)
+  if (elapsed < 60_000) return 60_000 - elapsed
+  if (elapsed < 3_600_000) return 60_000 - (elapsed % 60_000)
+  if (elapsed < 86_400_000) return 3_600_000 - (elapsed % 3_600_000)
+  return 86_400_000 - (elapsed % 86_400_000)
+}
+
+// Isolated subtree so re-renders don't cascade into the parent's
+// virtualized table.
+function LastUpdatedLabel({ lastUpdated }: { lastUpdated: Date }) {
+  const [, force] = useState(0)
+  useEffect(() => {
+    let id: ReturnType<typeof setTimeout>
+    function schedule() {
+      const delay = Math.max(1000, msToNextBucket(Date.now() - lastUpdated.getTime()))
+      id = setTimeout(() => {
+        force(t => t + 1)
+        schedule()
+      }, delay)
+    }
+    schedule()
+    return () => clearTimeout(id)
+  }, [lastUpdated])
+  return (
+    <div className="flex items-center gap-1.5 text-xs text-theme-text-tertiary">
+      <Clock className="w-3.5 h-3.5" />
+      <span>Updated {formatLastUpdatedBucket(Date.now() - lastUpdated.getTime())}</span>
+    </div>
+  )
+}
+
 export function ResourcesView({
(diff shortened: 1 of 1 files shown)


nadaverell:
Ran a full review pass — code/silent-failure checks are clean, two real findings to address before merge:

## 1. Comments reference PR/ticket numbers (will rot)

CLAUDE.md says comments must never reference current task / PR / issue. Several do:

- `ResourcesView.tsx:1761` — `LastUpdatedLabel` JSDoc ends with "(Cursor Bugbot caught this on PR #572.)" — strip
- `ResourcesView.tsx:2670` — same "(Cursor Bugbot caught this on PR #572.)" trailer; this whole comment block also duplicates the JSDoc 30 lines down on `LastUpdatedLabel` — delete or trim to one sentence
- `ResourcesView.tsx:2648` — drop "(SKY-820 / bug 16)" trailer; the WHY explanation above it is good, keep that
- `ResourcesView.tsx:3991` — drop "Reported in SKY-820" from the age-tooltip comment
- `useNow.ts:24` — drop "SKY-820 captured users perceiving…"; keep the behavioral explanation

The substantive WHYs (React Query bumps `dataUpdatedAt` on byte-identical responses, reference-vs-value gating) are exactly the kind of comment the codebase wants — they prevent a future "simplification" PR from re-introducing the bug. Just strip the ticket trailers.

## 2. The actual fix is untested

You extracted `shouldScheduleNow` so it'd be unit-testable — good instinct. Apply the same pattern to the load-bearing change:

```ts
// pure helper next to shouldScheduleNow
export function shouldBumpLastUpdated(dataUpdatedAt: number, resources: unknown, lastRef: unknown): boolean
```

Then 3 tests: no bump on identical reference, bump on new reference, no bump when `dataUpdatedAt` is falsy. ~10 lines. Without it, a future refactor that drops the `lastDataRef` guard or swaps `!==` for deep equality silently re-introduces the bug.

The `useNow` "callable hook with documented signature" test (`useNow.length <= 1`) is near-useless — TypeScript catches that first. Either drop it or replace with a fake-timers test against the effect body extracted as a pure function.

## Otherwise

The reference-stable `lastUpdated` logic is sound, structural sharing reasoning checks out, `LastUpdatedLabel` extraction correctly isolates the 1Hz re-render from the virtualized list, age tooltip null-guard is fine, no leaks in `useNow` cleanup. Ship after the two fixes above.

eliran-ops:
@nadaverell @hisco — ready for review. Cursor bugbot found 2 issues: tests didn't exercise the `useNow` hook itself, and `useNow(1000)` re-rendered all of `ResourcesView` (~4kloc) every second. Both addressed: 031411d (test imports the hook), b8b25ed (extracted `LastUpdatedLabel` so the 1Hz tick isolates to a tiny subtree). Cursor Bugbot SUCCESS on current HEAD.

PR #576, opened 2026-04-29
fix(sidebar): stop counts flickering to 0 during initial load

eliran-ops:
## Summary

Sidebar resource counts went to "0" for every kind during the window between mount and the count payload arriving — visible as ~2s of "data loss" on the cloud product where the API has WAN latency. (SKY-824 bug 39.)

## Root cause

The counts memo coerced every missing key to 0:

```ts
results[key] = resourceCounts[key] ?? 0
```

This conflates "API hasn't replied yet" with "API confirmed zero of this kind", rendering a confident "0" badge for every resource during the loading window.

## Fix

Change the internal counts type from `Record<string, number>` to `Record<string, number | null>`:
- `null` = not loaded; rendered as an en-dash placeholder with `text-theme-text-disabled` so it's visually subordinate
- `0` = API confirmed empty; rendered as "0"
- Category "show this kind" filter treats null as "show, might have data" (don't pre-emptively hide kinds during load)
- Category totals coerce nulls to 0 for the collapsed-summary badge — honest because the total is over a known finite set of kinds

## Out of scope

The related one-frame flat-fallback flicker during nav transitions (bug 65 — sidebar briefly collapses to the flat layout when `apiResources` is `undefined` for one render on remount) is intentionally **not** addressed here. The clean fix would be to move state above the `BaseResourcesView` remount boundary, which is a larger structural change. The flat-fallback flicker is less alarming than the "0" flicker; this PR addresses the more visible symptom.

The `Checks` page mentioned in bug 40 lives in `radar-hub-web`, not radar OSS.

## Test plan

- [x] `npm test` (k8s-ui) — 216 passing, none affected
- [x] Manual: navigate to `/resources/pods` on a slow connection — counts show en-dash placeholders during load, not "0"

### DIFF
--- a/packages/k8s-ui/src/components/resources/ResourcesSidebar.tsx
+++ b/packages/k8s-ui/src/components/resources/ResourcesSidebar.tsx
@@ -94,7 +94,9 @@ const CORE_RESOURCE_TYPES = [
 // Resource type button in sidebar
 interface ResourceTypeButtonProps {
   resource: APIResource
-  count: number
+  /** `null` means "count not loaded yet" — rendered as a placeholder so
+   *  the badge doesn't flicker to "0" while the API call is in flight. */
+  count: number | null
   isSelected: boolean
   /** Keyboard-highlight state (arrow nav in the filter input). */
   isHighlighted?: boolean
@@ -155,9 +157,11 @@ const ResourceTypeButton = forwardRef<HTMLButtonElement, ResourceTypeButtonProps
             <span className={clsx(
               'text-xs py-0.5 rounded text-center font-mono',
               isSelected ? 'bg-skyhook-500/30 selection-text' : 'bg-theme-elevated',
-              count < 1000 ? 'w-8' : 'w-9'
+              count === null
+                ? 'w-8 text-theme-text-disabled'
+                : count < 1000 ? 'w-8' : 'w-9',
             )}>
-              {count}
+              {count === null ? '–' : count}
             </span>
           )}
         </div>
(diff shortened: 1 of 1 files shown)


eliran-ops:
@nadaverell @hisco — ready for review. No outstanding cursor findings. Cursor Bugbot SUCCESS on current HEAD.

eliran-ops:
## Visual test - no flicker observed

### Initial load (`/resources/pods`)
Hierarchical sidebar (Workloads / Networking / Configuration / Storage / Access Control / Cluster) renders directly with real counts. No `0`-flash, no flat-fallback flicker:

fresh-load

### After fast nav `/resources/pods` → `/audit` → `/resources/deployments`
Sidebar layout is preserved across nav transitions. The hierarchical layout doesn't drop to the flat fallback even for a single frame:

nav-transition

### DOM check (post-load steady state)
All sidebar counts are real numbers - no en-dashes left over (correct for steady state where every kind has a known count):
```
[CronJob: 1, DaemonSet: 37, Deployment: 103, Job: 6, Pod: 232, ReplicaSet: 505, StatefulSet: 4, Ingress: 33, ...]
```

### Coverage caveat
The bug's distinguishing symptom is a ~2s window during initial load. On this fast local network the window is too short to catch a viewport-stable screenshot. The DOM has no `0`-counts at any sampled point, and the layout never collapsed to the flat fallback during multiple cross-page navs.

The `useLastDefined` hook + `null`-vs-`0` distinction (en-dash placeholder) are unit-tested per the PR's checked test plan; the visible behavior here matches what the fix promises.

PR #580, opened 2026-04-29
fix(topology): make the "Policies" toggle actually paint policy effects

eliran-ops:
Closes part of [SKY-832](https://linear.app/skyhook/issue/SKY-832).

## Symptom

Clicking the "Policies" button in the topology toolbar highlighted it blue, opened a `?policyEffect=true` SSE stream, and… changed nothing visible on the canvas.

## Root cause

The backend correctly annotates each edge with a `policyEffect` of `allowed`, `blocked`, or `unprotected` when policy effect is enabled, and the field is plumbed through to `TopologyEdge` in the shared types. But `buildEdges` / `getEdgeColor` / `getEdgeStyle` in `TopologyGraph.tsx` never consulted that field, so the edge colour and stroke were chosen entirely from `edge.type` / `viewMode`. The user's "Policies" click effectively had no rendering side-effect.

## Fix

- `getEdgeColor` and `getEdgeStyle` now accept an optional `policyEffect` and prioritise it over view-mode and per-type colors:
  - `allowed`     → emerald (`#10b981`)
  - `blocked`     → red (`#ef4444`), thicker (+0.5), dashed (`6 4`), and never animated (animation reads as "traffic flowing", opposite of `blocked`).
  - `unprotected` → amber (`#f59e0b`)
- `buildEdges` passes `edge.policyEffect` through to both helpers, and to the React Flow `markerEnd.color` so arrowheads stay in sync.
- `TopologyControls` now renders a small inline legend next to the toggle whenever it's on, so the colour meanings are obvious without scrolling/docs. Also adds `aria-pressed` to the button for screen readers.

## Tests

New `packages/k8s-ui/src/components/topology/edge-style.test.ts` (10 cases) pins:

- Per-type palette unchanged when no policy effect.
- Traffic-view green unchanged when no policy effect.
- Policy effect overrides per-type and traffic-view colours.
- `blocked` ⇒ red + dashed + +0.5 stroke (in both view modes).
- Unknown policy strings fall through gracefully.
- Animation flag still produces the dotted dash for traffic edges when no policy effect.

## Out of scope

Bugs 30 and 31 were also tracked under SKY-832:

- **30**: "Show All" clips the top row of `GatewayClass` nodes — needs an `elk` padding / `fitViewOptions` audit.
- **31**: Traffic mode renders an empty canvas while the sidebar reports 241/242 resources visible — needs investigation of the traffic-view edge filter.

Both will be follow-up PRs against this issue.

## Test plan

- [x] `npm test` (k8s-ui) — 79 passed (10 new for edge styling).
- [x] `npm run tsc` (k8s-ui + web) — clean.
- [x] `npm run build` (web) — clean.
- [ ] Manual: in the topology view, click "Policies"; verify that allowed edges turn green, blocked edges turn red + dashed + thicker, and the legend strip appears next to the toggle.

---

> [!NOTE]
> **Medium Risk**
> Touches core topology edge rendering and animation behavior, which could change visuals/perf across large graphs, but is isolated to UI styling with added unit coverage.
> 
> **Overview**
> Makes the **"Policies"** toggle actually affect the topology canvas by threading each edge’s `policyEffect` into edge rendering.
> 
> `getEdgeColor`/`getEdgeStyle` now prioritize `policyEffect` (allowed/blocked/unprotected) over view-mode/type colors, render `blocked` edges as red + dashed + thicker, and prevent `blocked` edges from animating; arrowhead colors are kept in sync. The toolbar also shows an inline legend when enabled and adds `aria-pressed` for accessibility, with new Vitest coverage pinning the styling/priority rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a7cbcd5d5a7ca66b36b37fb9b59311b4110c280. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/k8s-ui/src/components/topology/TopologyControls.tsx
+++ b/packages/k8s-ui/src/components/topology/TopologyControls.tsx
@@ -25,20 +25,46 @@ export function TopologyControls({
 }: TopologyControlsProps) {
   return (
     <div className="absolute top-4 right-4 z-10 flex items-center gap-2">
-      {/* Policy effect toggle */}
+      {/* Policy effect toggle + legend */}
       {onShowPolicyEffectChange && (
-        <button
-          onClick={() => onShowPolicyEffectChange(!showPolicyEffect)}
-          className={`flex items-center gap-1.5 px-2.5 py-1.5 text-xs rounded-lg border transition-colors ${
-            showPolicyEffect
-              ? 'bg-indigo-600 text-white border-indigo-600'
-              : 'bg-theme-surface/90 backdrop-blur text-theme-text-secondary border-theme-border hover:text-theme-text-primary'
-          }`}
-          title="Show NetworkPolicy effects on edges"
-        >
-          <ShieldCheck className="w-3.5 h-3.5" />
-          Policies
-        </button>
+        <>
+          {showPolicyEffect && (
+            <div
+              className="flex items-center gap-2 px-2.5 py-1.5 text-[10px] uppercase tracking-wide rounded-lg border bg-theme-surface/90 backdrop-blur text-theme-text-secondary border-theme-border"
+              role="legend"
+              aria-label="NetworkPolicy effect legend"
+            >
+              <span className="flex items-center gap-1">
+                <span className="inline-block w-3 h-[2px]" style={{ background: '#10b981' }} />
+                allowed
+              </span>
+              <span className="flex items-center gap-1">
+                <span
+                  className="inline-block w-3 h-[2.5px]"
+                  style={{ background: 'repeating-linear-gradient(90deg, #ef4444 0 4px, transparent 4px 6px)' }}
+                />
+                blocked
+              </span>
+              <span className="flex items-center gap-1">
+                <span className="inline-block w-3 h-[2px]" style={{ background: '#f59e0b' }} />
+                unprotected
+              </span>
+            </div>
+          )}
+          <button
+            onClick={() => onShowPolicyEffectChange(!showPolicyEffect)}
+            className={`flex items-center gap-1.5 px-2.5 py-1.5 text-xs rounded-lg border transition-colors ${
+              showPolicyEffect
+                ? 'bg-indigo-600 text-white border-indigo-600'
+                : 'bg-theme-surface/90 backdrop-blur text-theme-text-secondary border-theme-border hover:text-theme-text-primary'
+            }`}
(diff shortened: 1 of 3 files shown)


nadaverell:
Thanks for tackling this — but the premise is off.

`Edge.PolicyEffect` has never been written to anywhere in the codebase (`git log -S ".PolicyEffect ="` is empty). It was scaffolded in `ad0a6a2` with the evaluation logic deferred — see `docs/NETWORK_POLICY_FUTURE.md` from that commit (since deleted, recoverable), section 1: **"~5-7h MVP"** for the backend half. Your renderer is reading a field that's always `null`. The unit tests pass because they pass `'allowed'` directly to `getEdgeColor`.

Also: the toggle isn't a no-op today. `annotateNodePolicyCoverage` already adds a `!`/`✓` badge to each workload card via `policyStatus` — that's the shipped feature.

Two paths:

1. **Expand scope** — implement the backend evaluation per the doc (`LabelSelectorAsSelector()` against Service→Workload edges in `builder.go`). Then your renderer actually does something.
2. **Close** — I'll restore the FUTURE doc so the deferred-work signal is back in tree.

Renderer code itself is fine; it just needs a producer.

eliran-ops:
You're right, closing this.

I went back and confirmed:

- `git log -S '.PolicyEffect ='` is empty — nothing in the codebase ever writes that field on an edge, so my renderer change paints from a value that's always `null` and the user-visible behaviour is unchanged.
- The unit tests pass because they call `getEdgeColor('blocked')` directly rather than going through `buildEdges` with a real `TopologyEdge` from the SSE stream.
- And as you noted, `annotateNodePolicyCoverage` already attaches the `!`/`✓` workload badge — so the toggle isn't a no-op today, the per-edge-effect rendering is just a separate (unimplemented) feature.

Implementing the backend evaluation properly is the right path; that's a separate, scoped PR with the LabelSelector work, the SSE annotation, and the rendering. I'll do that as a fresh issue rather than extend this one.

The two real SKY-832 bugs that this PR was supposed to also touch but ran out of scope on:

- **bug 30**: `Show All` clips the top row of `GatewayClass` nodes — needs an `elk` padding / `fitViewOptions` audit.
- **bug 31**: Traffic mode renders an empty canvas while the sidebar reports 241/242 visible — needs investigation of the traffic-view edge filter.

Will spin those out into their own PRs against SKY-832.

PR #584, opened 2026-04-30
fix(helm): make Catalog empty state actionable; signpost failed releases

eliran-ops:
Closes part of [SKY-829](https://linear.app/skyhook/issue/SKY-829).

## Why

Three SKY-829 bugs share one root cause: the Helm UI tells the user "something is up" but gives no path forward.

### 33 — "My Repos" empty state had no actionable button

The empty state read "Try updating your repositories" — even though the toolbar already had an Update-all handler. Now the empty state inlines:
- An **Update all repositories** button (calls the same handler as the toolbar; spinner + disable on `isBatchUpdating` / `!canHelmWrite`).
- A secondary "Or browse ArtifactHub instead" link.

### 58 — Failed Helm releases had no signpost to the recovery actions

Failed releases rendered as a bare red badge in the list. Users had no idea the row was the path forward (where the drawer exposes Rollback / History / Uninstall).

Now `failed` releases wrap a Tooltip ("Click row to view rollback / history / logs and recover") and append a chevron to make the affordance visible.

**Scope note**: the predicate `isHelmReleaseActionable` currently matches only `failed`. The three `pending-*` statuses (`pending-install`, `pending-upgrade`, `pending-rollback`) are deliberately excluded — they're transient by design and without release-age in the client we'd surface a stuck-recovery affordance during every routine install. See the JSDoc on `ACTIONABLE_HELM_STATUSES` for the in-flight false-positive rationale; a follow-up can include them once age is plumbed through.

### 22 — argocd APP VERSION rendered as bare `-`

Indistinguishable from "Radar lost the data." Some charts genuinely don't declare an `appVersion` in `Chart.yaml` — that's valid Helm. Now we render an em-dash with a `cursor-help` tooltip ("This chart did not declare an appVersion in Chart.yaml.") so the user knows it's upstream, not a Radar bug.

## Bonus fix: bulk-update aggregate toast

While implementing the empty-state Update button (bug 33), we doubled the visibility of an existing UX issue in `handleUpdateAllRepos`: the bulk loop silently `try/catch`-swallowed per-repo failures, and the global MutationCache fired N identical "Failed to update repository" toasts with no repo name.

Replaced with:
- New `useUpdateRepositorySilent` (no `meta.errorMessage`) so the global MutationCache doesn't fire per-call toasts during the batch.
- Shared `updateRepositoryFn` + `invalidateHelmAfterRepoUpdate` so `useUpdateRepository` and the silent variant can't drift on URL / error parsing / invalidation keys.
- `isBatchUpdating` state to guard against re-entrant batches (`mutation.isPending` flips false between sequential `mutateAsync` calls).
- One aggregate toast at the end: all-success → "Updated N repositories"; all-fail → "Failed to update N repositories" + names; partial → "Updated N/M repositories" + "Failed: a, b, c".

## Shared helper

Added `isHelmReleaseActionable` to `packages/k8s-ui/utils/badge-colors` (next to the existing `getHelmStatusColor`) so the actionable-set rule is shared across radar OSS and cloud, and unit-tested.

## Tests

`packages/k8s-ui/src/utils/helm-status.test.ts` (7 cases) pins:
- `failed` → true
- The three `pending-*` statuses → **false** (with a comment explaining the in-flight exclusion)
- Case-insensitive
- Success / normal statuses (`deployed`, `superseded`, `uninstalled`, `uninstalling`, generic `error`) → false
- null / undefined / empty / unknown → false

## Out of scope

- **34** (ArtifactHub card height inconsistency) — needs a flex-vs-grid layout audit
- **56** (keda chart shows `v1alpha1` as chart version on prod) — looks like backend captured API version into chart version; needs Go-side investigation

Both will be follow-up PRs against this issue.

## Test plan

- [x] `npm test` (k8s-ui) — 7/7 `isHelmReleaseActionable` cases pass
- [x] E2E manual verification on kind cluster:
  - Chart with no `appVersion` → em-dash + tooltip ✓
  - Failed release row → chevron + tooltip + drawer ✓
  - Empty Catalog state → "Update all repositories" button ✓
  - Bulk update with mocked partial failure → aggregate toast `Updated 31/34 repositories` / `Failed: a, b, c` ✓

### DIFF
--- a/packages/k8s-ui/src/utils/badge-colors.ts
+++ b/packages/k8s-ui/src/utils/badge-colors.ts
@@ -286,6 +286,31 @@ export function getHelmStatusColor(status: string): string {
   return HELM_STATUS_COLORS[statusLower] || 'bg-theme-hover/50 text-theme-text-secondary'
 }
 
+/**
+ * Helm release statuses where the row UI should signpost the user
+ * toward the drawer (history / rollback / logs).
+ *
+ * Currently `failed` only. The `pending-*` statuses (install /
+ * upgrade / rollback) are excluded deliberately: they're Helm's
+ * normal in-flight states during every routine operation. Treating
+ * them as "actionable" would briefly attach an alarming chevron +
+ * tooltip to every install while it ran — indistinguishable from
+ * the genuinely-stuck case (controller crashed mid-flight). Until
+ * we have release age available client-side to disambiguate
+ * "in-flight" from "stuck > N min", we give up the stuck-detect
+ * signpost rather than wrongly alarm the common case.
+ *
+ * @see https://github.com/helm/helm/blob/dev-v3/pkg/release/status.go
+ */
+const ACTIONABLE_HELM_STATUSES: ReadonlySet<string> = new Set([
+  'failed',
+])
+
+export function isHelmReleaseActionable(status: string | null | undefined): boolean {
+  if (!status) return false
+  return ACTIONABLE_HELM_STATUSES.has(status.toLowerCase())
+}
+
 // =============================================================================
 // VULNERABILITY SEVERITY COLORS - for Trivy and other security scanners
 // =============================================================================
--- a/packages/k8s-ui/src/utils/helm-status.test.ts
+++ b/packages/k8s-ui/src/utils/helm-status.test.ts
@@ -1,0 +1,50 @@
+import { describe, it, expect } from 'vitest'
+import { isHelmReleaseActionable } from './badge-colors'
+
+// Pin the membership set so adding a new Helm status becomes a
+// deliberate decision here.
+
+describe('isHelmReleaseActionable', () => {
+  it('returns true for `failed`', () => {
+    expect(isHelmReleaseActionable('failed')).toBe(true)
+  })
+
+  it('is case-insensitive (matches Helm SDK serialisation variants)', () => {
+    expect(isHelmReleaseActionable('FAILED')).toBe(true)
+    expect(isHelmReleaseActionable('Failed')).toBe(true)
+  })
+
+  it('returns FALSE for the pending-* in-flight statuses', () => {
+    // These are Helm's NORMAL in-flight states during every
+    // routine install/upgrade/rollback. If we treated them as
+    // actionable, every routine install would briefly attach an
+    // alarming chevron + tooltip to its own row. Until we have
+    // release age available client-side to distinguish
+    // "transient" from "stuck > N min", we give up the
+    // stuck-controller signpost rather than alarm the common case.
+    expect(isHelmReleaseActionable('pending-install')).toBe(false)
+    expect(isHelmReleaseActionable('pending-upgrade')).toBe(false)
+    expect(isHelmReleaseActionable('pending-rollback')).toBe(false)
+  })
+
+  it('returns false for the success / normal statuses', () => {
+    expect(isHelmReleaseActionable('deployed')).toBe(false)
+    expect(isHelmReleaseActionable('superseded')).toBe(false)
+    expect(isHelmReleaseActionable('uninstalled')).toBe(false)
+  })
+
+  it('returns false for `uninstalling` (in-progress, not stuck)', () => {
+    expect(isHelmReleaseActionable('uninstalling')).toBe(false)
+  })
+
+  it('returns false for null / undefined / empty', () => {
+    expect(isHelmReleaseActionable(null)).toBe(false)
+    expect(isHelmReleaseActionable(undefined)).toBe(false)
+    expect(isHelmReleaseActionable('')).toBe(false)
+  })
+
+  it('returns false for unknown strings (defensive)', () => {
+    expect(isHelmReleaseActionable('mystery-status')).toBe(false)
+    expect(isHelmReleaseActionable('error')).toBe(false) // Helm uses 'failed', not 'error'
+  })
+})
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -2148,29 +2148,49 @@ export function useHelmRepositories() {
   })
 }
 
+// Shared mutation fn + cache invalidation for the helm-repo
+// update endpoint. Hoisted so useUpdateRepository and
+// useUpdateRepositorySilent can't drift on URL, error parsing, or
+// invalidation keys — only the toast meta differs.
+async function updateRepositoryFn(repoName: string): Promise<unknown> {
+  const response = await apiFetch(`${getApiBase()}/helm/repositories/${repoName}/update`, {
+    method: 'POST',
+  })
+  if (!response.ok) {
+    const error = await response.json().catch(() => ({ error: 'Unknown error' }))
+    throw new Error(error.error || `HTTP ${response.status}`)
+  }
+  return response.json()
+}
+
+function invalidateHelmAfterRepoUpdate(queryClient: ReturnType<typeof useQueryClient>) {
+  queryClient.invalidateQueries({ queryKey: ['helm-repositories'] })
+  queryClient.invalidateQueries({ queryKey: ['helm-charts'] })
+}
+
 // Update a repository index
 export function useUpdateRepository() {
   const queryClient = useQueryClient()
-
   return useMutation({
-    mutationFn: async (repoName: string) => {
-      const response = await apiFetch(`${getApiBase()}/helm/repositories/${repoName}/update`, {
-        method: 'POST',
-      })
-      if (!response.ok) {
-        const error = await response.json().catch(() => ({ error: 'Unknown error' }))
-        throw new Error(error.error || `HTTP ${response.status}`)
-      }
-      return response.json()
-    },
+    mutationFn: updateRepositoryFn,
     meta: {
       errorMessage: 'Failed to update repository',
       successMessage: 'Repository updated',
     },
-    onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: ['helm-repositories'] })
-      queryClient.invalidateQueries({ queryKey: ['helm-charts'] })
-    },
+    onSuccess: () => invalidateHelmAfterRepoUpdate(queryClient),
+  })
+}
+
+// Same endpoint as useUpdateRepository but without meta — the
+// caller surfaces ONE aggregate toast for the whole batch, so the
+// global MutationCache must NOT fire a per-call toast (otherwise
+// N failures = N identical "Failed to update repository" toasts
+// with no repo name).
+export function useUpdateRepositorySilent() {
+  const queryClient = useQueryClient()
+  return useMutation({
+    mutationFn: updateRepositoryFn,
+    onSuccess: () => invalidateHelmAfterRepoUpdate(queryClient),
   })
 }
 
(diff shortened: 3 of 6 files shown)


packages/k8s-ui/src/utils/badge-colors.ts
```diff
@@ -262,6 +262,34 @@ export function getHelmStatusColor(status: string): string {
...
+/**
+ * Helm release statuses that indicate the install or upgrade did not
+ * complete successfully and the user needs to take recovery action
+ * (rollback / uninstall / inspect logs).
+ *
+ * The Helm SDK distinguishes a handful of failure-shaped statuses;
+ * we treat them all as "actionable failure" in the list view so the
+ * row UI can prompt users toward the drawer's recovery actions.
+ *
+ * @see https://github.com/helm/helm/blob/main/pkg/release/status.go
```
nadaverell: bad link

packages/k8s-ui/src/utils/badge-colors.ts
```diff
@@ -262,6 +262,34 @@ export function getHelmStatusColor(status: string): string {
...
+ */
+const FAILED_HELM_STATUSES: ReadonlySet<string> = new Set([
```
nadaverell: these aren't failed

nadaverell:
Ran a deeper review pass. Comment-rot is the same pattern as #572 (PR/ticket numbers in comments), but two real findings deserve attention first:

## 1. `pending-*` ≠ failed — wrong predicate, potentially destructive UX

`packages/k8s-ui/src/utils/badge-colors.ts:273` — the set treats all three `pending-*` statuses as failure-shaped. But Helm emits `pending-install` / `pending-upgrade` / `pending-rollback` while an operation is **actively running** — they're the normal in-flight state, not a failure. Operations transition to `failed` on timeout/error, or remain pending only if the controller crashes mid-flight.

For a user mid-`helm install`, the row will now show *"Click row to view rollback / history / logs and recover"* — wrong, and potentially destructive (clicking rollback on a still-running install can leave the release in a worse state). The "stuck/orphaned" case the PR optimizes for is the minority; the common case during normal install traffic is "in progress."

The test at `helm-status.test.ts:21-23` codifies this as intended behavior, which would lock in the bug.

**Options:**
- **Narrow the set to `failed` only** — safest; gives up signposting on the truly-stuck case.
- **Keep pending-* but rewrite the tooltip:** *"Operation pending or stuck — click row to inspect."* No mention of rollback, accurate in both states.
- **Detect actually stuck:** would need release age (pending > N minutes), which isn't available from status alone — requires the timestamp.

I'd go with option 2. (The current copy `"Click row to view rollback / history / logs and recover"` reads like internal jargon anyway — slash-list of internal feature names. The rewrite covers both this finding and the copy nit.)

## 2. `handleUpdateAllRepos` silently absorbs per-repo failures

`web/src/components/helm/ChartBrowser.tsx:74-84` — the loop catches per-repo errors and continues. The empty-state button (a new entry point this PR adds) calls the same handler. If every repo fails, the user clicks "Update all," sees N identical "Failed to update repository" toasts (no repo name), and the empty state re-renders unchanged — no aggregate "0/N succeeded" signal. They just clicked the only action available and got back exactly what they started with.

Pre-existing, but you're widening the surface that hits this path. Worth fixing here:
- Include `repo.name` in the per-repo error
- Surface a single aggregate toast after the loop: *"Updated 3/5 repositories. Failed: foo, bar."*

## 3. Comments — same rot pattern as #572

CLAUDE.md: comments must never reference current task / PR / issue. Strip:

- `badge-colors.ts:289` — `(SKY-829 bug 58)` trailer
- `ChartBrowser.tsx:283-287` — diff narration ("used to read…") + bug ref → delete entirely; the JSX is self-explanatory
- `HelmView.tsx:450-452` — bug ref + over-explanation for em-dash → delete (tooltip text already conveys the meaning)
- `HelmView.tsx:464-469` — bug ref + diff narration → delete
- `helm-status.test.ts:1-9` — header references SKY-829 → trim to one sentence on the predicate's purpose

The `FAILED_HELM_STATUSES` JSDoc is the one comment worth keeping — the grouping decision IS non-obvious and a future contributor needs the rationale. But "failure-shaped" is vague; tighten the WHY (and if you take option 2 above, this comment will need rewriting anyway).

## Otherwise

Empty-state inline Update button is a clear win. RBAC gating + handler parity with the toolbar checks out (`canHelmWrite`, `helmWriteReason`, same `updateRepoMutation`). Em-dash + tooltip for missing `appVersion` is correct — `AppVersion` comes from the local Helm release secret, not an upstream fetch, so the "not declared in Chart.yaml" copy can't false-attribute a Radar bug.

packages/k8s-ui/src/utils/badge-colors.ts
eliran-ops: Fixed in fcb39b1: link points to `dev-v3` branch now (`main` was a 404 since helm uses `dev-v3`).

packages/k8s-ui/src/utils/badge-colors.ts
eliran-ops: Fixed in 243d1bb: `pending-*` statuses removed from the actionable set. Now only `failed` is actionable. Added a comment block explaining the trade-off (pending-* are normal in-flight states; treating them as actionable would alarm every routine install). See lines 265-281 in current HEAD.

eliran-ops:
@hisco @nadaverell — ready for review. Two human comments from @nadaverell on commit 75a8454 (bad helm status link, `pending-*` shouldn't be 'failed') replied inline — both addressed by fcb39b1 (link → `dev-v3`) and 243d1bb (narrowed ACTIONABLE to `failed` only). Cursor bugbot's 3 findings (batch update guard, dup mutation hook, pending-* misclassification) all addressed by 243d1bb / fcb39b1. Cursor Bugbot SUCCESS on current HEAD.

eliran-ops:
## Visual test

### Bug 58 - failed releases now signpost recovery actions ✓
Releases list shows 2 failed Helm releases (`mcp-s`, `prometheus`). The status badges have:
- Red `failed` text
- Chevron-right icon on the same line (`lucide-chevron-right` class on the inline svg) - the new affordance
- Tooltip on hover: **"Click row to view rollback / history / logs and recover"**

Releases list (chevrons visible next to each `failed` badge):
releases

Tooltip on hover (visible above the kube-fledged row):
tooltip

DOM excerpt confirming the chevron is inside the badge:
```html
<span class="badge ... bg-red-100 text-red-700 border-red-300 ...">
  failed
  <svg class="lucide lucide-chevron-right w-3 h-3 opacity-70" ...>
</span>
```

### Bug 33 - Catalog empty state with Update-all button
**Could not exercise** - this cluster has 79 charts across multiple repos in My Repos. The empty state only shows when there are zero repos. No regression on the populated state:
catalog

### Bug 22 - missing appVersion as em-dash with tooltip
**Could not exercise** - all visible charts in the Catalog declare an `appVersion`. Worth checking on a fresh chart (the unit tests already pin the helper).

### Verdict
Bug 58 visually verified end-to-end on real failed releases. Bugs 33 and 22 need cluster states this test cluster doesn't have - tests pass per the PR's checked test plan.

eliran-ops:
## Local visual verification

Tested on `gke_koalabackend_us-east1-b_nonprod-cluster-us-east1` (2 failed releases: `mcp-s`, `prometheus`). Branches needing cluster states the test cluster doesn't have were exercised by patching `window.fetch` in the page (mocked `/api/helm/charts` -> `[]` and `/api/helm/repositories/<name>/update` outcomes).

### Real-cluster paths

**Failed-release signposting (Installed list):** `failed >` chevron badge shows on both `mcp-s` and `prometheus` rows. DOM probe confirmed exactly 2 chevroned badges out of 13 rows -> `isHelmReleaseActionable` correctly narrowed to `failed` only. Tooltip "Click row to view rollback / history / logs and recover" verified via DOM.

**Failed-release drawer (mcp-s):** opens with red `failed` badge, full error description ("Release 'mcp-s' failed: 19 errors occurred: namespaces 'webrix-mcp-s' not found ..."), and trash + refresh actions in header.

### Mocked paths

**Catalog actionable empty state** (repos configured, 0 charts, no search term):
> No charts found
> Your repositories may be out of date.
> [Update all repositories]
> Or browse ArtifactHub instead

**Bulk-update toast aggregation** — single toast per click in all three outcomes (DOM probe confirmed 3 silent `/repositories/<name>/update` calls without per-call toasts -> `useUpdateRepositorySilent` working as designed):

| Outcome | Toast text |
|---|---|
| All 3 fail | `Failed to update 3 repositories` / `Failed: argo, bitnami, jetstack` |
| 2 ok / 1 fail | `Updated 2/3 repositories` / `Failed: jetstack` |
| All 3 ok | `Updated 3 repositories` (success ✓) |

### Console
No new errors related to this PR. Pre-existing background noise only (third-party CDN image 400, transient SSE reconnect).

Screenshots saved locally under `.playwright-mcp/visual-test/20260505-222436/` and `.playwright-mcp/visual-test/20260505-225333/` — will attach inline.

eliran-ops:
## Screenshots

### Real-cluster paths

**Failed-release signposting (Installed list)** — `failed >` chevron badge on `mcp-s` and `prometheus`:

helm-installed-with-failed

Tight crop of the actionable badge:

helm-failed-row-prometheus

**Failed-release drawer (mcp-s)** — error description + actions:

helm-failed-release-drawer

### Mocked paths (via patched `window.fetch`)

**Catalog actionable empty state** — repos configured, 0 charts, no search:

helm-catalog-actionable-empty

**Bulk-update toast aggregation** — single toast per click in all three outcomes:

All 3 fail:
helm-bulk-update-toast-during

2 ok / 1 fail (names the failed repo):
helm-bulk-update-toast-partial

All 3 ok:
helm-bulk-update-toast-success

eliran-ops:
@nadaverell — all three findings from #issuecomment-4350877743 addressed; verified locally with screenshots in the comments above.

**1. `pending-*` ≠ failed** — went with **option 1** (narrow set to `failed` only). `pending-*` deliberately excluded; rationale documented inline at `packages/k8s-ui/src/utils/badge-colors.ts:266-280` and re-asserted at the test site (`helm-status.test.ts:19-30`) so flipping the membership later is a deliberate decision, not a drive-by edit.

**2. Bulk-update silent failures** — `handleUpdateAllRepos` now collects failures and surfaces one aggregate toast in three branches:
- all ok: `Updated 3 repositories`
- partial: `Updated 2/3 repositories` / `Failed: jetstack`
- all fail: `Failed to update 3 repositories` / `Failed: argo, bitnami, jetstack`

Implemented via a separate `useUpdateRepositorySilent` mutation so the global `MutationCache` doesn't fire N per-call toasts. Why-comments at `web/src/api/client.ts:2106-2110` and `ChartBrowser.tsx:70-79` (also covers the `isPending`-flips-between-calls batch-state guard).

**3. Comment rot** — stripped all SKY-/bug refs and diff-narration comments per CLAUDE.md. The one comment kept is the `ACTIONABLE_HELM_STATUSES` JSDoc — rewritten to explain the in-flight-vs-stuck tradeoff (your option 2 wording, applied at the predicate site rather than the tooltip).

One small inconsistency I'll fix in a follow-up nit: `helm-status.test.ts:5` quotes an older tooltip string ("Operation pending or stuck…") that doesn't match the live `"Click row to view rollback / history / logs and recover"`. Test logic is unaffected.

PR #595, opened 2026-04-30
fix(contexts): refresh registry against disk on each list call (SKY-834 bug 52)

eliran-ops:
Linear: [SKY-834](https://linear.app/skyhook/issue/SKY-834)

Companion to #590 (bug 54 — cert-health namespace mismatch).

## Why
**Bug 52**: junk / stale clusters lingering in the cluster dropdown.

In multi-file (isolated-load) mode, `contextRegistry` + `perFileConfigs` were built **once** in `setupIsolatedLoad` and never refreshed. Single-file mode reloads on every `GetAvailableContexts` call (it walks `RawConfig()` through `DeferredLoadingClientConfig`), but the multi-file path served whatever was in memory.

So the dropdown kept showing entries for:
- kubeconfig files removed from a watched directory,
- kubeconfig files rewritten by `kubectl config delete-context`,
- kubeconfig files written then deleted by CAPI when a managed cluster was destroyed.

All three look identical to the user: a row that's there but errors out on switch.

## What
New helper `refreshContextRegistry` — surgical, mtime-driven reconciliation against disk.

| Disk state | Action |
|---|---|
| File missing | Drop every registry entry pointing at it. Drop cached config + mtime. |
| File mtime moved | Re-parse THAT file only. Drop entries for removed contexts; add entries for new ones. |
| File mtime unchanged | No-op (avoids re-parsing N files on every dropdown open). |
| File present but parse-fails | Keep prior entries (user might be mid-edit; silent pruning would be worse). |

Wired into `GetAvailableContexts` (multi-file branch only) under `clientMu` write lock. Same per-file isolation guarantees as `buildContextRegistry` — colliding user/cluster names across files remain structurally impossible.

## Out of scope (deliberately deferred)
1. **Re-scanning kubeconfig DIRECTORIES for newly-added files.** Today, a freshly-created kubeconfig still requires a Radar restart to appear. Same workaround as before. Worth a follow-up but raises a syscall-budget question (every dropdown open shouldn't `ReadDir` all watched dirs without throttling).
2. **fsnotify-based eager refresh.** Same data, more moving parts; on-demand mtime check on dropdown open is enough to clear the junk-cluster complaint.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/k8s/` (all green; 5 new subtests for refresh)
- [x] Manual: start Radar with two kubeconfig files; `kubectl config delete-context foo` in one of them; reopen the dropdown — `foo` is gone, no restart needed.
- [x] Manual: `rm` one of the watched kubeconfig files; reopen the dropdown — every context from that file is gone.

### Live-cluster verification

Built `radar` with `--kubeconfig-dir /tmp/sky834-kubeconfigs/` containing:
- `file-a.yaml` — 2 contexts (`...nonprod...`, `...management-cluster`)
- `file-b.yaml` — 1 context (`...prod...`)

| Step | Disk action | `/api/contexts` | Response size |
|---|---|---|---|
| Initial | — | 3 contexts | 646B |
| `kubectl config delete-context management-cluster` (file-a) | mtime moves | 2 contexts (`nonprod` + `prod`) | 440B |
| `rm file-b.yaml` | file gone | 1 context (`nonprod`) | 225B |
| Noop re-list | none | 1 context, served in 47µs | 225B |

Both PR symptoms recover without restart. Noop re-list hits the mtime fast-path (45-100µs vs ~1ms when refresh fires) — refresh only re-parses on real disk changes. The unchanged file's context (`nonprod` in file-a) survives correctly through file-b's deletion, confirming per-file isolation is preserved.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes concurrency/locking and map publication semantics around shared kubeconfig state, which could impact context switching and listing behavior under load.
> 
> **Overview**
> Fixes stale “junk cluster” entries in multi-kubeconfig mode by tracking per-file mtimes and refreshing the in-memory `contextRegistry`/`perFileConfigs` from disk on each `GetAvailableContexts` call, dropping removed files/contexts and re-parsing only changed files.
> 
> Adds `refreshContextRegistry` with copy-on-write map swapping (to avoid concurrent map read/write with snapshot-style readers), seeds/updates `perFileMtimes` during init and CAPI merges, and expands tests to cover file deletion, context removal/addition, parse-failure behavior, and concurrent refresh vs snapshot iteration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43c2cf58c6591605fd17e03bfba014eba443cf1a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -706,33 +706,45 @@ func GetAvailableContexts() ([]ContextInfo, error) {
 	// startup keep showing up in the dropdown until the user
 	// restarts Radar (the "junk clusters" complaint).
 	//
-	// CRITICAL: refresh, snapshot, AND iterate must all happen under
-	// the same lock. The previous shape released the lock and then
-	// iterated, but `registry` and `fileConfigs` still pointed at the
-	// LIVE maps that a second concurrent caller could be mutating
-	// via refreshContextRegistry — Go's "fatal error: concurrent map
-	// read and map write". Hold the write lock through the snapshot,
-	// then iterate over a private []ContextInfo we've already built.
+	// refreshContextRegistry returns NEW maps when anything changes,
+	// so we publish them atomically under the write lock. Snapshot
+	// readers (SwitchContext, WriteKubeconfigForCurrentContext) take
+	// bare references under RLock and use them after the unlock — that
+	// pattern is only safe as long as the maps they captured are never
+	// mutated. Returning fresh maps preserves that invariant.
 	clientMu.Lock()
 	if contextRegistry != nil {
-		// Lazy init: MergeAndSwitchContext promotes single-file mode
-		// to isolated-load mode without touching perFileMtimes, so a
-		// CAPI promotion can leave it nil. Seeding it here is safe
-		// because we always hold the write lock.
+		// Lazy init: a future code path that promotes single-file mode
+		// to isolated-load without touching perFileMtimes would leave
+		// it nil. Seeding it here is safe because we always hold the
+		// write lock and refresh's nil guard catches it too.
 		if perFileMtimes == nil {
 			perFileMtimes = make(map[string]time.Time, len(perFileConfigs))
 		}
-		refreshContextRegistry(contextRegistry, perFileConfigs, perFileMtimes)
+		newRegistry, newFileConfigs, newFileMtimes, changed := refreshContextRegistry(
+			contextRegistry, perFileConfigs, perFileMtimes,
+		)
+		if changed {
+			contextRegistry = newRegistry
+			perFileConfigs = newFileConfigs
+			perFileMtimes = newFileMtimes
+		}
 	}
+	registry := contextRegistry
+	fileConfigs := perFileConfigs
 	currentCtx := contextName
+	clientMu.Unlock()
 
-	if contextRegistry != nil {
+	if registry != nil {
 		// Isolated-load mode: enumerate every registered context, pulling
 		// cluster/user/namespace from the file it originally lives in.
 		// No merge happens — shared names across files stay distinct.
-		contexts := make([]ContextInfo, 0, len(contextRegistry))
-		for qName, entry := range contextRegistry {
-			cfg, ok := perFileConfigs[entry.SourceFile]
+		// Iterating outside the lock is safe because refresh publishes
+		// fresh maps on change rather than mutating in place, so the
+		// snapshot we captured is frozen.
+		contexts := make([]ContextInfo, 0, len(registry))
+		for qName, entry := range registry {
+			cfg, ok := fileConfigs[entry.SourceFile]
 			if !ok {
 				continue
 			}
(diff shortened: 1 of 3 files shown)


eliran-ops:
@nadaverell @hisco — ready for review. Cursor bugbot found 6 issues across 2 review cycles: 3 HIGH (concurrent map read/write panics across `refreshContextRegistry`/`SwitchContext`/`WriteKubeconfigForCurrentContext`, nil `perFileMtimes` panic after CAPI promotion), 1 MEDIUM (refresh losing files when all contexts removed), 2 LOW (CAPI fast-path mtime, nil-mtime test). All addressed: 4 in prior commits (d9fcfde, 5f2654d) and the remaining 2 in 43c2cf5 — `refreshContextRegistry` now builds new maps and atomically swaps under the write lock instead of mutating in place, restoring the post-init immutability invariant. New regression test `TestGetAvailableContexts_ConcurrentRefreshAndSnapshotIterate` runs 4 writers + 4 snapshotters for 200 iterations under `-race`. Full `go test ./internal/k8s/... -race -count=2` passes. Cursor Bugbot SUCCESS on current HEAD.

eliran-ops:
## Visual test - no regression on dropdown

Tested locally on `gke_koalabackend_us-east1-b_nonprod-cluster-us-east1`. Multi-file mode is active here (multiple GKE projects → grouped cluster picker), so the patched `refreshContextRegistry` path is the one being hit on each `GetAvailableContexts`.

### Cluster switcher renders correctly
Multiple GKE projects, contexts grouped per file, current selection highlighted. No stale entries observed:
cluster-switcher

### Coverage caveat
The actual refresh-vs-disk behavior (file removed → contexts gone, file rewritten → re-parsed) was not exercised in this run because doing so cleanly requires a controlled multi-file kubeconfig setup with a known mutation. The PR's own test plan covers this:

- Unit tests: `internal/k8s/context_registry_test.go` (411 lines, 5 new subtests for refresh) - already executed.
- Manual test plan from PR body still has 2 unchecked items - worth running before merge:
  - `kubectl config delete-context foo` then reopen dropdown - context should disappear without restart.
  - `rm` a watched kubeconfig file then reopen dropdown - all its contexts should disappear.

### Verdict on what I could exercise
Dropdown opens, contexts list correctly, switching works. Backend `GetAvailableContexts` path on this build is the patched one. No regression on the visible surface.

eliran-ops:
### Live-cluster verification (manual test plan boxes ticked)

Built `radar` from this branch, ran with `--kubeconfig-dir /tmp/sky834-kubeconfigs/` containing:
- `file-a.yaml` — 2 contexts (`...nonprod...`, `...management-cluster`)
- `file-b.yaml` — 1 context (`...prod...`)

| Step | Disk action | `/api/contexts` | Response size |
|---|---|---|---|
| Initial | — | 3 contexts | 646B |
| `kubectl config delete-context management-cluster` (file-a) | mtime moves | 2 contexts (`nonprod` + `prod`) | 440B |
| `rm file-b.yaml` | file gone | 1 context (`nonprod`) | 225B |
| Noop re-list | none | 1 context, served in 47µs | 225B |

Both PR symptoms recover without restart. Noop re-list hits the mtime fast-path (45-100µs vs ~1ms when refresh fires) — refresh only re-parses on real disk changes. The unchanged file's context (`nonprod` in file-a) survived correctly through file-b's deletion, confirming per-file isolation is preserved.

PR #588, opened 2026-04-30
fix(search): make zero-result topology + resource search self-explanatory

eliran-ops:
Closes part of [SKY-828](https://linear.app/skyhook/issue/SKY-828).

## Why

Two SKY-828 bugs share the same shape: the user searches and gets back \"No results found\", but the surrounding UI tells them something different exists. They conclude search is broken.

## Bug 45 — Fleet topology search returns 0 hits for known pod names

Fleet view of the topology only renders CAPI kinds (Cluster, Machine, MachineSet, ...). Pods, Deployments, etc. are filtered out of the \`nodes\` prop the search reads. Searching \"3scale\" or \"billing\" returned a flat \"No resources found\" with no clue that the matches existed in another view.

**Fix.** \`TopologySearch\` gains two optional props:

- \`allNodes\` — the unfiltered topology. When present we use it to count hidden matches.
- \`viewModeLabel\` — a human-readable label for the current view (e.g. \"Fleet\").

When the visible search yields zero AND \`allNodes\` contains hidden matches, the empty-state now reads:

> No resources found for \"3scale\" in Fleet
> 
> 2 matches are hidden by the current view (Fleet). Switch view to see them.

Backwards-compatible — call sites that don't pass \`allNodes\` see no change.

## Bug 46 — Sidebar count badge looks like it's lying when search returns 0

The sidebar's count badge shows the cluster-wide total (\`resourceCounts\`) and intentionally stays unfiltered so it survives kind switches and search-state. When a search yields zero results the user sees \"5 Pods\" in the sidebar and \"No Pods found\" in the pane and concludes the badge is wrong.

**Fix.** When the empty-state is shown with an active search, append:

> The sidebar shows N Pods in the cluster — the count is unfiltered.

No sidebar API change required.

## Tests

\`packages/k8s-ui/src/components/topology/topology-search-match.test.ts\` (6 cases) pins the hidden-match-count rule:
- 0 when the visible set has any match (don't shame the visible UI)
- count of hidden-but-matching nodes otherwise
- case-insensitive
- matches by name OR kind OR namespace OR \`kind/name\` OR \`namespace/name\`
- 0 for empty/whitespace queries
- never double-counts a node that's both visible and matches

## Out of scope

**Bug 47** (Global Search header \"338 pods across 2 connected clusters\" returning 0 hits for any pod name) is in the cloud product (radar-hub-web), not this OSS repo. Searched the repo for \`across.*clusters\` / \`connected clusters\` UI strings — no matches; only backend usages.

[...]

PR #633, opened 2026-05-05
fix(search): drop the empty-search count hint

eliran-ops:
Closes [SKY-855](https://linear.app/skyhook-io/issue/SKY-855). Supersedes the original "guard the hint with `hideSidebar`" approach — see below.

## What

Removes the empty-state hint that #588 added to `ResourcesView`'s zero-result search state:

> The sidebar shows N Pods in the cluster — the count is unfiltered.

## Why drop it instead of fixing it

The hint started as a forward-fix: it references "the sidebar", which is wrong in embedded consumers that hide it (radar-hub-web's Fleet `SearchPage` passes `hideSidebar=true`). The first cut here just suppressed it when the sidebar was hidden; the second reworded it to be sidebar-independent.

But stepping back, the hint isn't worth keeping in any form:

- The empty state already shows **"No results for `<term>`"** (explains *why* the table is empty) plus a **Clear filters** button (the actionable next step).
- In Radar the **sidebar badge already shows the same per-kind total** right next to the pane — the sentence restates what's on screen.
- The number itself is **non-actionable** — knowing there are exactly N doesn't help refine a search.

So the sentence was either redundant (Radar) or noise (embeds). Removing it leaves a cleaner empty state with nothing lost.

## Change

- Delete the hint block from `ResourcesView.tsx`.
- Delete its unit test (`resources-search-sidebar-hint.test.ts`).

Net: **+0 −102**, no new branches or props. `tsc` clean.

## Test plan

- [x] `make tsc` — clean
- [x] `/visual-test` against `kind-radar-gitops-demo`: zero-result search on `/resources/pods` now shows "No Pod found" + "No results for …" + Clear filters, with no count sentence. No regression to the rest of the empty state.


nadaverell:
Reworked rather than merged as-is: instead of suppressing the hint when `hideSidebar` is set, I reworded it to drop the "sidebar" reference entirely — "The cluster has N … — none match the current filter." Same fix for the embed, no `hideSidebar` branch needed, and embedded users keep the useful "N exist, your filter just matched none" signal instead of getting nothing. Also rebased onto current main. Verified via visual-test.

nadaverell:
Update: rather than reword, dropping the hint entirely. The empty state already shows "No results for <term>" + a Clear-filters button, and in Radar the sidebar badge shows the same per-kind total right beside it — so the sentence was restating on-screen info, and the number itself isn't actionable. Net change is now a deletion of the hint + its test. (Supersedes my previous comment about the reword.)

PR #654, opened 2026-05-06
fix(prometheus): dedupe memory scrape series

subhadipb7679:
## Summary
- Deduplicate Prometheus memory scrape series before aggregating resource metrics.
- Avoid double-counting repeated scrape samples for the same pod/container memory metric.

## Test plan
- `go test ./internal/prometheus`


nadaverell (approved):
Thanks!

Verified on a test cluster it works for me as well.

PR #658, opened 2026-05-07
feat(fleet): unified issues + CEL search; fix self-upgrade helm conflict

nadaverell:
## Summary

Two changes bundled on `feature/fleet-search`:

**1. `feat(fleet)`** — adds the backend powering the unified fleet view (paired with the radar-hub-web PR):
- `internal/issues` — cluster-health rollup merging problems / audit findings / events / CRD-condition fallbacks into a 3-tier severity model (critical / warning / info), with per-cluster offline+errored+hits envelopes for FleetCoverage.
- `internal/search` — CEL-based candidate scoring across cluster resources (parse, score, activation).
- `internal/filter` — shared filter primitives.
- `internal/server` — REST handlers (`/api/fleet/issues`, `/api/fleet/search`) and matching MCP tools.

**2. `fix(self-upgrade)`** — `/agent/self-upgrade` patched the Deployment with empty `PatchOptions{}`, so the apiserver derived field manager from User-Agent → recorded as `"radar"`. After the first self-upgrade, `.image` was permanently owned by `"radar"`, and every subsequent `helm upgrade` failed with the conflict error until users ran `--force`. Patch now uses `FieldManager: "helm"` + `Force: true` so ownership stays where helm expects it. Chart README also documents the new Cloud install pattern (`kubectl create secret` → `helm install --set cloud.existingSecret`).

## Test plan
- [ ] `go test ./internal/issues/... ./internal/search/...` (tests included in commit)
- [ ] Render chart with new install pattern: `helm template ... --set cloud.existingSecret=radar-cloud-config` — verify `RADAR_CLOUD_TOKEN` sources via `secretKeyRef` (validated locally)
- [ ] Render chart with legacy `--set cloud.token=...` path — verify still works (backward compatible, validated locally)
- [ ] Self-upgrade smoke test: trigger via Cloud, then `helm upgrade` without `--force` and confirm no conflict

---

> [!NOTE]
> **Medium Risk**
> Medium risk: introduces new `/api/issues` and `/api/search` surfaces with CEL-based filtering and RBAC gating, plus changes to upgrade patch semantics that affect deployments/upgrades.
> 
> **Overview**
> Adds two new read-only APIs, **`/api/issues`** and **`/api/search`**, plus matching MCP tools, to support fleet-style health and resource discovery. `issues` composes problem checks, optional audit/event sources (opt-in), and a generic CRD condition fallback; `search` provides ranked free-text lookup (name/namespace/labels/annotations/images/kind) with per-user RBAC-based kind suppression for sensitive/cluster-scoped resources.
> 
> Introduces a shared `internal/filter` package using **CEL** for structured boolean predicates (with caching and eval-error reporting) and wires it into both endpoints.
> 
> Fixes operational issues: the Helm chart’s cloud Helm RBAC bindings now render correctly when `cloud.enabled=true` regardless of `auth.mode`, the chart README documents the cloud install pattern using an external Secret, and `/agent/self-upgrade` now patches with `FieldManager: "helm"` (with more granular error handling) to avoid Helm ownership conflicts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 987016b7ebeb0bd8f499bc44dea47c12c36114c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
Reviewer round-3 pass — two real findings:

**F1. Cloud install left helm RBAC bindings unbound** (chart side).
`deploy/helm/radar/templates/cloud-rbac-helm.yaml:6` gated on
`(ne .Values.auth.mode "none")` while sibling templates accept
`cloud.enabled` alone. Cloud installs (default \`auth.mode\`) rendered the
helm ClusterRoles but not the \`radar-cloud-{owner,member}-helm\` /
\`radar-cloud-owner-helm-admin\` bindings. Dropped the redundant check;
verified with \`helm template\` (before: only \`radar-cloud-{owner,member,
viewer}\` render; after: also the helm-tier bindings). The wizard PR
([radar-hub-web#48](https://github.com/skyhook-dev/radar-hub-web/pull/48))
also adds \`--set auth.mode=proxy\` so installs on existing chart caches
get the fix immediately.

**F2. /api/search exposed sensitive kinds to namespace-bound users.**
Added \`Options.SkipKinds\` to \`search.Options\`. New
\`internal/server/search_rbac.go\` runs parallel cluster-scope SARs
against (Secret, Node, PersistentVolume, StorageClass, Namespace) as
the END user; denied kinds get dropped from the scan — including for
explicit \`kind:Secret\` queries, which silently return zero hits (same
shape as the existing SA-forbidden lister path). In \`auth-mode=none\`
the map is nil and the SA's own RBAC stays the only filter.

Trade-off documented in \`search_rbac.go\`: a user with per-namespace
\`list secrets\` loses /api/search coverage for secrets and must hit
\`/api/resources/secrets?namespace=X\` directly. Per-namespace SAR
fanout is a follow-up if customer evidence asks for it. The cluster-
scope SAR is the right minimum for the Cloud-viewer threat model.

Coverage in \`skipkinds_test.go\`: SkipKinds drops the kind even for
explicit \`kind:Secret\`, nil map is a no-op, skipping Secret doesn't
affect Deployments / Pods.

Pushed [5d1aecb7](https://github.com/skyhook-io/radar/pull/658/commits/5d1aecb7).

nadaverell:
## Round 4 reviewer findings — addressed

**Critical 3 (RBAC leak + dup): Node problems leaked through namespace-scoped `/api/issues`**

`internal/issues/provider.go:39` — `CacheProvider.DetectProblems` looped allowed namespaces and called `k8s.DetectProblems(cache, ns)`. That function appends cluster-scoped problems (Node, etc.) regardless of the `ns` arg, so a namespace-bounded Cloud viewer:

1. **Saw Node problems they don't have `list nodes` RBAC for at cluster scope** (real RBAC leak).
2. Saw each Node problem **duplicated `len(namespaces)` times** in the response.

Fix in 7b06b94: extracted `flattenNamespacedProblems` — concats per-namespace lists while dropping empty-Namespace rows. Applied to both `DetectProblems` and `DetectCAPIProblems` (CAPI is namespaced today but defensive against future cluster-scoped CAPI emitters). The unrestricted-caller path (`namespaces == nil`) is unchanged — it still calls `k8s.DetectProblems(cache, "")` exactly once and includes cluster-scoped rows.

Three tests pin the contract (leak-strip, namespaced-row preservation across slices, empty-input safety).

The C1 (frontend build break) and C2 (cache-key tenant-isolation) findings are addressed in the parallel radar-hub-web and radar-hub PRs.

nadaverell:
## Reviewer findings — addressed in 987016b7

All three confirmed real after verification. Triage + fixes:

**CRITICAL · selfupgrade Force on StrategicMergePatch**

Verified against `k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go:199` — the apiserver explicitly forbids `Force` on non-Apply patches with `field.Forbidden(\"force\", \"may not be specified for non-apply patch\")`. The previous commit's `Force: &force` flipped self-upgrade from "broken silently" to "always 422" — caught by our `IsInvalid` branch and masked as a generic 400 "invalid patch."

Fix: drop `Force`, keep `FieldManager: "helm"`. The FieldManager alone is what fixes the original bug (apiserver no longer registering "radar" from User-Agent as the .image owner); Force was never necessary for that. Inverted the tripwire test so a future regression that puts Force back trips immediately.

Rejected the reviewer's alternative (switch to ApplyPatchType + full apply YAML) because it's significantly bigger surface — for a single-field patch, the simpler "FieldManager-only" path achieves the same outcome without rewriting the request body.

**IMPORTANT · MCP issues source=event silently empty**

Verified — `ComposeWithStats` (post-events-default-off) requires both `wantSource(SourceEvent)` AND `IncludeEvents`. The MCP tool's source-loop only flipped `IncludeAudit`, so `source: "event"` returned zero event rows while the HTTP path worked.

Fix: switch the loop to a `switch` that sets `IncludeEvents` for SourceEvent (mirroring audit). Also defaulted `Since=1h` when events are enabled with no Since param, matching the HTTP handler's behavior — without it, an MCP caller silently inherits the full event-cache backlog. Updated the `source` + `since` jsonschema descriptions so the opt-in semantics are discoverable from the tool listing.

**IMPORTANT · MCP issues total_matched missing**

Verified — HTTP `/api/issues` emits `total_matched`, MCP didn't. The field is already on `ComposeStats` so no compose-layer change needed.

Fix: emit `total_matched` in the MCP response map. Agents can now distinguish "200 returned" from "200 of 1000" the same way HTTP callers can.

Validation: `go test ./internal/filter ./internal/search ./internal/issues ./internal/server` passes locally.

PR #661, opened 2026-05-08
fix(layout): scope BottomDock and ResourceDetailDrawer to their container instead of the viewport

arlenvasconcelos:
## Summary
- Switch `BottomDock` from `fixed` to `absolute` positioning and base maximized height on `100%` of the parent.
- Apply the same fix to `ResourceDetailDrawer`: `fixed` → `absolute`, with width/height based on `100%` of the parent (covers the expanded full-screen mode too).
- Mark the App shell `relative` so both surfaces anchor to the Radar surface, not the viewport.

## Why
When Radar is consumed as an embedded library (e.g. Radar Hub), viewport-anchored positioning made the dock and the expanded resource drawer escape the Radar surface and overlap the host app's chrome. Scoping these surfaces to their container lets them size correctly inside any embedding context while remaining unchanged for the standalone app.

## Test plan
- [x] Standalone Radar: open dock, expand, maximize — verify height/position match previous behavior.
- [x] Standalone Radar: open the resource drawer, expand it to full-screen — verify it fills the Radar surface as before.
- [x] Embedded (Radar Hub): dock no longer escapes the Radar surface or overlaps host chrome.
- [x] Embedded (Radar Hub): expanded resource drawer stays scoped to the Radar surface.
- [x] Resize window with dock open at each size mode (collapsed / expanded / maximized).

### DIFF
--- a/packages/k8s-ui/src/components/dock/BottomDock.tsx
+++ b/packages/k8s-ui/src/components/dock/BottomDock.tsx
@@ -105,11 +105,11 @@ export function BottomDock({ renderTabContent, renderTabHeaderExtra, leftOffset:
     return null
   }
 
-  const effectiveHeight = !isExpanded ? 36 : isMaximized ? `calc(100vh - ${MAXIMIZED_TOP_OFFSET}px)` : height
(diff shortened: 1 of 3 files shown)


Saurabhtbj1201 (approved):

nadaverell (approved):
verified also in embedded, thanks for the fix!

PR #772, opened 2026-05-25
feat(k8score): idle reaper for namespace-scoped dynamic informers

nadaverell:
## Summary

Follow-up to #770 (**stacked on it** — base is the B2.1 branch; retarget to `main` once #770 merges). B2.1 lets a GVR have many namespace-scoped informers; over the life of a long-running, namespace-restricted deployment those can accumulate as new namespaces get watched. This adds an **idle reaper** that stops a namespace-scoped informer after `InformerIdleTTL` (default **30m**) without a read, re-creating it transparently on next access.

## Liveness = the read path (no SSE lease needed)

Every read — `List` / `Get` / `Count` / the `EnsureWatching` confirm-step — stamps the entry's `lastAccess` (atomic, off the hot lock). The frontend re-lists open views every **60–120s** (`ResourcesView.tsx` / `GitOpsView.tsx` React Query `refetchInterval`), and SSE topology rebuilds also `List` — both flow through that stamp. So an open view keeps its informer warm and **can't be reaped out from under a live stream**, which is the concern that would otherwise motivate an SSE subscriber lease. The TTL (30m) is comfortably above the poll interval, so the lease/refcount coupling is unnecessary; I went with the simpler read-driven design deliberately.

- **Cluster-wide informers are exempt** (one per GVR, already bounded, serve every namespace — churning them is pure cost). The common cluster-wide-RBAC deployment is therefore completely unaffected: nothing is ever reaped.
- A **negative `InformerIdleTTL` disables** reaping; zero selects the default.

## Why informers are now built directly (not via a factory)

A `DynamicSharedInformerFactory` caches one informer per GVR, so after the reaper stops an informer the factory would hand back the **stopped** instance on re-request — and a `SharedInformer` can't be re-`Run`. Switching to `NewFilteredDynamicInformer` gives standalone informers, each under its own cancelable context, that can be stopped and re-created freely. This also let me drop the per-namespace factory map and simplify `Stop()` (closing `stopCh` cancels every informer's context; no factory shutdown).

## Tests

`pkg/k8score`: idle namespace-scoped informer is reaped while a freshly-read one is kept and re-created on next access; cluster-wide informers are exempt even when ancient; negative TTL disables reaping. Reaper is driven with an explicit `now` for deterministic, timing-free tests. `go build ./...`, `go test ./internal/server ./internal/k8s ./pkg/k8score` all green.

## Note
Default behavior change is limited to namespace-restricted deployments (reaper trims idle per-ns informers there). Cluster-wide deployments see no behavioral change.

---

> [!NOTE]
> **Medium Risk**
> Changes core dynamic cache lifecycle and informer construction; mis-tuned TTL or missed `touch()` paths could stop watches for idle namespaces (transparent re-sync on next access). Cluster-wide RBAC deployments are explicitly exempt from reaping.
> 
> **Overview**
> **Namespace-restricted dynamic CRD watches** can accumulate one informer per namespace; this PR adds an **idle reaper** that stops namespace-scoped informers after **`InformerIdleTTL`** (default **30m** without a read) and **re-creates them on the next List/Get**. **Cluster-wide** informers are never reaped; a **negative TTL** turns reaping off.
> 
> Liveness is **read-driven**: `lastAccess` is updated atomically on list/get/count paths (including `EnsureWatching` confirm and `ListWatched` for background scanners), so active UI polling keeps informers warm without an SSE lease.
> 
> **Implementation shift:** informers are built with **`NewFilteredDynamicInformer`** instead of a shared **`DynamicSharedInformerFactory`**, so reaped/stopped informers are not reused from a factory cache. **`Stop()`** no longer shuts down factories—per-informer contexts cancel via **`stopCh`**. Sync completion only marks **`synced`** if the entry is still the one that started syncing (avoids races after a reap/re-create).
> 
> **`DynamicCacheConfig.InformerIdleTTL`** is the new knob (zero = default, negative = disabled). Tests cover idle vs fresh namespaces, cluster-wide exemption, disable flag, and transparent re-create after reap.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 721654d95aa788fc9d18ec89333ddb0558fc0278. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/audit/runner.go
+++ b/internal/audit/runner.go
@@ -93,7 +93,7 @@ func listCrossplaneDynamic(namespaces []string) (mrs, xrs []*unstructured.Unstru
 		if gvr.Group == "pkg.crossplane.io" || gvr.Group == "apiextensions.crossplane.io" {
 			continue
 		}
-		items, err := cache.List(gvr, "")
+		items, err := cache.ListWatched(gvr)
 		if err != nil {
 			if !apierrors.IsForbidden(err) && !apierrors.IsUnauthorized(err) {
 				log.Printf("[audit] Crossplane scan: skipping %s/%s: %v", gvr.GroupResource(), gvr.Version, err)
--- a/internal/k8s/policy_reports.go
+++ b/internal/k8s/policy_reports.go
@@ -340,7 +340,7 @@ func listPolicyReportsAll(gvrs []schema.GroupVersionResource) []*unstructured.Un
 	}
 	var all []*unstructured.Unstructured
 	for _, gvr := range gvrs {
-		items, err := cache.List(gvr, "")
+		items, err := cache.ListWatched(gvr)
 		if err != nil {
 			log.Printf("[policy-reports] list %s: %v", gvr, err)
 			continue
--- a/internal/k8s/topology_adapter.go
+++ b/internal/k8s/topology_adapter.go
@@ -207,6 +207,10 @@ func (a *topologyDynamicProvider) List(gvr schema.GroupVersionResource, namespac
 	return a.dynCache.List(gvr, namespace)
 }
 
+func (a *topologyDynamicProvider) ListNamespaces(gvr schema.GroupVersionResource, namespaces []string) ([]*unstructured.Unstructured, error) {
+	return a.dynCache.ListNamespaces(gvr, namespaces)
+}
+
 func (a *topologyDynamicProvider) Get(gvr schema.GroupVersionResource, namespace, name string) (*unstructured.Unstructured, error) {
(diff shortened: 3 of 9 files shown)


nadaverell:
Closing as obsolete rather than rebasing. The namespace-scoped informer growth mechanism still exists in theory, but current main's multi-namespace fanout/completeness state and event-driven derived-cache consumers make this read-idleness design unsafe. Revisit from current main if diagnostics or user reports show real resource pressure.

PR #803, opened 2026-05-27
Issues: symptom classification + owner-grouping (engine)

nadaverell:
Builds the Radar Issues engine: every operational issue is classified by symptom **category**, gets a stable **identity**, and is **grouped** under its owning workload — so `/api/issues` and the MCP `issues` tool emit a triage queue, not a per-object feed. Pure/deterministic, table-tested, MCP-first.

### What changed (3 commits)

1. **Classify** — a pure `(Source, Kind, Reason)` → `category` classifier (~25 categories → 11 groups, `unknown` first-class), wired into `Compose`. Every row carries `category` + `category_group`, both server-emitted labels and CEL filter bindings. Grounded in radar's actual reason vocabulary.
2. **Identity** — resolve each Pod problem's topmost stable controller (Pod→Deployment, not the intermediate ReplicaSet) at detection time; derive a `grouping_scope` and a deterministic cluster-local `id = hash(scope, subject key, category)`. `resourceKey` reuses `pkg/audit.ResourceKey` so issues and audit deep-links share one key format.
3. **Group** — `GroupIssues` folds the flat evidence rows into the public model: one row per `id` with `affected` counts + bounded member refs. Grouped by default on `/api/issues` + MCP; the cap now counts issue groups, not replica fan-out.

### Notes for review

- **`?view=flat`** on `/api/issues` returns the raw pre-fold rows for debugging ("what folded into this group?"). MCP stays grouped-only — agents use `get_resource`/`get_events` for raw state.
- `Compose()` stays flat internally, so `summarycontext`'s per-resource index is unchanged; a `Filters.Grouped` flag gates the fold.
- Representative rules are deterministic: severity = max member, subject = topmost owner, reason/message/crash-context from the worst member, age = oldest onset, last_seen = newest, members sorted + capped at 10 with `members_truncated`.
- Taxonomy gaps fall through to `unknown` deliberately (CronJob/Job/CAPI/PVC-Lost/Node-Cordoned, and categories whose detectors don't exist yet).

Pairs with skyhook-dev/radar-hub#52 (forwards the new fields through the fleet pivot). SPA grouped `IssuesView` is a follow-up.

### DIFF
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -11,12 +11,12 @@
 //
 // Two compile entry points:
 //
-//   CompileObjectFilter — bindings shaped to a K8s object:
-//     kind, apiVersion, metadata, spec, status, labels, annotations
+//	CompileObjectFilter — bindings shaped to a K8s object:
+//	  kind, apiVersion, metadata, spec, status, labels, annotations
 //
-//   CompileIssueFilter — bindings shaped to an issues.Issue:
-//     severity, source, kind, group, namespace, name, reason, message,
-//     count, last_seen, cluster
+//	CompileIssueFilter — bindings shaped to an issues.Issue:
+//	  severity, source, category, category_group, kind, group, ns,
+//	  name, reason, message, count, last_seen, cluster
 //
 // Both return a Filter whose Match(activation) yields (bool, error).
 // Compile errors are returned verbatim (CEL's parser produces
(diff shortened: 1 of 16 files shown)


nadaverell:
Subsumed into #811 — the classification engine (the three `issues:` commits) is now the bottom of that stack, which also adds the unified `pkg/subject` resolver, the GA-blockers, and the grouped triage UI. Consolidating to one PR per review request. Reopen if we want the engine reviewed separately.

PR #804, opened 2026-05-28
feat(gitops): clickable Repository + Path in ArgoCD Application Source

hisco:
## Summary

The Source section in the ArgoCD Application resource tray showed Repository and Path as plain text. Now both are real anchors — when we can produce a safe URL.

- **Repository** — links to `source.repoURL` exactly as the user configured (whitespace-trimmed only), provided it parses as `http:` or `https:`. SCP form (`git@host:o/r`), `ssh://`, `git+ssh://`, `oci://`, `file://`, `javascript:`, etc. all stay plain text — we don't rewrite user URLs to make them linkable. An `ExternalLink` icon trails the URL, and `title=` provides a hover tooltip for full-URL readability when `break-all` wraps long URLs.

- **Path** — deep-links to the directory at the target ref, but only when the host is a provider we know how to route into: `github.com`, `gitlab.com`, `bitbucket.org`, `dev.azure.com`, `*.visualstudio.com`. Unknown hosts fall back to plain text. Bitbucket without an explicit `targetRevision` also falls back, because Bitbucket Cloud's `/src/` endpoint doesn't accept `HEAD` as a ref token.

`targetRevision` handling: empty or `"HEAD"` (case-insensitive) means default branch — GitHub/GitLab accept the literal `HEAD` in the URL; Azure omits the version param; Bitbucket falls back to text. 40-hex SHAs (case-insensitive) use Azure's `GC` prefix; other refs use `GB`. Branch names with slashes (`feature/foo`) are preserved as literal path segments — `encodeURIComponent` would have produced `%2F` and 404s. Multi-source apps inherit the behaviour automatically via the shared `SourceProperties` component.

## Implementation

- New `packages/k8s-ui/src/utils/git-provider-urls.ts` — pure helpers `buildRepoBrowseUrl` + `buildPathBrowseUrl`, both `string | null`. Discriminated-union `ParsedRepo` type so the per-provider switch is provably exhaustive. Exported from `packages/k8s-ui/src/utils/index.ts` barrel for library consumers.
- `SourceProperties` in `ArgoApplicationRenderer.tsx` wraps the existing `Property` rows with conditional `<a>` anchors. Reuses the existing link styling pattern from `drawer-components.tsx` `ExternalLinksSection` (`text-blue-400 hover:text-blue-300 hover:underline break-all`). Anchors carry `target="_blank"`, `rel="noopener noreferrer"`, and `title=` for hover tooltip.

## Test plan

- [x] 50 vitest unit cases covering each provider, http-only scheme guard, SCP/ssh/oci/file/javascript rejection, `.git` suffix, HEAD/empty/lowercase-head fallback, SHA (lower/upper) vs branch refs on Azure, nested GitLab subgroups, percent-encoded Azure segments (no double-encoding), slash branches, Bitbucket explicit-ref requirement, too-few-segments downgrades, Azure `_git` index boundaries, non-default ports, custom hostnames.
- [x] Full k8s-ui suite: 462/462 pass.
- [x] `tsc` clean on both `k8s-ui` and `radar-app`.
- [x] CI all green on `402af2f`, Cursor Bugbot pass (no findings on the latest commit).
- [x] Visual test against live cluster (`argocd/keda-nonprod-cluster-us-east1`): Repository renders as anchor with external-link icon; Path links to `github.com/KoalaOps/deployment/tree/HEAD/argocd/addons/keda/keda/nonprod-cluster-us-east1`; layout holds at 1280 and 1920 widths; no new console errors.

---

> [!NOTE]
> **Low Risk**
> UI-only linking with http(s) validation and null fallbacks; no auth or cluster API changes. Main risk is occasional wrong deep links for unusual repo URL shapes.
> 
> **Overview**
> ArgoCD Application **Source** rows now render **Repository** and **Path** as external links when safe URLs can be built, instead of plain text only.
> 
> New **`git-provider-urls`** helpers power this: **`buildRepoBrowseUrl`** links `http`/`https` `repoURL` as configured (trimmed only); non-HTTP schemes and invalid input stay text. **`buildPathBrowseUrl`** deep-links the app path for GitHub, GitLab, Bitbucket, and Azure DevOps (including legacy `visualstudio.com`), with provider-specific ref/path encoding and fallbacks (e.g. no path link for unknown hosts, Bitbucket without an explicit ref, or Helm chart sources).
> 
> **`SourceProperties`** in **`ArgoApplicationRenderer`** uses those helpers, adds an external-link icon on repository links, and skips linking for Helm chart registry URLs. Helpers are exported from the k8s-ui utils barrel; vitest covers URL building and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbf3d403130f9492c77d6178640cecfb4a3c6d99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
Claude review results (after triaging to remove noise):

The PR is solid — URL logic is correct across all four providers, security is airtight (`new URL` + http(s) gate rejects `javascript:`/`file:`/`oci:`; credentials stripped via `url.host`; per-segment encoding prevents query/fragment breakout), and comments are accurate. No blocking issues. Six findings worth addressing:

**Type design**
- `ParsedRepo.segments` is a flat optional bag (5 independent `?` fields) for what are really 3 disjoint shapes. The symptom is the three `!` assertions (`azureOrg!` etc.) at `git-provider-urls.ts:156`. A discriminated union on `provider` removes all three `!`, deletes the dead `return null` at `:159` (switch becomes provably exhaustive), and costs ~5 lines. It's internal-only, so zero compat cost — genuinely cheaper than the status quo.

**Helm sources**
- For Helm chart sources, `repoURL` is a chart registry, not a browsable git repo, so the new Repository anchor links to a 404/index page. The renderer already knows it's Helm (`source.chart` at `ArgoApplicationRenderer.tsx:77`). One-line gate: `const repoHref = source.chart ? null : buildRepoBrowseUrl(source.repoURL)`.

**Test gaps** (the riskiest branches — the segment/index arithmetic — are currently untested):
- Known-host-but-too-few-segments downgrade (`:62`, `:72`) — e.g. `https://github.com/onlyowner` silently yields no path link. Untested for all three providers; an off-by-one in `< 2` would go undetected.
- Azure `_git`-index boundaries — the `< 2` (dev.azure.com) vs `< 1` (visualstudio.com) distinction, missing `_git`, and `_git`-as-last-segment are all unpinned.
- Lowercase `"head"` (`:134` uses `.toUpperCase()`) — prime "simplify to `=== 'HEAD'`" regression target.
- Short-hex ref staying `GB` not `GC` (`:15` `SHA_RE` requires exactly 40 chars) — loosening it would mis-prefix Azure refs and 404, undetected.

**Triaged out as noise** (not worth fixing): unknown single-segment links (malformed repoURL is effectively can't-happen; gating would break legit self-hosted root repos), Azure SSH `ssh.dev.azure.com` dead link (uncommon; proper fix adds real complexity), the hardcoded `text-blue-400` (matches the existing `ExternalLinksSection` link convention in the same file), and the "copy button regression" (the old plain-text rows never had a copy button either).

nadaverell (approved):
Reviewed in depth + visual-tested against a live ArgoCD cluster (kind-radar-gitops-demo).

- URL logic correct across all four providers; security airtight (new URL + http(s) gate, per-segment encoding, no SCP/ssh/oci rewrite).
- All 6 findings from the earlier triaged review are addressed (discriminated union, Helm-source gate, boundary tests for segment/index arithmetic, lowercase head, short-hex GB-vs-GC).
- 50/50 vitest pass, tsc clean.
- Visual test: Repository renders as anchor + external-link icon; Path deep-links to tree/HEAD/<path>; layout holds at 1920; dark mode contrast good; no new console errors.

Fallback everywhere is the old plain-text behavior, so no regression risk. LGTM.

Issue #57, opened 2026-02-01
Is there a way to exclude namespaces? Like kube-system.


kamal-shaker:
## Problem statement

Is there a way to exclude namespaces? Like kube-system for example/

## Proposed solution

In the namespaces dropdown box, maybe be able to select all/none and the pick ones to exclude/include.

Currently it's either "All" or one.

PR #811, opened 2026-05-28
Issues: classification engine + unified subject resolver + grouped triage UI

nadaverell:
End-to-end Issues overhaul in one PR: detection engine, identity layer, grouped triage UI, and the foundation hardening that came out of review.

## Engine (`internal/issues`, `internal/k8s`)
- Classify each Issue by symptom category + group; compose across the four detection sources.
- Owner-resolved, deterministic grouping identity (stable `issue_id` keyed on subject + category).
- Fold flat rows into the grouped issue model.

## Unified subject resolver (`pkg/subject`) — new
- One resolver for **identity** and **app-grouping**: Tier-1 `Subject` (owner-collapsed controller, derived from ownerRefs, zero setup) + Tier-2 `AppOverlay` (8-tier declared-key overlay — Flux / Argo / Helm / `app.kubernetes.io/*` — with provenance, confidence, and retained `conflicts[]`).
- `internal/issues/identity.go` consumes it; `StableID` is byte-identical to the prior hash (no re-key).
- The package doc is honest about scope: it's the shared identity **vocabulary** consumed by Issues today; `pkg/topology` consumption is **staged for #823, not yet wired** (AppOverlay ships as tested-but-unconsumed scaffold for that work).

## GA-blockers
- **Monotonic crashloop** classification — key on `RestartCount` + `LastTerminationState` so oscillation no longer churns `issue_id`.
- **Double-row suppression** — severity-gated, so a critical parent is never downgraded behind a warning child.
- **Critical-into-unknown gaps** closed — PVC Lost, Job / CronJob failures (and CAPI + bad-image/container-create, below).
- **CRD-condition noise floor** — transient-reason aware.

## UI (`packages/k8s-ui` IssuesView)
- Grouped, single-open-accordion triage queue; detector reason surfaced on the collapsed row.
- Stable ordering keyed on **`first_seen`** (onset) on both server and UI — `last_seen` churns to compose-time every poll, so sorting on it reshuffles rows under auto-refresh. Stable row identity keeps the expanded card from dropping on refetch.
- Onset **age column** (chronic-vs-acute signal) + "Started X ago · last seen Y ago" in the expanded body.
- `group` / `namespace` optional on the `Issue` type to match the backend `omitempty` wire.

## Hosts
- **Radar OSS** mounts the shared `IssuesView` at a per-cluster `/issues` route (thin `IssuesPane` host over a `useIssues` → `/api/issues` hook), mirroring the hub's fleet ProblemsPage. Reachable but **not yet linked in the nav** — deliberate, so the surface lands before it's promoted.

## Detection precision/recall — validated against real clusters
An empirical audit (GKE / EKS / kind) drove a precision/recall pass, each fix re-verified live:
- **GitOps reconciler health** — new `DetectGitOpsProblems` surfaces ArgoCD Application (health Degraded/Missing, sync OutOfSync when automated, ComparisonError) + Flux Kustomization/HelmRelease (Ready=False, non-transient) — a class the generic CRD-condition fallback structurally can't read. gitops-demo went **0 → 7** GitOps issues. Flux **source** CRDs (GitRepository et al.) and Argo control-plane CRDs classify as `operator_condition_failed`, not force-fit into `gitops_sync_failed`.
- **Fewer false positives** — recovered-after-crash pods no longer flagged; completed-Job pods skipped by missing-ref detection; scaled-to-0-backed Services labeled as such (including mid 1→0 scale-down); cadence-aware CronJob staleness.
- **Severity calibration** — inert dangling refs (singleton-StatefulSet headless Service, deprecated GKE PSP RoleBindings) dropped from blanket critical.
- **New `high_restart` category** for genuine thrash that isn't a classic CrashLoopBackOff.

## Detection/classification gaps closed (second review pass)
- **CAPI now triageable** — Cluster / KubeadmControlPlane → `control_plane_not_ready`; Machine / MachineDeployment / MachineHealthCheck → `machine_not_ready` (gated on `cluster.x-k8s.io`). Detection already existed; only the last classification hop was missing, so a control-plane outage was un-triageable.
- **Bad-image-tag / container-create failures detected** — `InvalidImageName` / `RunContainerError` / `CreateContainerError` / `ImageInspectError` (shared `isFatalWaitingReason`; classified in lockstep). These were silently healthy before — `InvalidImageName` never self-resolves.
- **StatefulSet** compares ready vs *desired* (spec), not just created (status.replicas), so an ordinal-0-wedged rollout surfaces.
- **`Count`** is the subject-excluded fan-out size (matches the UI "Affected resources (N)" header + TS), not `len(members)`.

## Contract hardening (review)
- **Grouped filtering on the public shape** — Compose folds to grouped *before* applying severity/kind + CEL, so `kind=Deployment` matches a pod-evidenced Deployment issue and `count > N` sees the member total.
- **RBAC-safe CRD namespace fanout** — cluster-scoped CRDs use the gated cluster-wide list; explicit namespace sets list per-namespace; namespace-scoped informers are no longer silently dropped.
- **`info` excluded from the live queue** — severity normalized to **critical|warning** (info stays honest at the Problem layer but isn't "what's broken now").
- **Owner grouping consistent** — Argo `automated.enabled:false` treated as manual; cache-aware pod owner resolution (no phantom Deployments, in Issues *and* `/top`); pod missing-ref issues fold under their workload.
- **CEL** — added `first_seen`, `grouping_scope`, `restart_count`, `last_terminated_reason` bindings (`first_seen` is the queryable onset axis; `last_seen` is near-useless for "older than…").

## Foundation hardening (review-driven)
- Detector output type `Problem` → `Detection`; the detector layer (`problems.go`) split into `detection.go` / `capi.go` / `gitops.go` with explicit layer framing.
- Identity/status ownership moved off `pkg/audit` / `pkg/packages` onto **neutral leaf packages**: `pkg/resourceid` (`ResourceKey` + builtin Kind→Group) and `pkg/conditions` (transient-reason vocabulary + one shared `FindFalseCondition`). Future Applications/Packages surfaces depend on the leaves, not on audit/packages.
- `issues.go` god-file split by concern (`compose` / `source_conditions` / `normalize` / `dedupe` / `filters`); `Group` → `CategoryGroup`; single shared `SeverityRank` (killed 3 clones); `navigateToResourceList` deduped in `App.tsx`.

## Cross-repo dependency
Merge alongside **[radar-hub #57](https://github.com/skyhook-dev/radar-hub/pull/57)** — mirrors the new CEL bindings into the hub's filter env. Additive (no removed/renamed bindings), but without it the hub's pre-validation would reject fleet-issue filters that reference the new fields.

## Precision + footgun detection (later commits)
A benchmark/empirical-driven hardening pass extended the engine with config-level root-cause detectors and precision fixes: faster clearing of recovered crashloops, multi-replica ReadWriteOnce volume conflicts, eviction-blocking PodDisruptionBudgets, rollout-deadlock root causes, and scheduler-verdict ordering by blast radius. Per-resource drill-down (`get_resource` / `diagnose`) was made owner-aware and uncapped so an object's issues agree whether you enter from the list or the resource, and classification was tightened (CAPI provider-CRD fallback, namespace-scoped CRD handling, Flux/Argo matching).

---
*Deferred follow-ups (agreed in review):*
- *Cross-subject incident correlation / blast-radius ranking (one ConfigMap → N workloads as one incident; PVC↔pod cross-layer dedup).*
- *`Provider` → per-source interfaces (R2) — defer until the source set stabilizes.*
- *`Affected` fixed struct → kind-keyed map (R10) — API debt; bundle with an IssueGroup/IssueEvidence DTO split.*
- *Frontend queue-primitive extraction shared with Checks (B2/B6/B8).*

---

> [!NOTE]
> **High Risk**
> Large refactor of the live issues compose path, grouping identity, and CEL contract (including removed `cluster` binding); behavior changes affect API/MCP consumers, sorting (`first_seen`), and what rows appear after dedup and noise suppression.
> 
> **Overview**
> This PR **restructures the issues pipeline** from a flat compose path into classify → dedupe → optional **grouped fold** → filters → CEL, and splits the old monolithic `issues.go` into focused modules (`compose`, `grouping`, `category`, `dedupe`, `source_conditions`, etc.).
> 
> **Symptom taxonomy:** Each row gets a derived **`category`** and **`category_group`** via a new `Classify()` mapper over source/kind/reason (GitOps, CAPI, batch, storage, and pod failure shapes). **`enrichIdentity`** assigns stable **`id`**, **`grouping_scope`**, and owner-collapsed subjects via **`pkg/subject`**, with **fingerprints** so distinct missing-ref causes on the same workload do not collapse.
> 
> **Grouped public model:** `GroupIssues` folds replica fan-out into one row per subject+category (**`count`** = non-subject members, inline **`members`** capped). **`Filters.Grouped`** applies severity/kind/CEL **after** folding so `kind=Deployment` and `count > N` match the triage shape. **Dedup** prefers scheduling over generic pod problems and drops **`workload_degraded` / `rollout_stalled`** when an equal-or-higher-severity child symptom exists.
> 
> **CEL / wire:** Issue filters gain **`category`**, **`category_group`**, **`first_seen`**, **`grouping_scope`**, **`restart_count`**, **`last_terminated_reason`**; **`cluster`** is removed from bindings. **Info**-severity detections are dropped at compose; issues stay **critical|warning** only.
> 
> **Detection layer:** `k8s.Problem` becomes **`Detection`** with richer pod owner, restart, and fingerprint fields; **`DetectGitOpsProblems`** is composed as its own source. Generic CRD conditions use **`pkg/conditions`**, transient/suspend/observedGeneration noise floor, kind-specific curated skips (CAPI core vs provider CRDs), and **`ListDynamicAllNamespaces`** for cluster-wide namespaced CRD scans. **`detect.go`** adds precision fixes (RWO multi-replica conflicts, PDB eviction blocks, WaitForFirstConsumer PVC suppression, service scale-to-zero labeling, deployment rollout vs degraded dedup at detection).
> 
> **Shared packages:** **`pkg/resourceid`**, **`pkg/conditions`**, and **`pkg/subject`** centralize resource keys, condition reading, and stable IDs so issues/topology/audit do not drift.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 518e8d4a7812c16fd037780cc19f17482a5b27d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
Follow-up `3512f39`: the detector reason/message now rides the **collapsed** row (was expand-only), filling the dead band between the title and the severity badge so the highest-value triage signal reads without a click. Full text + crash context stay in the expanded body. Also makes `Issue.group`/`namespace` optional to match the omitempty wire.

Consumer: the hub fleet Issues page lands in radar-hub-web #85.

PR #823, opened 2026-05-29
feat: Applications backend + package overlay data

nadaverell:
Adds the per-cluster Applications backend surface on top of the shared subject/app-overlay model.

Current scope:

- Exposes `/api/applications`, grouping concrete workloads into deployable app/release rows using topology structural roots plus the `pkg/subject` Tier-2 overlay.
- Rows carry provenance tier/confidence, image-tag versions, health, warning events, relationships, and workload-scoped debug inputs.
- Includes `workload_class` (`service`, `worker`, `job`, `unknown`) on rows and workloads.
- Treats Jobs and CronJobs as first-class application/release units.
- Treats add-on as classification evidence, not identity: mixed app/add-on groups report `category=mixed` instead of forcing the whole app to `addon`.
- Keeps system-namespace workloads visible; classification/UI can fold or badge them, but the raw row remains present.
- Threads app-overlay metadata through `/api/packages` as data-layer support for Applications and future package/add-on UI iteration.

This PR does not add or require package UI changes.

## Grouping-accuracy fixes

Two over-merge fixes from pressure-testing the model against the full topology edge inventory:

- **`structuralRoot` stops at GitOps managers.** The `EdgeManages` walk had no kind discrimination, so it climbed past a Flux Kustomization through the `GitRepository→Kustomization` source-ref edge (also an `EdgeManages` edge) to the GitRepository — collapsing a whole Flux mono-repo (one GitRepository, N Kustomizations) into one app via the structural union-find atom. It now halts at the first in-cluster manager (Application/Kustomization/HelmRelease); ownerRef chains + operator CRs still climb to the topmost owner. Regression test added.
- **`app.kubernetes.io/instance` added as overlay tier 6 (medium confidence).** It's Argo CD's *default* tracking label and Helm's install-identity label — its absence meant default-config Argo apps fell through to the semantic name/part-of tiers and two releases of one chart (identical `app.kubernetes.io/name`) over-merged. Placed above name/part-of, below the explicit Helm-release annotation (tiers renumbered to 9). Env-specific, so it stays a single-cluster overlay signal and is excluded as a cross-cluster join key in the Hub.

**Lockstep:** the tier renumber is a wire contract — the SPA's `TIER_PROVENANCE` / `sourceOf` consumers are updated in radar-hub-web #92; merge/deploy together.

Consumed by radar-hub #55 and radar-hub-web #92.

### DIFF
--- a/internal/server/applications.go
+++ b/internal/server/applications.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"net/http"
+	"slices"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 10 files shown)


nadaverell:
Superseded. The Applications backend + `pkg/subject` Tier-2 app overlay landed on `main` through #875 (app topology and identities) and #958 (declared-only app identity), with later health refactors (#1024/#1032) touching the same files. The `app.kubernetes.io/instance` overlay tier shipped as `TierInstance Tier = 6` (kept at 6 rather than the renumber-to-9 proposed here), and `workload_class` / the `/api/applications` surface are all present on `main`. This branch is stale and conflicting; closing.

PR #775, opened 2026-05-25
feat: pod-startup diagnostics + make issues mean "what's broken right now"

nadaverell:
Two coupled changes. (1) Makes **pod-startup failures** a first-class signal — radar's biggest blind spot. (2) Tightens **`issues`** into a single curated "what's broken right now" stream and stops exporting detector taxonomy as an API/agent knob.

## 1. Pod-startup diagnostics
Unlike state problems (CrashLoop, OOM) or reference problems (missing PVC), the *why* of a pod that can't reach Running is diffuse — spread across the Pod's `PodScheduled` condition, a sibling ReplicaSet's `FailedCreate`, and the join between a pod's constraints and the fleet's node labels. The scheduler already did the analysis; it just hands it back as one opaque string. We parse it and (the differentiator) **resolve "didn't match node affinity/selector" to the specific offending label** by joining the node cache:

> Unschedulable: **no node has kubernetes.io/arch=arm64** — 2 node(s) carry [amd64] (0/2 available)

Three layers (`internal/k8s/scheduling.go`, ~30 tests): **bind-time** (`PodScheduled=False` + node-fit resolver; `SchedulingGated` correctly excluded), **admission** (controller `FailedCreate`: quota / LimitRange / PodSecurity / webhook — the layer with no Pod to inspect, latest-blocker-wins dedup), **post-bind** (`ContainerCreating` → CNI IP-exhaustion / volume attach-mount, latest-event-wins).

Plumbed end-to-end: `ResourceQuota` added to the typed informer cache (browsable, RBAC-probed) **and to the Helm ClusterRole's read-only grant** (self-hosted/Cloud installs need this on upgrade or the quota section silently no-ops); `scheduling` issue source; MCP `issues` + `diagnose` (`startupBlockers` section) + dashboard; UI Pod banner / Namespace quota bars / topology tooltips.

## 2. `issues` = "what's broken right now"
`issues` is now one curated operational stream (problems + dangling refs + pod-startup blockers + False CRD conditions), severity-ranked. Detection provenance is no longer a filter axis:
- **Dropped `source=`** from `/api/issues` + the MCP tool — `source` survives only as an output label + CEL binding.
- **Removed `event` + `kyverno`** from composition (and `include_events`/`include_kyverno`). Raw events → `get_events`/timeline; policy posture → `get_cluster_audit`.
- **Removed proactive quota saturation** from the stream — it's namespace capacity context, not a live failure (reactive `FailedCreate` + the Namespace quota UI still cover it).
[...]

PR #826, opened 2026-05-29
feat: scheduling-blocker + admission-failure detection (ResourceQuota / LimitRange / PodSecurity / webhook)

nadaverell:
Surfaces admission-time and scheduling-time pod-template rejections as a first-class failure class in issues / diagnose / dashboard, alongside a typed `ResourceQuotas` cache so quota-blocked pods are reachable by name.

This branch was originally split out of #775 to keep that PR reviewable; the scheduling half is the work that didn't come along for the ride. Picking it up now because the SREGym bench exposed `namespace_memory_limit` as the one scenario both arms (kubectl + radar) failed twice — neither tool has anywhere to look for "the apiserver is rejecting this Deployment's pod template," and the upcoming bench batch has five more admission-layer scenarios in the same shape (`pvc_claim_mismatch`, `taint_no_toleration`, `service_port_conflict`, `persistent_volume_affinity_violation`, `resource_request_too_large`).

### Engine (`internal/k8s/scheduling.go`, `internal/issues`)
- `DetectAdmissionProblems` — reads controller `FailedCreate` events and classifies the message into `exceeded quota`, `LimitRange`, `PodSecurity`, or `webhook`. Bounded by a 30-minute "still happening" window (the controller re-emits continuously while stuck), one row per blocked owner-collapsed subject, latest-blocker semantics so a quota-cleared-then-webhook-rejected sequence shows the current cause not the first-seen one.
- Scheduling source for the existing issues pipeline, severity-tagged, dedup'd against runtime-pod issues on the same owner so a `workload_degraded` row isn't surfaced alongside the admission-failure row that explains it.
- Deliberately reactive: a quota that's merely saturated is *not* surfaced proactively — namespace capacity context is not a live failure. Only quota-rejected pod creates count.

### Typed cache (`pkg/k8score`)
- `ResourceQuotas` lister + capability bit + RBAC, so consumers (engine, diagnose, REST) can read quotas by name without going through the dynamic client every time.
- `internal/k8s/fetch.go` quota-fetch paths so the resource is addressable by `get_resource` etc.

### MCP / REST / UI
- `feat(mcp): surface scheduling in issues / diagnose / dashboard` — scheduling blockers show up in the MCP tools agents use, with the blocker's reason and the parsed quota/LimitRange/webhook detail in the diagnose narrative.
- `feat(ui): surface scheduling root cause in pod / namespace / topology` — pods stuck Pending render the blocker reason in the renderer chrome; namespace view surfaces active quota saturation as context.
- Dashboard REST endpoint includes scheduling rows.

### Tests
- `scheduling_test.go` covers Job/DaemonSet admission cross-check, quota boundary, post-bind event handling, dedup semantics (latest-blocker, owner-collapsed).
- Issues engine integration tests around scheduling source.

## State and what's needed before merge

- **Branch is 28 commits behind \`main\`.** A merge or rebase is needed; the branch's prior pattern was `Merge origin/main`. I did not refresh it before pushing because that touches code I haven't reviewed and we wanted it visible in origin first.
- Last two commits are `fix(review):` responses to earlier feedback (Bugbot + manual) — those review threads were on the pre-split state; current reviewers may want a fresh pass.
- Risk profile is similar to #811: it changes the shape of issues / diagnose output for a new failure class, but it's reactive (event-window-bounded) so it can't reintroduce noise in clusters that don't have active `FailedCreate` events.

## Why this is the right priority right now

`radar-bench-sregym/radar-improvement-ideas.md` (next to this checkout, under `~/sky/ws3/`) flagged this as the highest impact-to-effort gap from the N=2 bench: the running radar build is runtime-pod-centric and has no admission-layer signal, which is exactly the failure class neither tool cracked in the bench. The work is already done; it just needed to come back out of the drawer.


nadaverell:
Closing — investigation showed this work is already in `main`.

`internal/k8s/scheduling.go` on this branch is byte-identical to `main`'s, and `main`'s version has been there since `fab0799` (PR #775, merged 2026-05-26). All the admission-detection symbols I claimed were "unmerged" (`DetectAdmissionProblems`, `detectAdmissionFailures`, `classifyAdmissionFailure`, `admissionFailureWindow`) are in `main` and verified present in our SREGym bench daemon binary (`vcs.revision=5598d94`, built 2026-05-27).

My earlier check used `git merge-base --is-ancestor <branch-sha> HEAD` to test "is this work merged" — that returned "no" because #775 squash-merged and the original commit SHAs didn't survive into `main`. The content was there the whole time. I should have compared file contents, not commit SHAs.

Net result of cherry-picking this branch onto current `main`: would *roll back* ~12 merged PRs (certificates inventory, metrics, audit raw, chart RBAC coverage tests, `internal/mcp/tools.go` slim-down, etc.) for no benefit, since the scheduling work is already there. The one branch-only file (`packages/k8s-ui/src/components/resources/resources-search-sidebar-hint.test.ts`) tests a hint UI that #633 explicitly dropped.

Real follow-up is on the SREGym bench side: `namespace_memory_limit` failed on both arms in two runs even though the admission detector *is* in the binary. Investigating directly whether the scenario produces `FailedCreate … exceeded quota` events the detector can classify, vs whether it's a surfacing-into-MCP-tools gap.

PR #839, opened 2026-05-31
Add namespace-scoped informer mode

### LINKED ISSUE #684: feat: add --namespace-scope flag to force namespace-scoped informers on large clusters
## Problem

On very large clusters (~54,000 estimated topology nodes), Radar gets OOM-killed shortly after startup. The informer cache boots cluster-wide regardless of the `--namespace` flag, loading all resources into memory before the topology builder even runs.

The `--namespace` flag currently only sets a RBAC fallback namespace and a UI filter — the informers still cache every resource cluster-wide (confirmed by reading `internal/k8s/capabilities.go` L723: "the cache always boots cluster-wide and per-user view filtering happens at the HTTP layer").

The large-cluster optimizations in `topology/builder.go` (hiding ConfigMaps/PVCs, aggressive pod grouping, requiring a namespace filter for topology) help the frontend, but the backend still holds the full cluster in memory.

## Observed behavior

INFO [topology] Large cluster detected (53977 estimated nodes >= 1000 threshold), applying optimizations SSE watcher: warmup complete (46087 resources), switching to 5s debounce Killed

Peak memory grows beyond available RAM during informer sync and the process is killed by the OOM killer before the UI is even usable.

## Proposed solution

Add a flag (e.g. `--namespace-scope` or `--watch-namespaces`) that forces **all namespaceable informers** to use `informers.WithNamespace(...)` instead of cluster-wide watches — even when RBAC allows cluster-wide access.

This would:
1. Drastically reduce memory footprint on large clusters by only caching the namespaces the user cares about
2. Allow Radar to be usable on clusters with tens of thousands of resources
3. Be opt-in, so existing behavior is unchanged for small/medium clusters

The infrastructure for namespace-scoped factories already exists in `pkg/k8score/cache.go` (`nsFactories` / `informers.WithNamespace`), it's just gated behind RBAC probe failures today.

## Workaround

Currently the only workaround is to create a kubeconfig with a ServiceAccount that has **no** cluster-wide list permission, so the RBAC probe falls back to namespace-scoped informers. This works but is cumbersome.

## Environment

- Radar version: latest
- Cluster: ~54,000 topology nodes, ~46,000 cached resources
- OS: Linux (WSL2)

nadaverell:
Adds an opt-in `--namespace-scope` mode for large clusters where watching every namespace cluster-wide makes Radar's informer caches too heavy (memory pressure, slow first paint). When enabled, namespaced informers are pinned to a single namespace while cluster-scoped resources (Nodes, Namespaces, PVs, StorageClasses, CRDs without a namespace) continue to load cluster-wide. It's an operator escape hatch — off by default; the existing namespace picker keeps acting as a per-user view filter when the flag isn't set.

Fixes #684.

## What changed

**Scope resolution.** The pinned namespace is resolved by precedence: a runtime rescope override → the startup `--namespace` (only while on the context it was set for) → the kubeconfig context namespace → a saved local single-namespace pick. `--namespace` is treated as an *initial* value, not a permanent pin: after a cross-cluster switch the new context's own namespace takes over instead of dragging the old value along.

**Per-kind pinning.** RBAC probes, typed informers, and dynamic/CRD informers all honor the pin for namespaced kinds. Cluster-scoped kinds — including cluster-scoped CRDs and their size-gating count probe — stay cluster-wide (a namespaced list of a cluster-scoped resource returns nothing, so pinning them would blank them out).

**Local rescope vs. multi-user lock.** In local/no-auth sessions, picking a different namespace rebuilds the scoped cache in place (`PerformNamespaceRescope`) and the pick persists per kubeconfig context, so it survives reconnects and restarts. In auth/proxy/OIDC/cloud mode the picker is locked to the process cache scope — a different pick returns 403 — so one user cannot reshape the shared backend cache for everyone. Per-user reads (REST + MCP + Helm) are clamped to the pinned namespace; an out-of-scope request resolves to an empty view, never a widened one.

**Context-switch / rescope lifecycle.** The destructive operations (context switch, reconnect, CAPI connect, namespace rescope) are serialized under one lock so two concurrent requests can't interleave cache teardown/rebuild. A switch validates the new context has a usable scope target *before* any teardown — if it doesn't, the request is rejected as a typed preflight error and the current connection is left intact (no false "disconnected"). Active port-forward/exec sessions are stopped only at the point of no return, so a switch/rescope that fails before teardown leaves them running. The saved pick, the live cache scope, and the rescope move as one committed unit: the pick is validated and persisted before the rebuild and re-synced to the live scope if the rebuild fails, so a later reconnect can never restore a namespace the cache isn't actually serving.

**API / UI.** The namespace-scope response exposes `cacheScoped`, `cacheScopeNamespace`, and `namespaceRescope`. When scoped, the picker becomes single-select and explains, in plain language, that Radar is watching one namespace to stay fast on large clusters, that re-pointing rebuilds the cache (takes a moment, closes open terminals), and that auth/cloud instances are locked. A rescope POST uses a longer client timeout (a cache rebuild far outlasts a view-filter change) and refetches data once the new cache is live.

### Unit / static
- `go test ./internal/k8s/... ./internal/server/... ./internal/app/... ./pkg/k8score/...`
- `go test -race ./internal/k8s/...` on the context-switch/rescope paths
- `cd web && npx tsc --noEmit`
- visual-test: skipped — the only UI change is picker copy/affordances that render only under `--namespace-scope`; no layout change.

### Live verification (kind, real EKS, real GKE-on-IAM), driven via the HTTP API + server logs
- **Cache pins:** scoped to one namespace, only that namespace's resources are cached; other namespaces return empty. Cluster-scoped resources stay cluster-wide — Node counts served by the scoped cache match `kubectl` on multi-node EKS (2) and GKE (9).
- **CRD scope:** namespaced CRDs pinned (Argo `Application`); cluster-scoped CRDs served cluster-wide (Crossplane `Composition`/`CompositeResourceDefinition`).
- **Probe-based RBAC gating across authorizers:** correct under standard RBAC (EKS — denies the kinds the identity can't list) and under a GKE IAM/webhook authorizer (admits all listable kinds) — the case where `SelfSubjectAccessReview` and real list access can disagree.
- **Multi-user lock:** with auth enabled, a cluster-admin user's attempt to rescope to a different namespace is rejected (403) and the shared cache is left unchanged.
- **Connection safety:** a context switch to a context with no usable scope target is rejected before any teardown and leaves the current cluster connected.
- **Rescope:** rebuilds the cache in place against real clusters (~1.1–1.4s) and persists the pick per context.
- **Startup:** a multi/invalid namespace under `--namespace-scope` is rejected with a clear error.
- **Memory:** the cluster-wide cache shows a higher initial-sync peak than a scoped cache; at steady state, memory is dominated by the Go runtime + the cluster-scoped informers both modes run, so the saving scales with cluster size and is most significant on the large clusters this flag targets (negligible on small/mid clusters).

## Notes / limitations

- **Single namespace only.** The cache pins to exactly one namespace; scoping to several at once is not supported yet (a client-go informer watches one namespace or all, so N namespaces would mean N informers per kind across the whole typed cache). A multi/invalid target (e.g. `--namespace=a,b`) is rejected at startup with a clear error. Scoping to a small set of namespaces is the most likely follow-up; the dynamic/CRD cache, persistence, and per-user read filtering are already list-shaped, so the remaining work is the typed cache (per-(kind, namespace) informers with unioned reads) plus widening the scope target to a list.
- Under `--namespace-scope`, a resource that exists outside the pinned namespace isn't in the cache, so a direct GET/bookmark to it returns not-found — expected for a deliberately scoped instance, but a clearer "outside Radar's current scope" signal on that path is a possible follow-up.

---

> [!NOTE]
> **Medium Risk**
> Changes core informer lifecycle, context switching, and shared-cache behavior in auth mode; mistakes could leave users disconnected or serving wrong namespace data, but preflight guards, rollback, and extensive tests mitigate this.
> 
> **Overview**
> Adds **`--namespace-scope`**, an opt-in mode that pins **namespaced** informer caches to **one** namespace (cluster-scoped kinds stay cluster-wide) to cut memory and sync cost on large clusters.
> 
> **Scope resolution** uses runtime override → startup `--namespace` (only on the context it was set for) → kubeconfig context namespace → saved local pick. Startup rejects empty, multi-namespace, or invalid targets. Auth/cloud skips local settings and locks the shared cache to the startup namespace.
> 
> **Local rescope** (`PerformNamespaceRescope`) rebuilds caches when the picker changes namespace; persisted picks stay aligned with the live scope on failure. **Context switch / rescope** are serialized, preflight-validated (`ErrContextSwitchPreflight`), and stop port-forward/exec only at teardown commit; session stopping moved out of HTTP handlers into k8s.
> 
> **API/UI**: namespace-scope response adds `cacheScoped`, `cacheScopeNamespace`, `namespaceRescope`; reads and Helm clamp to the pinned namespace. Picker becomes single-select with rescope messaging and a longer client timeout during cache rebuilds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26a13e2920769c1ede74abbe17b2cb72cf838a9f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ radar
 | `--kubeconfig` | `~/.kube/config` | Path to kubeconfig file |
 | `--kubeconfig-dir` | | Comma-separated directories containing kubeconfig files |
 | `--namespace` | (all) | Initial namespace filter (supports multi-select in the UI; also used as RBAC fallback for namespace-scoped users) |
+| `--namespace-scope` | `false` | Pin namespaced informer caches to one namespace for large clusters. Requires `--namespace`, a kubeconfig context namespace, or a saved local single-namespace pick. Local mode can rebuild the cache when switching namespaces; auth/cloud mode locks the shared cache to the startup namespace. |
 | `--port` | `9280` | Server port |
 | `--no-browser` | `false` | Don't auto-open browser |
 | `--browser` | | Browser to use when opening the UI, e.g. `firefox`, `google-chrome`, or `Google Chrome` on macOS |
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ radar
 | `--kubeconfig` | `~/.kube/config` | Path to kubeconfig file |
 | `--kubeconfig-dir` | | Comma-separated directories containing kubeconfig files |
 | `--namespace` | (all) | Initial namespace filter (supports multi-select in the UI; also used as RBAC fallback for namespace-scoped users) |
-| `--namespace-scope` | `false` | Pin namespaced informer caches to one namespace for large clusters. Requires `--namespace`, a kubeconfig context namespace, or a saved local single-namespace pick. Local mode can rebuild the cache when switching namespaces; auth/cloud mode locks the shared cache to the startup namespace. |
+| `--namespace-scope` | `false` | Pin namespaced informer caches to a **single** namespace for large clusters (scoping to multiple namespaces is not supported yet). Requires `--namespace`, a kubeconfig context namespace, or a saved local single-namespace pick. Local mode can rebuild the cache when switching namespaces; auth/cloud mode locks the shared cache to the startup namespace. |
 | `--port` | `9280` | Server port |
 | `--no-browser` | `false` | Don't auto-open browser |
 | `--browser` | | Browser to use when opening the UI, e.g. `firefox`, `google-chrome`, or `Google Chrome` on macOS |
--- a/cmd/explorer/main.go
+++ b/cmd/explorer/main.go
@@ -63,7 +63,7 @@ func main() {
 	podShellDefault := flag.String("pod-shell-default", "", "Override the default pod exec shell command (runs as 'sh -c <value>'; empty = built-in bash -il → ash → sh cascade)")
 	debugImage := flag.String("debug-image", fileCfg.DebugImage, "Image for ephemeral debug containers and node debug pods (empty = busybox:latest; point at a mirror for air-gapped/private-registry clusters)")
 	listPageSize := flag.Int64("list-page-size", 0, "Paginate the initial LIST of high-cardinality kinds (Pods, ReplicaSets) at this page size on clusters without WatchList streaming. 0 = off (single LIST). Try 2000 if a very large cluster fails to sync.")
-	namespaceScope := flag.Bool("namespace-scope", false, "Scope namespaced informer caches to one namespace. Requires --namespace or a kubeconfig context namespace. Local mode can rescope by switching namespaces; auth/cloud mode locks to the startup namespace.")
+	namespaceScope := flag.Bool("namespace-scope", false, "Scope namespaced informer caches to a single namespace (multiple namespaces are not supported yet). Requires --namespace or a kubeconfig context namespace. Local mode can rescope by switching namespaces; auth/cloud mode locks to the startup namespace.")
 	// Timeline storage options
 	timelineStorage := flag.String("timeline-storage", fileCfg.TimelineStorageOr("memory"), "Timeline storage backend: memory or sqlite")
 	timelineDBPath := flag.String("timeline-db", fileCfg.TimelineDBPath, "Path to timeline database file (default: ~/.radar/timeline.db)")
--- a/cmd/explorer/main.go
+++ b/cmd/explorer/main.go
@@ -63,6 +63,7 @@ func main() {
 	podShellDefault := flag.String("pod-shell-default", "", "Override the default pod exec shell command (runs as 'sh -c <value>'; empty = built-in bash -il → ash → sh cascade)")
 	debugImage := flag.String("debug-image", fileCfg.DebugImage, "Image for ephemeral debug containers and node debug pods (empty = busybox:latest; point at a mirror for air-gapped/private-registry clusters)")
 	listPageSize := flag.Int64("list-page-size", 0, "Paginate the initial LIST of high-cardinality kinds (Pods, ReplicaSets) at this page size on clusters without WatchList streaming. 0 = off (single LIST). Try 2000 if a very large cluster fails to sync.")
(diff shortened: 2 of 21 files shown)


nadaverell:
Rebased onto current main (was 104 commits behind) and ran a careful review loop. Force-pushed: 2 commits — the rebased feature + review fixes.

**Rebase:** 4 small conflicts resolved (struct fields, an import, a frontend error string, and a keep-both test merge); `context_manager.go` merged **additively** (main's evolved 299-line version + the rescope functions, no main code reverted). The scoping core cleanly **activates main's previously-dormant `forceNamespace` pin-path** in `probeResourceAccess` — exactly the hook it was reserved for.

**Review-loop fixes (self + Codex):**
- **F1 (High):** namespaced **CRD/dynamic** informers weren't pinned under `--namespace-scope` (only typed were), so a CRD-heavy large cluster could still OOM cluster-wide. Now namespaced CRDs pin to the target via the `gvrIsNamespaced` guard in `probeScope`; cluster-scoped CRDs stay cluster-wide.
- **F2 (High):** `PerformNamespaceRescope` rollback could clobber a newer context switch/rescope. Added a generation token (bumped in `CancelOngoingOperations`); the rescope skips rollback + stale notify when superseded.
- **F3 (Med):** auth-mode scope endpoint advertised the pinned namespace as accessible to users without RBAC to it. Now only reports it when the user can actually list it.

Confirmed sound: multi-user POST safety (auth users get 403 on rescope), typed cluster-only kinds stay cluster-wide. Green: `go build`, `internal/{k8s,server,app}`, `pkg/k8score`, `tsc`.

Fast-follow (not blocking): a unit test pinning the F1 behavior (forced-mode keeps cluster-scoped CRDs cluster-wide).

PR #868, opened 2026-06-05
Add endpointslices.discovery.k8s.io to rbac permissions

### LINKED ISSUE #848: Support editing Endpoint Slice
Hello,

I just want to say you've written an amazing application and my team has swapped to using it full time as it honestly blows other solutions out of the water.

## Problem statement
Currently Radar doesn't support EndpointSlice, so if someone has created a service with manual resolution via EndpointSlice they are unable to edit/view this information.

https://kubernetes.io/docs/reference/kubernetes-api/discovery/endpoint-slice-v1/

One use case we have here is we have an external dialing system, where we create a service that allows our internal Kubernetes services to call these external diallers without each service needing to know their IPs as they can just reference this internal service name.

```
apiVersion: v1
kind: Service
metadata:
  name: dial-lb
  namespace: external
spec:
  type: ClusterIP
  ports:
    - name: http
      port: 80
      targetPort: 80
---
apiVersion: discovery.k8s.io/v1
kind: EndpointSlice
metadata:
  name: dial-lb-slice
  namespace: external
  labels:
    kubernetes.io/service-name: dial-lb
addressType: IPv4
ports:
  - name: http
    appProtocol: http
    port: 80
endpoints:
  - addresses:
      - 10.9.0.43
      - 10.9.0.44
      - 10.9.0.45
      - 10.9.0.46
      - 10.9.0.47
      - 10.9.0.48
    conditions:
      ready: true
      serving: true
```

Its a fairly niche use case but it is handy.

Moulick:
## Description

Add missing permissions to get/list/watch `endpointslices.discovery.k8s.io`. Seems to have been missed in https://github.com/skyhook-io/radar/pull/850

## Type of change

## How has this been tested?

Adding to `additionalCrdGroups` locally

Describe the tests you ran to verify your changes.

Error fixed: `radar 2026/06/05 14:53:52 [resource-counts] Failed to count EndpointSlice: failed to list resources: endpointslices.discovery.k8s.io is forbidden: User "system:serviceaccount:radar:radar" cannot list resource "endpointslices" in API group "discovery.k8s.io" at the cluster scope`

## Related issues

Fixes #848

### DIFF
--- a/deploy/helm/radar/templates/clusterrole.yaml
+++ b/deploy/helm/radar/templates/clusterrole.yaml
@@ -25,6 +25,11 @@ rules:
       - resourcequotas
     verbs: ["get", "list", "watch"]
   - apiGroups:
+      - discovery.k8s.io
+    resources:
+      - endpointslices
+    verbs: ["get", "list", "watch"]
+  - apiGroups:
       - policy
     resources:
       - poddisruptionbudgets


Moulick:
Question, the #848 talks about editing Endpoint slices, this PR only adds get/list/watch, should it also have patch/update/delete @nadaverell?

nadaverell:
Good question — keep it `get/list/watch` only. The chart's ClusterRole is deliberately read-only across the board (no write verbs on pods, deployments, or anything else — the only exception is the `rbac.helm` wildcard, since charts can create arbitrary kinds).

Editing works through a different path: with auth enabled, Radar impersonates the requesting user, so the apiserver enforces *their* RBAC, not the ServiceAccount's. For no-auth installs where you want the SA itself to write, `rbac.additionalRules` is the escape hatch.

So this PR fully fixes #848 — once the SA can read EndpointSlices, editing them works wherever editing anything else works. Thanks for the fix!

nadaverell (approved):
LGTM — thanks for catching this!

PR #949, opened 2026-06-19
Cluster discovery oracle: detect observability-vendor signatures

hisco:
## What

Adds a cluster-side **discovery oracle**. radar scans the cluster for
observability-vendor *signatures* and assembles a per-cluster report it pushes
to the configured cloud upstream. radar detects only — it does not own, install,
or call any integration.

- `internal/cloud/discovery/` — probe framework (`discovery.go`, `runner.go`,
  `readers.go`) and a `pusher` that sends the assembled `Report` upstream.
- Signature probes that detect a vendor's presence from its own CRs/Secrets:
  Coralogix, Datadog, Dynatrace, Honeycomb, New Relic, Sentry. Each returns a
  `Hit` — the catalog type ids it flags, non-secret prefill values, and
  human-readable evidence.
- `cmd/explorer/` — wires the sweep into the agent boot path.

## Trust boundary (by design)

- Reads only Secrets/CRs the customer's existing RBAC already grants radar.
- **Never reads secret material** — e.g. the Datadog probe reports *where* the
  API key lives (`secret/X[key] (not read)`) and prefills only non-secret config
  such as `site`. Credentials never leave the cluster.
- A failing probe is logged and skipped — one broken signature can't blind the
  rest of the sweep, and "vendor absent" is the normal case, not an error.

## Why

Surface what radar can already read from a cluster's own Secrets/CRs as a
structured report, so vendor presence and non-secret config don't have to be
entered by hand elsewhere.

## Notes

- Purely additive; no existing radar package imports this.

## Testing

- `go build ./...` green.

---

> [!NOTE]
> **Medium Risk**
> Cluster-wide pod/CR lists and outbound pushes to a bearer-authenticated URL increase data exposure and API load; probes are designed not to read secret material, but misconfiguration or future probe changes could widen what leaves the cluster.
> 
> **Overview**
> Adds an **optional cloud-discovery loop** that periodically scans the cluster for observability vendor signatures and **POSTs a per-cluster `Report`** to a configured upstream ingest URL. When `--cloud-discovery-url` (or `RADAR_CLOUD_DISCOVERY_URL`) is empty, nothing runs—OSS behavior is unchanged.
> 
> New **`internal/cloud/discovery`** package: probe framework (`Run` with per-probe 15s timeouts and fail-soft logging), narrow `KubeReader`/`DynReader` interfaces, HTTP **`Pusher`** (Bearer token), and **`Start`** runner (3s warmup, then ticker, re-resolves kube clients and cluster id each sweep for context switches). **Six probes** flag catalog types and optional non-secret **prefill** plus **evidence** text: Datadog (`DatadogAgent` CR, site only), Dynatrace (`DynaKube`, `environment_id`), Coralogix (`CORALOGIX_DOMAIN` → region), and presence-only Sentry, New Relic, Honeycomb from env vars/images.
> 
> **`cmd/explorer`** adds `--cloud-discovery-url`, `--cloud-discovery-token`, and `--cloud-discovery-interval` (default 60s), wiring live k8s/dynamic clients and cluster id (falls back to kubeconfig cluster name when `--cluster-name` is unset).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c76dab4c35a87809da6934ffda0ba5f1a5936a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/explorer/main.go
+++ b/cmd/explorer/main.go
@@ -19,6 +19,7 @@ import (
 	"github.com/skyhook-io/radar/internal/auth"
 	"github.com/skyhook-io/radar/internal/cloud"
 	"github.com/skyhook-io/radar/internal/config"
+	"github.com/skyhook-io/radar/internal/cloud/discovery"
 	"github.com/skyhook-io/radar/internal/k8s"
 	mcppkg "github.com/skyhook-io/radar/internal/mcp"
 	"github.com/skyhook-io/radar/internal/server"
@@ -108,6 +109,16 @@ func main() {
 	cloudURL := flag.String("cloud-url", os.Getenv("RADAR_CLOUD_URL"), "Radar Cloud WebSocket URL (e.g. wss://api.radarhq.io/agent) — empty = local-only. Env: RADAR_CLOUD_URL")
 	cloudToken := flag.String("cloud-token", os.Getenv("RADAR_CLOUD_TOKEN"), "Cluster token from the Radar Cloud install wizard (rhc_<random>). Env: RADAR_CLOUD_TOKEN")
 	cloudClusterName := flag.String("cluster-name", os.Getenv("RADAR_CLOUD_CLUSTER_NAME"), "Human-readable cluster name for Radar Cloud (required with --cloud-url). Env: RADAR_CLOUD_CLUSTER_NAME")
+	// Cloud-discovery flags. When --cloud-discovery-url is set radar scans
+	// the cluster for observability vendor signatures (Datadog operator
+	// CR, Sentry env vars, etc) and pushes the report to the configured
+	// cloud upstream. Empty disables the loop entirely — OSS users see
+	// no behavior. The naming + env shape mirror --cloud-url / --cloud-
+	// token so all "talks to a cloud upstream" knobs share the prefix;
+	// nothing about Skyhook's hub specifically lives in the OSS binary.
+	cloudDiscoveryURL := flag.String("cloud-discovery-url", os.Getenv("RADAR_CLOUD_DISCOVERY_URL"), "Cloud discovery ingest URL (e.g. https://api.example.com/api/internal/discovery). Empty = disabled. Env: RADAR_CLOUD_DISCOVERY_URL")
+	cloudDiscoveryToken := flag.String("cloud-discovery-token", os.Getenv("RADAR_CLOUD_DISCOVERY_TOKEN"), "Bearer token authenticating discovery push to the cloud upstream. Env: RADAR_CLOUD_DISCOVERY_TOKEN")
+	cloudDiscoveryInterval := flag.Duration("cloud-discovery-interval", 60*time.Second, "How often radar re-runs the cluster discovery sweep and re-pushes to the cloud upstream.")
 	// Tunable deadlines for slow / high-latency / SSH-tunneled clusters.
 	// Defaults preserve the original behavior. Each flag falls back to an
 	// environment variable so Kubernetes deployments can source values from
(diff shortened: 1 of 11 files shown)


internal/cloud/discovery/pusher.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// Pusher ships a Report to the cloud upstream's discovery ingest endpoint.
+//
+// Auth: a static bearer token shared via env (HUB_INTEGRATIONS_DISCOVERY_TOKEN
+// on both sides). This is the demo-only auth. The production path will
```
nadaverell, line 16: why do we need this? i don't really understand from the comment. we can test locally e2e without a token, no?

also I don't see where HUB_INTEGRATIONS_DISCOVERY_TOKEN is used

internal/cloud/discovery/runner.go
```diff
@@ -0,0 +1,78 @@
...
+type RunnerConfig struct {
+	UpstreamURL string
+	Token       string
+	Interval    time.Duration // 0 → 60s
+}
+
+// Start kicks off the discovery loop in a goroutine. Returns
+// immediately. Cancel via the supplied ctx.
+//
+// First push runs ~3s after start to give the kube client time to warm
```
nadaverell, line 21: this is not enough time, only small clusters are done in 3 seconds. can take 30+ seconds in larger clusters

internal/cloud/discovery/runner.go
```diff
@@ -0,0 +1,78 @@
...
+// cluster identity change, and a stale capture would keep listing the old
+// API server and pushing under the old cluster id.
+func Start(ctx context.Context, cfg RunnerConfig, kubeFn func() KubeReader, dynFn func() DynReader, clusterFn func() (id, name string)) {
+	if cfg.UpstreamURL == "" {
+		slog.Info("discovery: UpstreamURL empty, loop disabled")
+		return
+	}
+	interval := cfg.Interval
+	if interval <= 0 {
+		interval = 60 * time.Second
```
nadaverell, line 36: so we're running this discovery in a loop forever every minute?

nadaverell: I'm missing the other half - what are we doing with this info? Do we need this pushed constantly to hub, or should it be on-demand instead?

cmd/explorer/main.go
```diff
@@ -107,6 +108,16 @@ func main() {
...
 	cloudToken := flag.String("cloud-token", os.Getenv("RADAR_CLOUD_TOKEN"), "Cluster token from the Radar Cloud install wizard (rhc_<random>). Env: RADAR_CLOUD_TOKEN")
 	cloudClusterName := flag.String("cluster-name", os.Getenv("RADAR_CLOUD_CLUSTER_NAME"), "Human-readable cluster name for Radar Cloud (required with --cloud-url). Env: RADAR_CLOUD_CLUSTER_NAME")
+	// Cloud-discovery flags. When --cloud-discovery-url is set radar scans
+	// the cluster for observability vendor signatures (Datadog operator
+	// CR, Sentry env vars, etc) and pushes the report to the configured
+	// cloud upstream. Empty disables the loop entirely — OSS users see
+	// no behavior. The naming + env shape mirror --cloud-url / --cloud-
+	// token so all "talks to a cloud upstream" knobs share the prefix;
+	// nothing about Skyhook's hub specifically lives in the OSS binary.
+	cloudDiscoveryURL := flag.String("cloud-discovery-url", os.Getenv("RADAR_CLOUD_DISCOVERY_URL"), "Cloud discovery ingest URL (e.g. https://api.example.com/api/internal/discovery). Empty = disabled. Env: RADAR_CLOUD_DISCOVERY_URL")
```
nadaverell, line 119: why do we need this in addition to the main cloud url? for the specific endpoint (path)? isn't it a bit redundant, if someone is pointing radar to cloud already?

hisco:
Closing. Shipping the integrations feature as hub-only manual catalog install to keep the surface area minimal. Auto-detect/discovery (this PR) is being dropped for now to reduce complexity — the working feature (hub #68 + web #116) doesn't depend on it. Cluster-side auto-detect can return as a clean follow-up if prioritized.

PR #954, opened 2026-06-20
Timeline: hide-deletes toggle + GitOps empty-state polish

nadaverell:
## Summary

Two small, independent timeline/GitOps improvements that came out of the Radar Cloud / Hub fleet work.

1. **Timeline "Deleted" toggle** — operators can now hide delete-event noise in both the list and swimlane timeline views. Delete events are the *only* surviving record of a resource that no longer exists (a removed Deployment, a Pod that's already gone), so they remain visible by default; the toggle exists to de-noise a churny window on demand.
2. **GitOps empty-state polish** — small presentation change to `GitOpsTableView` so a host (Radar Hub's fleet view) can render a richer empty state.

## What changed

**Timeline deleted-events filtering**
- New `ExcludeDeleted` query option on the timeline store (`include_deleted=false` on `/api/changes`), default off (deletes shown).
- SQLite store applies the delete filter **in SQL, before `ORDER BY … LIMIT`**, so hiding deletes can't under-fill a page when the newest rows happen to be deletes — this keeps the SQLite and in-memory backends consistent (regression test added for the limit-skew case).
- Removed the unused `IncludeDeletedManaged` option: the list view queries with the `all` preset (`IncludeManaged: true`), so managed-resource deletes already surface there; the option was dead on the real UI path. Managed deletes continue to follow `IncludeManaged` as before.
- UI: a `Deleted` control in the list toolbar (with a live count) and the swimlane toolbar; threaded through `useChanges` (`includeDeleted`).

**GitOps empty-state**
- `emptyStateTitle` / `emptyStateBody` widened from `string` → `ReactNode` (a host can split the body across lines or emphasize part of it; plain strings still work — non-breaking).
- Empty-state card anchored higher in the viewport instead of dead-center, so it reads better on a tall page.

## Testing
- `make tsc` ✓ · `go test ./...` for `pkg/timeline` + `internal/timeline` + `internal/server` ✓ (incl. new `DeletedFiltering` and `ExcludeDeleted_LimitSkew` tests).
- Visual-tested against a local `kind` cluster on the real `filter=all` path: toggle ON → delete rows shown (incl. Pods/ReplicaSets); toggle OFF → zero deletes, other events intact.

## Notes
- The list timeline intentionally shows all managed activity (pod creates/updates/deletes) by default; the toggle is purely a way to hide deletes, so it defaults ON for consistency.

---

> [!NOTE]
> **Low Risk**
> Read-path timeline filtering and presentation tweaks with regression tests; no auth, data writes, or security-sensitive behavior.
> 
> **Overview**
> Adds a **Deleted** control on timeline list and swimlane views (shared state when switching views) so operators can hide delete-event noise; deletes stay **on** by default.
> 
> Hiding deletes is wired through **`/api/changes`** via `include_deleted=false`, a new **`ExcludeDeleted`** query option on timeline stores, and **`useChanges`** (`includeDeleted`). The SQLite backend applies the delete filter **before** `ORDER BY … LIMIT` so a page of recent deletes cannot leave the UI empty; memory store matches. Managed-resource deletes still follow **`IncludeManaged`** as before.
> 
> **GitOpsTableView** empty-state copy props widen from `string` to **`ReactNode`** for richer host messaging, and the empty card sits higher in the viewport instead of vertically centered.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13762c415cf7a108189dd795b3d0ca4a720e21f2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -2375,6 +2375,7 @@ func (s *Server) handleChanges(w http.ResponseWriter, r *http.Request) {
 	filterPreset := r.URL.Query().Get("filter")
 	includeK8sEvents := r.URL.Query().Get("include_k8s_events") != "false" // default true
 	includeManaged := r.URL.Query().Get("include_managed") == "true"       // default false
+	includeDeleted := r.URL.Query().Get("include_deleted") != "false"      // default true
 	sourcesParam := r.URL.Query().Get("sources")                           // comma-separated, e.g. "k8s_event"
 
 	// Parse since timestamp
@@ -2410,6 +2411,7 @@ func (s *Server) handleChanges(w http.ResponseWriter, r *http.Request) {
 		Since:            since,
 		Limit:            limit,
 		IncludeManaged:   includeManaged,
+		ExcludeDeleted:   !includeDeleted,
 		IncludeK8sEvents: includeK8sEvents,
 		FilterPreset:     filterPreset,
 		// The persistent store retains events from previously-connected
--- a/internal/timeline/sqlite_store.go
+++ b/internal/timeline/sqlite_store.go
@@ -350,6 +350,13 @@ func (s *SQLiteStore) Query(ctx context.Context, opts QueryOptions) ([]TimelineE
 		query.WriteString(")")
 	}
 
+	// Filter deletes in SQL (before ORDER/LIMIT) so hiding them can't under-fill
+	// the page when the newest rows happen to be deletes.
+	if opts.ExcludeDeleted {
+		query.WriteString(" AND event_type != ?")
+		args = append(args, string(EventTypeDelete))
+	}
+
 	if opts.ClusterContext != "" {
 		query.WriteString(" AND cluster_context = ?")
 		args = append(args, opts.ClusterContext)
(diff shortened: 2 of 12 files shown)


nadaverell:
Both Bugbot findings addressed in 13762c415:

- **Swimlane hides deletes client-side only** — `showDeleted` is now lifted to `TimelineView` and fed into the swimlane's `useChanges({ includeDeleted })`, so hiding deletes excludes them in SQL **before** the 10k `LIMIT` (matching the list path) instead of only filtering the already-capped page in the browser.
- **List/swimlane toggles unsynchronized** — the two views now share that single lifted state, so the toggle carries across the view switch.

The shared `k8s-ui` `TimelineList`/`TimelineSwimlanes` gained optional `showDeleted`/`onShowDeletedChange` props (controlled when supplied, internal state otherwise), so external consumers are unaffected. Verified live: list toggle off → swimlane checkbox inherits off + swimlane fetch sends `include_deleted=false` (0 deletes returned).

PR #972, opened 2026-06-23
Settings dialog: polish + apply Prometheus URL without a restart

nadaverell:
## Summary

Cleans up the Settings dialog and removes a sharp edge: it's effectively a startup-config editor — almost every field is read once at launch and inert until restart — but the layout didn't say so, its most valuable control sat last, and the one field that *can* take effect immediately (the Prometheus URL) forced a needless restart loop. This makes the next-launch reality visible at the moment you save, surfaces the AI Tools config, and lets the Prometheus connection — URL **and** auth headers — apply live with inline reachability feedback.

## What changed

**Layout & comprehension**
- **Config grouped into bordered card sections** — Cluster / Server / AI Tools / Timeline / Integrations — for clear visual separation (replacing thin dividers), in a slightly wider dialog so fields and the header editor have room. Dialog tooltips route through the shared `Tooltip` for consistent styling.
- **Version moved into the header** (`Radar v… · by Skyhook`), baseline-aligned beside the title. The old bottom "About" bar (and its redundant GitHub link) is gone; the version is now visible to non-owners too.
- **Persistent "Applies on next launch" cue beside Save** — the dialog's dominant fact, previously small text that scrolled away, now sits at the decision point.
- **MCP promoted to its own "AI Tools" section near the top**, where the port it depends on is already configured (was buried last under Integrations).
- Input/select focus borders moved to the brand `skyhook-500` token; footer padding trimmed.

**Apply the Prometheus connection live (no restart)**
- Unlike the rest of config, the Prometheus URL + headers live in mutable globals the metrics path reads per query — so they can be re-pointed at runtime, which is exactly when operators touch them (empty metrics/cost maps otherwise mean restart-to-test).
- New **`PUT /api/integrations/prometheus`**: validates the URL (rejects anything startup would `log.Fatalf` on, so Apply can't brick the next launch), persists URL + headers, re-points the running client (`SetManualURL`/`SetHeaders` + `Reset()` to drop the cached connection), then probes and returns the live reachability result. Empty URL reverts to auto-discovery.
- **Auth headers** for token / multi-tenant backends (Bearer, `X-Scope-OrgID` — Mimir/Cortex/Thanos): an inline key/value editor with password-masked values. Values are write-only — `GET /config` returns only the configured header *names* (`prometheusHeaderKeys`), never the values — so the editor replaces all headers on apply rather than pretending to round-trip secrets.
- Field shows live status (`Connected to X` / `Saved, but not reachable` / `Couldn't apply`) and opts out of the next-launch framing since it genuinely applies immediately.

**Namespace clarifier**
- "Default Namespace" help now notes it's the startup default only and the active namespace is live-switchable from the header — so users don't restart needlessly. (No behavior change.)

### Reviewer notes
- Live-apply re-points the **shared** Prometheus client, so **Istio** mesh traffic (which uses it) and resource-utilization/cost go live immediately. **Cilium/Hubble** traffic uses Hubble Relay (gRPC), not Prometheus — unaffected. The legacy Caretta source caches its own endpoint and updates on restart; it's slated for deprecation.
- The other restart-only fields (kubeconfig, port, timeline, MCP toggle) stay next-launch by design — captured into closures/singletons/the immutable router at boot, and for a local binary a restart is cheap.
- Concurrency: the runtime writers are lock-correct — `SetManualURL` writes under `c.mu` (matching `SetHeaders`/discovery reads), and the `traffic` metrics-config globals are mutex-guarded against the context-switch rebuild. Verified with `go test -race`.
- SSRF: the endpoint is owner-gated; in no-auth mode the whole instance is already an unauthenticated console, so an owner-issued probe isn't the weak link (consistent with the existing config-set path). Accepted deliberately.

## Testing

- `make tsc`, `go build ./...`, `go test -race ./internal/server/ ./internal/prometheus/ ./internal/traffic/` — all clean.
- Visual-tested against a live cluster (light + dark): card sections, wider dialog, header version, footer cue, AI Tools section, namespace help, custom tooltips.
- Exercised the live apply end-to-end: applying URL + `X-Scope-OrgID` header persisted both and returned the probe result; `GET /config` exposed the header *key* with the value redacted; clearing reverted config to `{}`; an invalid URL returned 400 without persisting.

---

> [!NOTE]
> **Medium Risk**
> Owner-gated endpoint mutates shared Prometheus/traffic globals and persists secrets; concurrency is mutex-guarded but the live HTTP probe to operator-supplied URLs is an intentional SSRF surface consistent with existing config paths.
> 
> **Overview**
> Adds **live Prometheus reconfiguration** so owners can persist and apply URL and auth headers without restarting, plus a clearer Settings layout.
> 
> **Backend:** New owner-gated `PUT /api/integrations/prometheus` validates HTTP(S) URLs, writes config first, updates the running Prometheus and traffic metrics globals, resets the cached client, and returns a probe result. `GET /api/config` exposes `prometheusHeaderKeys` only (values stay redacted). Context-switch Prometheus reinit no longer reapplies startup URL/headers, so live changes survive switches. `SetManualURL` / `SetHeaders` and traffic metrics config use mutexes for safe concurrent updates.
> 
> **UI:** Settings uses bordered sections (Cluster, Server, AI Tools, Timeline, Integrations), a wider dialog, version in the header, and an “Applies on next launch” hint beside Save. Prometheus gets **Apply now** with inline connected/unreachable feedback and an optional auth-header editor that replaces stored headers on apply without round-tripping secrets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10e0e35f5b3183162b56f4f1f53b8db4681a4f0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/app/bootstrap.go
+++ b/internal/app/bootstrap.go
@@ -148,14 +148,11 @@ func RegisterCallbacks(cfg AppConfig, timelineStoreCfg timeline.StoreConfig) {
 		return traffic.ReinitializeWithConfig(k8s.GetClient(), k8s.GetConfig(), k8s.GetContextName())
 	})
 
+	// Reinitialize carries the current manual URL + headers forward (including any
+	// applied live via /integrations/prometheus). Re-applying the captured startup
+	// cfg here would revert a live change on context switch, so we don't.
 	k8s.RegisterPrometheusFuncs(prometheuspkg.Reset, func() error {
 		prometheuspkg.Reinitialize(k8s.GetClient(), k8s.GetConfig(), k8s.GetContextName())
-		if cfg.PrometheusURL != "" {
-			prometheuspkg.SetManualURL(cfg.PrometheusURL)
-		}
-		if len(cfg.PrometheusHeaders) > 0 {
-			prometheuspkg.SetHeaders(cfg.PrometheusHeaders)
-		}
 		return nil
 	})
 }
--- a/internal/prometheus/client.go
+++ b/internal/prometheus/client.go
@@ -77,12 +77,19 @@ func newMCPHTTPClient() *http.Client {
 }
 
 // SetManualURL sets the --prometheus-url override on the global client.
+// manualURL is read under the per-client c.mu (discover, Reinitialize), so the
+// write must take c.mu too — not clientMu — to avoid a lock-mismatch race once
+// this is callable at runtime. Mirrors SetHeaders.
 func SetManualURL(rawURL string) {
-	clientMu.Lock()
-	defer clientMu.Unlock()
-	if globalClient != nil {
-		globalClient.manualURL = strings.TrimRight(rawURL, "/")
+	clientMu.RLock()
+	c := globalClient
+	clientMu.RUnlock()
+	if c == nil {
+		return
 	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.manualURL = strings.TrimRight(rawURL, "/")
 }
 
 // SetHeaders sets HTTP headers attached to every Prometheus request on the
--- a/internal/prometheus/client.go
+++ b/internal/prometheus/client.go
@@ -78,35 +78,35 @@ func newMCPHTTPClient() *http.Client {
 
 // SetManualURL sets the --prometheus-url override on the global client.
 // manualURL is read under the per-client c.mu (discover, Reinitialize), so the
-// write must take c.mu too — not clientMu — to avoid a lock-mismatch race once
-// this is callable at runtime. Mirrors SetHeaders.
+// write takes c.mu too. clientMu is held (read) across the whole write so a
+// concurrent Reinitialize can't swap globalClient out from under us and leave the
+// new pointer with stale settings. Lock order clientMu→c.mu matches Reinitialize.
 func SetManualURL(rawURL string) {
 	clientMu.RLock()
-	c := globalClient
-	clientMu.RUnlock()
-	if c == nil {
+	defer clientMu.RUnlock()
+	if globalClient == nil {
 		return
 	}
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	c.manualURL = strings.TrimRight(rawURL, "/")
+	globalClient.mu.Lock()
+	defer globalClient.mu.Unlock()
+	globalClient.manualURL = strings.TrimRight(rawURL, "/")
 }
 
 // SetHeaders sets HTTP headers attached to every Prometheus request on the
-// global client. Pass nil or an empty map to clear.
+// global client. Pass nil or an empty map to clear. Holds clientMu (read) across
+// the write for the same reason as SetManualURL.
 func SetHeaders(h map[string]string) {
 	clientMu.RLock()
-	c := globalClient
-	clientMu.RUnlock()
-	if c == nil {
+	defer clientMu.RUnlock()
+	if globalClient == nil {
 		return
 	}
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	c.headers = copyHeaders(h)
+	globalClient.mu.Lock()
+	defer globalClient.mu.Unlock()
+	globalClient.headers = copyHeaders(h)
 	// Drop the cached prom.Client so the next request rebuilds its transport
 	// with the new headers.
-	c.prom = nil
+	globalClient.prom = nil
 }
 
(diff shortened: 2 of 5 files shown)


nadaverell:
Triaged the Cursor Bugbot + Codex review findings (commit cc69af5a):

**Fixed**
- **Runtime updated before persist** — reordered to persist-first; a failed disk write no longer leaves the running client repointed while config disagrees.
- **False "Saved" message on failure** — now distinguishes a failed request (`Couldn't apply` — nothing persisted) from a saved URL whose probe failed (`Saved, but not reachable`).
- **(Codex) Boot-brick risk** — the handler now validates the URL is `http(s)` before persisting, matching startup's `log.Fatalf` check, so "Apply now" can't save a config that fails the next launch.
- **(Codex) Lock-mismatch race** — `SetManualURL` now writes `manualURL` under `c.mu` (the lock discovery reads under), mirroring `SetHeaders`. Verified with `go test -race`.
- **Traffic metrics-config race** — the runtime write to `configuredMetricsURL` is now mutex-guarded against the context-switch rebuild read.

**Acknowledged — follow-up, not this PR**
- **Traffic metrics URL stale (Bugbot High / Codex High)** — accurate: "Apply now" re-points the Prometheus *metrics/cost* client live (what the probe confirms), but the already-running Caretta traffic source caches its endpoint and only picks up the new URL on restart (via persisted config) or context switch. Making the live Caretta source re-point is a separate change against its immutable-by-design assumption; tracking separately.
- **SSRF surface in no-auth mode** — the endpoint is owner-gated, but no-auth deployments bound to 0.0.0.0 expose an arbitrary-URL probe (the risk `handleConnect` deliberately avoids). Flagging for a product decision before expanding the surface.

**Won't fix**
- **Port-forward not stopped on re-point** — low-impact; the PF lifecycle is pre-existing and bounded. Tracking only if it proves costly.

nadaverell:
Triaged the latest Cursor Bugbot findings (commit 7b4e09ec):

**Fixed**
- **Context switch reverts applied URL** (Medium) — real bug. The prometheus reinit closure re-applied the captured startup `cfg` after `Reinitialize`, which already carries the current (live-applied) URL + headers forward — so the re-apply could only clobber a live change. Removed it; live values now survive a context switch.
- **Header keys state desyncs** (Low) — the summary now derives from the prop (with an optimistic post-apply override) instead of a mount-time snapshot, so it no longer shows "No auth headers" while config is still loading.
- **Reset leaves apply status** (Low) — Reset / any external URL edit now clears a stale "Connected"/"Saved" message so it can't describe an emptied field.

**Already fixed (stale re-flags)**
- **Runtime updated before persist** — the handler is already persist-first: `config.Update` runs before any `SetManualURL`/`SetHeaders`/`Reset`.
- **False saved message on failure** — already split into `Couldn't apply` (request failed, nothing saved) vs `Saved, but not reachable` (persisted, probe failed).

Verified: `go build ./...`, `go test -race ./internal/{server,prometheus,traffic,app}`, `tsc` — all clean.

nadaverell:
Latest Bugbot round (commits 3804a26e, 10e0e35f):

**Fixed**
- **Open editor clears auth headers** (High) — opening the header editor for a URL-only change no longer wipes stored secrets. Blank/in-progress rows leave headers untouched; only entered rows replace them; removing every row clears.
- **Apply blocked without cluster** (Medium) — dropped `requireConnected` from the apply endpoint. Persisting + probing a manual URL is plain HTTP, so an external Prometheus can be applied while the cluster is unreachable (matches `handlePutConfig`).
- **Stale client pointer race** (Medium) — `SetManualURL`/`SetHeaders` now hold `clientMu` (read) across the per-client write, so a concurrent context-switch `Reinitialize` can't swap `globalClient` mid-write and leave it stale. Lock order `clientMu→c.mu` matches `Reinitialize`; `go test -race` clean.

**Already resolved (carry-over threads re-anchored to the latest commit)**
- Context switch reverts applied URL — fixed (7b4e09ec).
- Reset leaves apply status — fixed (7b4e09ec).
- Runtime updated before persist — already persist-first.
- False saved message — already split `Couldn't apply` vs `Saved, but not reachable`.

Verified: `go build ./...`, `go test -race ./internal/{prometheus,server}`, `tsc` — all clean.

PR #978, opened 2026-06-23
feat: use WebSocket API for exec & portforward, fall back to SPDY

0rax:
## Description

Migrates pod streaming (exec, attach, cp, port-forward) from the legacy SPDY transport to the new WebSocket API (see [Kubernetes 1.31: Streaming Transitions](https://kubernetes.io/blog/2024/08/20/websockets-transition/)), with automatic fallback to SPDY when the `apiserver` doesn't support the WebSocket subprotocol.

Kubernetes deprecated SPDY in favor of WebSocket ([KEP-4006](https://github.com/kubernetes/enhancements/issues/4006)) making WebSocket the default in kubectl since 1.31. As a result, some proxies/gateways no longer allow SPDY through (e.g. GKE Connect Gateway).

This switches Radar to the new API so streaming works against modern clusters and gateways, while the SPDY fallback keeps older apiservers working.

As a user of GKE Connect Gateway, this is a big deal: most features that require port-forward / exec when running Radar locally are currently unavailable. The goal here was to fix that by modernizing the connection logic.

## Type of change

- [ ] Documentation update

## How has this been tested?

- [ ] Added/updated unit tests

Tested against:
- a local Rancher Desktop cluster (v1.34.6+k3s1)
- two GKE clusters behind Connect Gateway (v1.35.5-gke)
- an older Rancher Desktop cluster without the new WebSocket APIs (v1.28.15+k3s1)

I've been running these changes locally for a week on the first 3 clusters (with network observability, Prometheus and OpenCost integrations active) without any issue. I also tested the fallback mechanism thoroughly on Kubernetes v1.28 to make sure no legacy clusters are impacted.

## Checklist

## Related issues

None

---

> [!NOTE]
> **Medium Risk**
> Touches core cluster connectivity (exec, cp, port-forward) across a client-go major bump; fallback logic limits blast radius on older apiservers but WebSocket-first behavior should be validated on target gateways and proxies.
> 
> **Overview**
> Pod **exec**, **file copy**, and **port-forward** paths now go through shared helpers that prefer the Kubernetes **WebSocket** streaming APIs and **fall back to SPDY** when the apiserver or gateway rejects the upgrade (e.g. GKE Connect Gateway).
> 
> New **`pkg/remotecommand.NewExecutor`** wraps WebSocket + SPDY fallback for exec; **`pkg/k8score`**, **`internal/server/copy`**, and related call sites switch from direct `NewSPDYExecutor` to that helper. **`pkg/portforward.NewDialer`** does the same for tunnels (WebSocket/SPDY-over-WebSocket first, legacy SPDY second) and is used by **`RunPortForward`** and the server’s port-forward sessions.
> 
> **`pkg/go.mod`** bumps **client-go** / **k8s.io** to **v0.36.2** and adds **`k8s.io/streaming`** for upgrade-failure detection. Comments and error wording drop “SPDY-only” framing where behavior is now transport-agnostic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fed9259e5e1d7e5c856b0393056d9c2adadca12e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/remotecommand/remotecommand.go
```diff
@@ -0,0 +1,26 @@
...
+func NewExecutor(config *rest.Config, u *url.URL) (remotecommand.Executor, error) {
+	wsExec, err := remotecommand.NewWebSocketExecutor(config, "GET", u.String())
+	if err != nil {
+		return nil, err
+	}
+	spdyExec, err := remotecommand.NewSPDYExecutor(config, "POST", u)
+	if err != nil {
+		return nil, err
+	}
+	return remotecommand.NewFallbackExecutor(wsExec, spdyExec, httpstream.IsUpgradeFailure)
```
nadaverell: The fallback only checks `IsUpgradeFailure`, but kubectl uses `IsUpgradeFailure(err) || IsHTTPSProxyError(err)` for both exec and port-forward. That second case is the HTTPS-proxy path — exactly what GKE Connect Gateway hits — so as-is a proxy error during the WS upgrade won't fall back to SPDY, and the gateway case this PR fixes could still break. Mind adding it to both helpers?

```go
shouldFallback := func(err error) bool {
    return httpstream.IsUpgradeFailure(err) || httpstream.IsHTTPSProxyError(err)
}
```

pkg/portforward/portforward.go
```diff
@@ -10,22 +10,42 @@ import (
...
+		return nil, fmt.Errorf("failed to create websocket dialer: %w", err)
+	}
+
+	transport, upgrader, err := spdy.RoundTripperFor(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create round tripper: %w", err)
+	}
+	spdyDialer := spdy.NewDialer(upgrader, &http.Client{Transport: transport}, "POST", u)
+
+	return portforward.NewFallbackDialer(wsDialer, spdyDialer, httpstream.IsUpgradeFailure), nil
```
nadaverell: Same here — needs `|| httpstream.IsHTTPSProxyError(err)` alongside `IsUpgradeFailure` so the SPDY fallback fires on HTTPS-proxy (Connect Gateway) failures too.

pkg/k8score/exec.go
```diff
@@ -28,15 +30,10 @@ func NewPodExecExecutor(client kubernetes.Interface, config *rest.Config, namesp
...
 			TTY:    tty,
 		}, scheme.ParameterCodec)
 
-	executor, err := remotecommand.NewSPDYExecutor(config, http.MethodPost, req.URL())
-	if err != nil {
-		return nil, fmt.Errorf("failed to create exec executor for %s/%s/%s: %w", namespace, podName, containerName, err)
-	}
-
-	return executor, nil
+	return rcpkg.NewExecutor(config, req.URL())
```
nadaverell: Minor: this used to wrap the error with namespace/pod/container — worth keeping for setup failures. Not blocking.

nadaverell (changes requested):
Thanks for the fix! I guess it was an oversight to ship SPDY, so thanks for correcting this. I tested it against our GKE clusters — exec, file browser/download, and port-forward all working on the WebSocket path, no regressions. One thing to fix before merge, then happy to approve.

pkg/remotecommand/remotecommand.go
0rax: Oooh that's an oversight on my part and did not surface in my testing. I should have checked the exact way `kubectl` catch those. This is now fix for both fallback dialers.

pkg/k8score/exec.go
0rax: Restored the original error check back to keep parity with the original code.

nadaverell (approved):
Looks good - thanks again!

PR #979, opened 2026-06-23
fix(sse): strip cluster-scoped topology kinds the user can't list

nadaverell:
## What
The REST `/api/topology` handler strips cluster-scoped graph nodes (Nodes, PVs, StorageClasses, NodePools, GatewayClasses) a user lacks `list` RBAC for. The SSE topology path did not, so a namespace-restricted user without cluster-scoped read still received those nodes when they appeared in the graph.

## Change
- Resolve the denied cluster-scoped kinds once per SSE connection (in `handleSSE`, where the request is available) and thread them to the broadcaster — the broadcast goroutine never runs a SAR.
- Apply `StripNodeKinds(deny)` to the initial topology frame and each broadcast build, matching the REST path.
- Broadcast grouping keys on the denied set, so a denied user gets a distinct stripped frame instead of a more-privileged peer’s bytes; full-access users coalesce (no-op).
- No-auth mode and namespace filtering unchanged.

## Tests
`TestDeniedKindsKey`; strip covered by `pkg/topology/strip_test.go`. `go test ./internal/server` green.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -3643,7 +3643,12 @@ func (s *Server) handleSSE(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 	r.URL.RawQuery = q.Encode()
-	s.broadcaster.HandleSSE(w, r)
+	// Cluster-scoped topology kinds (Nodes, PV, StorageClass, NodePool, …) have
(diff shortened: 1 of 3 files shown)


nadaverell:
Combined with the k8s_event fix into #984 (now retargeted to main, covers both). Same commits.

PR #1011, opened 2026-06-24
feat(helm): track upgrades for your own OCI-published charts

nadaverell:
## Problem

When a Helm release shows **"upstream unknown"**, Radar offers no way to upgrade it. This blocks the common modern case of upgrading your **own** charts: `helm install oci://ghcr.io/myorg/charts/mychart` from CI.

Two root causes:
1. **Helm doesn't persist install provenance** (still true in **v4** — verified) — a release knows only its chart name/version, not the repo/OCI ref it came from. Radar reverse-infers the source by scanning configured repos' cached `index.yaml`; "upstream unknown" = inference found nothing. Upstream tracking exists only as a *draft* proposal — [HIP-0018](https://helm.sh/community/hips/hip-0018/) (`repoURL` + tarball digest in release metadata), unshipped as of Helm 4 ([#4256](https://github.com/helm/helm/issues/4256), [#6464](https://github.com/helm/helm/issues/6464)). If/when it lands, Radar can read `repoURL` directly and surface it through the existing `sourceType` field — so this design is forward-compatible.
2. **OCI has no `helm repo add`** — classic repos live in `repositories.yaml`, but `helm registry login` stores only *auth*, so there's no "configured source list" for OCI to match against.

## What this does

Adds **registered OCI chart sources** — the OCI analog of `helm repo add`. You declare a registry/org prefix once (`oci://ghcr.io/myorg/charts`); Radar probes `<prefix>/<chartName>` for each untracked release, discovers newer versions via the Helm registry client, and lights up the **existing** update-available badge + upgrade flow. Matching is by chart name at query time, so **nothing per-release is persisted** and prefixes are global to the Radar instance (not cluster-scoped).

Also:
- Reframes the dead-end **"upstream unknown"** chip as a neutral, actionable **"source not tracked"** affordance that opens a *Track chart source* dialog.
- **GitOps-managed (Flux) releases** show "source via GitOps" / a non-clickable version and route to the GitOps view instead of offering a direct `helm upgrade` that would be reverted at the next reconcile.
- **Upgrades now use `--reset-then-reuse-values`** (was `--reuse-values`). Since the version picker lets users jump across chart versions, plain reuse-values is a footgun — it keeps the old merged values and renders nil for keys a newer chart added (verified: podinfo 6.3→6.14 fails on `.Values.hooks`). Reset-then-reuse resets to the new chart's defaults then re-merges the user's overrides on top — preserving customizations while picking up new keys. Applies to all Helm upgrades (classic + OCI).

## Security model — configured-only

- The server only ever probes **user-registered** prefixes. The upgrade path **re-derives** the `oci://` ref from a registered prefix — it never trusts a client-supplied ref.
- Credentials reuse the user's existing `helm registry login` store; **Radar stores no registry secrets**.
- Registry requests are bounded by an `http.Client` timeout (an unreachable registry can't stall the list view).
- Link-local / cloud-metadata literal IPs are rejected at registration.
- Pulled chart name is validated against the release's chart (no silent chart-swap).

OCI fallback runs **only on genuine classic-repo absence**, never on classic *ambiguity*, and surfaces classic index-load errors **before** falling back — so it can't mask or override a release that actually came from a classic repo.

## Test plan

- [x] `go build ./...`, `go vet`, full `go test ./...` green
- [x] New unit tests: prefix normalization (incl. link-local rejection, loopback/private allowed), discovery selection across prefixes, tag-cache dedupe, `applyOCIUpgrade` field-setting, newest-version selection
- [x] `make tsc` green
- [x] **On-cluster visual-test** (kind + real OCI registry, `podinfo` via `oci://ghcr.io/stefanprodan/charts`): full flow — source-not-tracked → register prefix → OCI discovery → version picker → **real applied upgrade/downgrade** pulled from the registry. Screenshots in the comment below.

## Open / deferred (see NOTES.md)

- ⚠️ **SSRF posture (needs a product call):** only link-local/metadata *literal* IPs are blocked; loopback/RFC1918 are allowed so local dev registries (`localhost:5000`) work, and a DNS name resolving into private ranges is **not** blocked. Full private-range + DNS-resolution blocking is hosted-mode hardening that would break local registries — deferred deliberately, flagged for review.
- Per-release source pin (for OCI charts that don't follow a consistent prefix); hosted-mode credential storage — both per the signed-off plan.

---

> [!NOTE]
> **Medium Risk**
> Adds server-initiated registry probes and OCI upgrade execution with partial SSRF guards; mistakes in classic-vs-OCI fallback logic could mislabel sources or pull from the wrong registry.
> 
> **Overview**
> Adds **registered OCI chart sources** (persisted in settings, exposed via list/add/remove APIs) so Radar can discover upgrades for charts installed from `oci://` when classic `repositories.yaml` has no match. Upgrade checks, batch list enrichment, and chart resolution now **fall through to OCI** only when classic repos are genuinely absent—not when indexes fail or classic sources are ambiguous—and surface **`untracked`**, **`sourceType`**, and **`chartRef`** on `UpgradeInfo` for the UI.
> 
> **Upgrade path:** OCI pulls use the user's `helm registry login` credentials with bounded HTTP timeouts; upgrades resolve `oci://` refs server-side from registered prefixes, validate chart name before apply, and share tag caching across batch checks.
> 
> **Version picker:** New `/versions` endpoint returns newest-first semver lists (capped at 50) from the matched classic repo or OCI prefix; the release drawer upgrade dialog lets users pick a target version and labels downgrades.
> 
> **UI:** Replaces the dead-end "upstream unknown" chip with **source not tracked** (opens `TrackChartSourceDialog`), distinguishes repo errors vs untracked sources, and routes Flux-managed releases away from direct `helm upgrade`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02d499b5e84453082a15b3033d8b856e21b234fe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/helm/client.go
+++ b/internal/helm/client.go
@@ -1278,6 +1278,106 @@ func (c *Client) applyOCIUpgrade(info *UpgradeInfo, chartName, currentVersion st
 	return true
 }
 
+// AvailableVersions is AvailableVersionsAsUser without impersonation.
+func (c *Client) AvailableVersions(namespace, name string) ([]string, error) {
+	return c.availableVersions(namespace, name, "", nil)
+}
+
+// AvailableVersionsAsUser returns the newest-first list of chart versions a
+// release could be upgraded (or downgraded) to, resolved from its source — the
+// matching classic repo's index or, failing that, a registered OCI source. Lets
+// the upgrade dialog offer a specific target version instead of only "latest".
+// Returns an empty list (not an error) when the source can't be determined; the
+// dialog then falls back to latest-only.
+func (c *Client) AvailableVersionsAsUser(namespace, name, username string, groups []string) ([]string, error) {
+	return c.availableVersions(namespace, name, username, groups)
+}
+
+func (c *Client) availableVersions(namespace, name, username string, groups []string) ([]string, error) {
+	var actionConfig *action.Configuration
+	var err error
+	if username != "" {
+		actionConfig, err = c.getActionConfigForUser(namespace, username, groups)
+	} else {
+		actionConfig, err = c.getActionConfig(namespace)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	rel, err := action.NewGet(actionConfig).Run(name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get release: %w", err)
+	}
+	chartName := rel.Chart.Metadata.Name
+
+	// Resolve the classic repo the same way the upgrade check does, then return
+	// that repo's full version list — never a union across repos, which could mix
+	// an unrelated same-named chart's versions.
+	var candidates []repoVersionInfo
+	versionsByRepo := map[string][]string{}
+	if f, err := repo.LoadFile(c.settings.RepositoryConfig); err == nil {
+		cacheDir := c.settings.RepositoryCache
+		for _, r := range f.Repositories {
+			idx, err := repo.LoadIndexFile(filepath.Join(cacheDir, fmt.Sprintf("%s-index.yaml", r.Name)))
+			if err != nil {
+				continue
+			}
+			entries, ok := idx.Entries[chartName]
+			if !ok {
+				continue
+			}
+			latest := ""
+			all := make([]string, 0, len(entries))
+			hasCurrent := false
+			for _, v := range entries {
+				all = append(all, v.Version)
+				if latest == "" || compareVersions(v.Version, latest) > 0 {
+					latest = v.Version
+				}
+				if v.Version == rel.Chart.Metadata.Version {
+					hasCurrent = true
+				}
+			}
+			if latest != "" {
+				candidates = append(candidates, repoVersionInfo{repoName: r.Name, repoURL: r.URL, latestVersion: latest, hasCurrentVersion: hasCurrent})
+				versionsByRepo[r.Name] = all
+			}
+		}
+	}
+
+	if len(candidates) > 0 {
+		sourceHosts := chartSourceHosts(rel.Chart.Metadata.Home, rel.Chart.Metadata.Sources)
+		if _, repoName := findBestUpgradeVersion(candidates, sourceHosts); repoName != "" {
+			return capVersions(sortVersionsDesc(versionsByRepo[repoName])), nil
+		}
+		// Ambiguous classic source — don't guess a version list.
+		return nil, nil
+	}
+
+	return capVersions(c.discoverOCIVersions(chartName)), nil
+}
+
+// maxAvailableVersions bounds the version list returned to the upgrade dialog.
+// Some charts publish hundreds of versions; the newest N covers realistic upgrade
+// targets without an unwieldy dropdown or a large payload. The list is already
+// sorted newest-first, so this keeps the most relevant versions.
+const maxAvailableVersions = 50
+
+func capVersions(versions []string) []string {
+	if len(versions) > maxAvailableVersions {
+		return versions[:maxAvailableVersions]
+	}
+	return versions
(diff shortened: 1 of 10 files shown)


nadaverell:
Triaged Bugbot's 3 findings — all valid, all fixed in a4de84e8:

1. **[High] OCI upgrade never reached** — `resolveUpgradeChartPath` aborted on a missing `repositories.yaml` before the OCI fallback, so a pure-OCI user saw a discovered update but the upgrade failed. Now treats a missing/unreadable repo file as an empty classic set and proceeds to OCI resolution. ✅
2. **[Medium] OCI masks classic index failures** — discovery (`checkForUpgrade` + batch) now surfaces a classic index-load failure *before* the OCI fallback, matching the upgrade path, so a broken repo index can't mislabel a classic-repo release as OCI-tracked. ✅
3. **[Medium] Registry errors cached in batch** — OCI discovery now caches only *successful* `Tags` lookups; a transient timeout is retried per release instead of poisoning the batch cache. Added a regression test. ✅

nadaverell:
Fixed the new finding in c007349c: added a structured `untracked` flag (set by the backend only in genuine no-source-found states), and the drawer now gates the clickable 'track source' affordance on it — repo-side errors (stale index / classic ambiguity) show a neutral 'upgrade source unresolved' chip instead of steering to OCI registration.

Note: the two earlier findings re-anchored to this commit ('OCI upgrade never reached', 'OCI masks classic index failures') were already addressed in a4de84e8 — their LOCATIONS still point at the pre-fix line ranges.

nadaverell:
## ✅ On-cluster verification (kind, real OCI registry)

Tested the full **OCI source** flow end-to-end against a disposable `kind` cluster with `podinfo` installed two ways — including a **real applied mutation** pulled from the OCI registry. Chart used: `oci://ghcr.io/stefanprodan/charts/podinfo` (public, dual-published).

**1. OCI-installed release with no configured repo → neutral "source not tracked" badge**
source not tracked

**2. Click it → Track chart source dialog (chart name interpolated; `helm registry login` note)**
track source dialog

**3. Register the real OCI prefix `oci://ghcr.io/stefanprodan/charts`**
source registered

**4. Badge re-resolves → "update available 6.14.0" via OCI discovery** (`sourceType: oci`, probed `<prefix>/podinfo` on ghcr.io)
update available

**5. Version picker populated with real OCI tags** (6.14.0 latest → 6.3.0 current → 6.1.0)
version picker

**6. Select an older target → "Downgrade" label + warning** (the semver-based downgrade detection)
downgrade state

**7. Applied on the cluster** — `podinfo-oci` now at **revision 2 / 6.1.0 deployed**, resources healthy (real mutation pulled from OCI)
applied result

### Notes
- Backend was independently confirmed via API: `untracked:true` before registration → `updateAvailable, sourceType:oci, chartRef` after; semver-sorted versions + 50-cap; link-local SSRF rejection.
- Console clean (no feature-related errors).
- **Observation (pre-existing, not this PR):** upgrades run with Helm `--reuse-values`, which doesn't merge a *newer* chart's new default value keys — so jumping podinfo 6.3.0→6.14.0 fails in the chart's own template (`index of untyped nil` on `.Values.hooks`). Radar surfaces that error cleanly in the progress log. The downgrade (older chart, fewer keys) applied cleanly. Worth considering `--reset-then-reuse-values` for the upgrade action separately.

_Screenshots hosted on branch `pr-1011-assets` (delete after merge, or I can re-attach via drag-drop for permanence)._

PR #1020, opened 2026-06-25
feat(probe): probe a Service's HTTP endpoint from the browser

nadaverell:
> **Stacked on #1017** (port-forward in-cluster gating). Review/merge that first; GitHub will retarget this to `main` once it lands. The diff below is Probe-only.

## Summary

When Radar runs in the cloud, port-forward can't work (the listener would live on the in-cluster Radar pod, unreachable from the browser — #1017 gates it off). But operators still need to answer "what does this Service's endpoint return / is it alive?" **Probe** fills that gap: click a Service's HTTP port → Radar issues a GET to it **from inside the cluster** and shows the response in the browser. No port-forward, no local listener, no cluster credentials.

## How it works

`POST /api/probe/service` issues the GET through the apiserver's **`services/proxy` subresource** using the **impersonated caller** config, so the **apiserver enforces RBAC** (`get services/proxy`) — there's no raw ClusterIP dialing and no arbitrary egress (the target is the named Service). It's also the same mechanism a future browser App-Access proxy would use, validated here at low risk.

**Security hardening (the parts worth reviewing):**
- **Strict input validation** — `namespace`/`name`/`port` are regex-validated before being spliced into the proxy URL, so a crafted value can't escape the `services/{target}/proxy` path.
- **Redirects are not followed** — client-go's transport re-attaches the cluster bearer token + `Impersonate-*` headers to *every* request, so following a Service-controlled `3xx` to another host would leak credentials (SSRF). The `3xx`+`Location` is returned to the caller instead. Covered by a regression test.
- **Read errors surfaced** — a target that sends headers then stalls until the 10s timeout is reported as an error, not a clean partial 200.
- GET-only · 10s timeout · 512 KiB body cap.

**Frontend:** a `Probe` button on HTTP-ish TCP Service ports (heuristic over `appProtocol` / port name / common HTTP port numbers; UDP excluded) opens a dialog — scheme, path, Send, and a response view (status, duration, bytes, headers, body, error/timeout banner).

## Testing

- `make tsc` ✓ · `go test ./internal/server/...` ✓ (incl. validation + redirect-not-followed security tests)
- Live cluster: `kube-dns:9153/metrics` → `200` + metrics body; no-endpoints Service → apiserver `503`; injection attempts → `400`.
- **Visual-tested** the button → dialog → response flow; confirmed the button shows on `http`/`https` ports and is correctly absent on a `metrics-udp` (UDP) port.

## Scope / follow-ups

MVP is **Services + GET only**. Deliberately out of scope (tracked separately): `pods/proxy`, non-GET methods, and **C — true TCP via radar-as-local-client** (the datastore path), which gets its own design. No new capability flag — apiserver RBAC enforces access and a `403` surfaces in the response.

---

> [!NOTE]
> **Medium Risk**
> New credentialed server-side HTTP proxy path with deliberate SSRF/credential-leak mitigations; scope is GET-only to named Services but still security-sensitive and worth careful review.
> 
> **Overview**
> Adds **in-cluster Service HTTP probing** for when live port-forward is not usable (Radar Cloud / in-cluster): operators can GET a Service from inside the cluster and see status, timing, headers, and a capped body in the UI.
> 
> The backend exposes **`POST /api/probe/service`**, which issues a single GET through the apiserver **`services/proxy`** subresource using the **impersonated caller** (RBAC on `get services/proxy`). Inputs are regex-validated, paths are normalized, redirects are **not** followed (avoids leaking bearer/impersonation headers on SSRF), responses are limited to **10s / 512 KiB**, and apiserver `Status` failures are surfaced as plain-English errors instead of raw JSON.
> 
> On the frontend, **HTTP-ish TCP ports** on Services get a **Probe** toggle with an **inline panel** (scheme, path, Send, response view). Shared **ServiceRenderer** gains **`renderPortPanel`** and richer port metadata for actions.
> 
> Related UX: **`useIsLocalDeployment()`** still shows port-forward affordances in non-local deployments (copy-paste **`kubectl port-forward`**), and copy in the in-cluster port-forward dialog is clarified.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a864e49f4860d193ab8de05af60dcae4e2aa5f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/server/probe.go
+++ b/internal/server/probe.go
@@ -1,0 +1,233 @@
+package server
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"regexp"
+	"strings"
+	"time"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/skyhook-io/radar/internal/auth"
+)
+
+const (
+	probeTimeout      = 10 * time.Second
+	probeMaxBodyBytes = 512 * 1024 // cap the previewed response body at 512 KiB
+)
+
+// dnsName matches a DNS-1123 subdomain (namespace/service names). Port accepts a
+// numeric port or a DNS-1123 port name. Both are validated before being spliced
+// into the apiserver proxy URL so a crafted value can't escape the
+// services/{target}/proxy path structure.
+var (
+	probeNameRe = regexp.MustCompile(`^[a-z0-9]([-a-z0-9.]{0,251}[a-z0-9])?$`)
+	probePortRe = regexp.MustCompile(`^[a-zA-Z0-9]([-a-zA-Z0-9]{0,14})?$`)
+)
+
+type probeRequest struct {
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`   // Service name
+	Port      string `json:"port"`   // port number or named port
+	Scheme    string `json:"scheme"` // "http" (default) or "https"
+	Path      string `json:"path"`   // request path, defaults to "/"
+}
+
+type probeResponse struct {
+	Status     int               `json:"status"`
+	DurationMs int64             `json:"durationMs"`
+	Headers    map[string]string `json:"headers"`
+	Body       string            `json:"body"`
+	Truncated  bool              `json:"truncated"`
+	BodyBytes  int               `json:"bodyBytes"`
+	// Error is set when the response was received but reading its body failed
+	// (e.g. the target stalled until the timeout). Status/body still reflect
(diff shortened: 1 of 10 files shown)


nadaverell:
Consolidated into #1017 — keeping in-cluster port-forward gating, the kubectl copy-command fix, and the HTTP probe in a single PR so #1017 doesn't merge as a UX degradation.

PR #1025, opened 2026-06-26
feat(audit): Traefik ref-integrity for chain middlewares + errors.service

nadaverell:
## Summary

Traefik ships no admission webhook or linter, so a typo'd reference is accepted silently and the route/middleware just doesn't do what you wrote — until someone reads the controller logs. Radar's Cluster Audit already catches the two most common cases (router → Service, router → Middleware). This extends that coverage to the next two low-hanging dangling-ref cases an operator actually hits:

- **`traefikChainMissingMiddleware`** — a `chain` Middleware whose `spec.chain.middlewares[]` lists a Middleware that doesn't exist. Traefik runs the chain but skips the missing link, so part of the intended pipeline silently doesn't apply.
- **`traefikErrorsMissingService`** — an `errors` Middleware whose `spec.errors.service` references a Service (or TraefikService) that doesn't exist, so users get the default error page instead of the custom one.

## What changed

- **`pkg/audit/traefik.go`** — the route-ref resolution was extracted into two shared closures (`checkServiceRef`, `checkMiddlewareRef`) so the new chain/errors subjects reuse the *exact* same matching: group-family isolation (a `traefik.io` ref isn't satisfied by a `traefik.containo.us` object), cluster-wide target inventory for legitimate cross-namespace refs, and the cluster-wide-synced **authoritative gate** so a namespace-scoped cache never fabricates a "missing" finding. Route behavior is unchanged.
- **Subject scoping** (`internal/audit/runner.go` + `CheckInput.MiddlewareSubjects`) — middlewares are gathered cluster-wide as *targets* (for cross-ns resolution) but only the ones in **audited namespaces** are treated as *subjects*. This mirrors how routes are already scoped and prevents a middleware in a non-audited namespace from leaking into a namespace-scoped scan.
- **Early-return guard** now also runs when a namespace has middlewares but no IngressRoutes (otherwise a middleware-only namespace would skip the new checks).
- **`pkg/audit/registry.go`** — catalog entries (title/category/description/remediation) for the two CheckIDs, mirroring the existing `traefikRouteMissing*` entries.

Still deliberately out of scope (not low-hanging): nested TraefikService weighted/mirror sub-services, ServersTransport refs, TLS-secret refs.

## Testing

- `pkg/audit` + `pkg/checks` + `internal/server` tests green; `go build ./...` + `go vet` clean.
- New table tests (`pkg/audit/traefik_test.go`): chain missing/present, chain cross-namespace resolves, chain non-authoritative → skip (no false positive), errors missing/present, `errors.service` with `kind: TraefikService`, and a regression guard that a **middleware-only namespace** (no routes) still runs the checks.

---

> [!NOTE]
> **Low Risk**
> Additive audit checks with conservative absence gating; existing route behavior is preserved via shared helpers.
> 
> **Overview**
> Extends cluster audit **Traefik dangling-reference** coverage beyond IngressRoutes to **chain** and **errors** Middlewares, with two new reliability checks: `traefikChainMissingMiddleware` and `traefikErrorsMissingService`.
> 
> **`checkTraefikDanglingRefs`** now audits Middlewares in audited namespaces via new **`MiddlewareSubjects`** (cluster-wide `Middlewares` still resolve cross-namespace targets). Route logic is refactored through shared **`checkServiceRef`** / **`checkMiddlewareRef`** helpers so chain/errors use the same group-family matching and authoritative-cache gates as routes. The runner loads cluster-wide **Services** when there are middleware subjects (not only routes), and the check no longer no-ops on middleware-only namespaces.
> 
> Registry entries and table tests cover missing/present refs, cross-namespace resolution, non-authoritative skip, and TraefikService-backed `errors.service`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dce38405dd2ee0c169bbc0402900595d3569c734. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/audit/runner.go
+++ b/internal/audit/runner.go
@@ -66,12 +66,13 @@ func RunFromCache(cache *k8s.ResourceCache, namespaces []string, opts *RunOption
 	// being fabricated as "missing". `authoritative` marks which target GVRs are
 	// served by a synced cluster-wide informer — only those let the check assert
 	// absence (see checkTraefikDanglingRefs).
-	input.IngressRoutes, input.Middlewares, input.TraefikServices, input.TraefikAuthoritativeKinds = listTraefikDynamic(namespaces)
+	input.IngressRoutes, input.Middlewares, input.TraefikServices, input.MiddlewareSubjects, input.TraefikAuthoritativeKinds = listTraefikDynamic(namespaces)
 	// Core Services for Traefik ref resolution. Only populate (→ only assert a
 	// missing-Service finding) when the Services informer is cluster-wide, so
 	// absence is authoritative for every namespace — same coverage gate the
 	// dynamic Traefik kinds use. Namespace-scoped fallback → leave nil → skip.
-	if len(input.IngressRoutes) > 0 && cache.IsKindClusterWide("services") {
+	// Middleware subjects (errors.service refs) need it too, not just routes.
+	if (len(input.IngressRoutes) > 0 || len(input.MiddlewareSubjects) > 0) && cache.IsKindClusterWide("services") {
 		input.AllServices = listNamespaced(cache.Services(), nil)
 	}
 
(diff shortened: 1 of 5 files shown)


nadaverell:
**Live smoke test — green.** Ran the branch binary against a kind cluster with real `traefik.io` CRDs (not just the unit-test's hand-built `CheckInput`), to exercise the two seams the unit tests don't cover: the `runner.go` `listTraefikDynamic` gather/scoping (`MiddlewareSubjects` + authoritative gate) and real-CRD unstructured parsing of `spec.chain.middlewares[]` / `spec.errors.service`.

Fixtures: a chain Middleware referencing a missing middleware, an errors Middleware referencing a missing service, plus clean controls (chain→existing middleware, errors→existing service).

`GET /api/audit`:
- `bad-chain` → `traefikChainMissingMiddleware` ✅
- `bad-errors` → `traefikErrorsMissingService` ✅
- `good-chain` / `good-errors` → no finding ✅ (no false positives)

Per-resource drill-down (`/api/audit/resource/Middleware/vt-audit/<name>`) returns both findings, confirming the empty-Group keying reaches the drawer. Net: runner wiring + real-object parsing validated end-to-end, no false positives/negatives.

PR #1028, opened 2026-06-27
feat(audit): badge resources with high-signal Cluster Audit findings (topology + list)

nadaverell:
## Summary

A resource that's **actually broken** — a dangling Traefik/Ingress backend ref, a Service whose selector resolves to no pods, a deprecated API that breaks on upgrade — was invisible everywhere an operator looks: it showed only in its own drawer, while the topology graph and resource list rendered it healthy. This surfaces those findings as a per-resource badge on both surfaces, fixes a pre-existing bug that hid built-in findings from the drawer, and removes a column that was contradicting the badge.

Composes with the health consolidation (#1024): that owns *runtime* health (is it running); this owns *cross-resource ref-integrity + upgrade-risk* (does its config reference things that exist). Different axes, no overlap.

## Keeping it a signal, not noise

Audit also emits posture/best-practice findings (`runAsRoot`, missing limits, single-replica) that fire on nearly every resource — *any-finding* badged **112** resources on a real staging namespace. So badges are gated on a registry **`badgeWorthy`** flag, true only for the 8 checks that mean a specific resource is broken: the 4 Traefik dangling-ref checks, `ingressNoMatchingService`, `serviceNoMatchingPods`, `deprecatedAPIVersion`, `crossplaneStuck`. Same namespace: **112 → 15** badged resources. (`stuckTerminating` was intentionally excluded — #1024's health pipeline already renders it degraded, and the name cell shows a "Terminating" chip; a badge would be a third redundant signal.)

## What changed

**Backend**
- `badgeWorthy` flag on the check registry (`pkg/checks` `CheckMeta` + `pkg/audit/registry.go`), exposed via the audit catalog the frontend already receives — curation lives with the checks, no drifting frontend list.
- Topology builder stamps each node with `data.auditKey` (`group|Kind|ns|name`, audit convention; collision pseudo-kinds like `IstioGateway`→`Gateway` remapped) so the frontend joins findings by one string lookup.
- **Drawer fix**: the per-resource endpoint looked up `group=""`, so built-in findings keyed under a real group (`Deployment`→`apps`) never reached the drawer. Now resolves via `GroupForBuiltinKind`.

**Frontend**
- **Topology** (`K8sResourceNode`) + **resource list** (`ResourcesView`): a per-resource indicator (triangle / name-cell badge) for badge-worthy findings only, colored by worst severity (danger red / warning amber). Injected via the existing context; audit is one React-Query fetch, deduped. Hovering a badge lists the actual finding messages inline (shared `AuditBadgeTooltip`, danger-first, capped) — the operator reads *what* is wrong without opening the drawer.
- **Service Endpoints column**: previously guessed a green "Active" for any service with a selector (it can't see live pods), contradicting the `serviceNoMatchingPods` badge on the same row. Now consumes that audit check and shows "No endpoints" when the selector resolves to nothing — column and badge agree.

## Testing

- Rebased on #1024 (clean). `make tsc` ✓ · k8s-ui vitest **674** (incl. ratchet + `stampAuditKeys`) ✓ · Go `pkg/{checks,audit,topology,health}` + `internal/server` ✓
- **Live-verified** (real GKE cluster): curation cuts **112 → 15** badged resources; Deployments list shows **0** badges (posture noise gone); a broken Service badges + its Endpoints column reads "No endpoints" instead of green "Active"; drawer endpoint returns built-in findings (was 0 pre-fix). Light + dark.

---

> [!NOTE]
> **Medium Risk**
> Touches audit API lookup semantics and topology/list UX; changes are additive with defensive fallback lookup, but incorrect keying could still miss or mis-attach badges.
> 
> **Overview**
> High-signal **Cluster Audit** findings (broken refs, missing backends, deprecated APIs) now show on the **topology graph** and **resource list**, not only in the audit drawer.
> 
> A registry **`badgeWorthy`** flag limits badges to reference-integrity / lifecycle checks so posture noise (`runAsRoot`, missing limits, etc.) stays in Audit views. Topology nodes get **`data.auditKey`** from the builder; the web app joins findings via `buildAuditSeverityMap` and injects counts into `K8sResourceNode` and `ResourcesView` name cells. **Service** rows override the optimistic green **Endpoints** column when `serviceNoMatchingPods` fired.
> 
> **Drawer fix:** `handleAuditResource` now resolves built-in kinds with `GroupForBuiltinKind` instead of `group=""`, so findings keyed under `apps`, `batch`, etc. appear again.
> 
> `AuditFinding.group` and `CheckMeta.badgeWorthy` are exposed on the shared k8s-ui types for joins.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40d9143cae7833d94fc21413139492000f989e79. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/server/audit_handlers.go
+++ b/internal/server/audit_handlers.go
@@ -166,17 +166,22 @@ func (s *Server) handleAuditResource(w http.ResponseWriter, r *http.Request) {
 	}
 	index := bp.IndexByResource(results.Findings)
 
-	// Try exact kind first, then map API resource name (e.g. "deployments") to Go kind (e.g. "Deployment").
-	// This handler is the UI's per-resource audit drill-down — group isn't on
-	// the URL today (the UI doesn't list grouped CRDs here yet), so we look
-	// up with group="" which matches the built-ins the audit suite scans.
-	// When CRD audit lands (#35 follow-up), thread group through the URL.
-	findings := index[bp.ResourceKey("", kind, namespace, name)]
-	if findings == nil {
-		goKind := apiResourceToKind(kind)
-		if goKind != kind {
-			findings = index[bp.ResourceKey("", goKind, namespace, name)]
-		}
+	// Resolve to the Go kind (built-in plural "deployments" → "Deployment"), then
+	// look up with the audit's keying convention. Findings carry group backfilled
+	// from the builtin table (built-ins → their group e.g. "apps"; CRDs → ""), so
+	// a bare group="" lookup missed every built-in finding keyed under a real
+	// group (Deployment under "apps", Job under "batch", …) — the drawer showed
+	// nothing for them. Group still resolves to "" for CRDs, so those keep working.
+	goKind := kind
+	if mapped := apiResourceToKind(kind); mapped != kind {
+		goKind = mapped
+	}
+	group := bp.GroupForBuiltinKind(goKind)
+	findings := index[bp.ResourceKey(group, goKind, namespace, name)]
+	if findings == nil && group != "" {
+		// Defensive: resolve even if a finding for this kind was emitted with an
+		// empty group despite the builtin table classifying it under a real one.
+		findings = index[bp.ResourceKey("", goKind, namespace, name)]
 	}
 	if findings == nil {
 		findings = []bp.Finding{}
--- a/packages/k8s-ui/src/components/audit/AuditAlerts.tsx
+++ b/packages/k8s-ui/src/components/audit/AuditAlerts.tsx
@@ -5,6 +5,10 @@ import { SEVERITY_TEXT, BP_CATEGORY_BADGE, DEFAULT_BADGE_COLOR } from '../../uti
 
 export interface AuditFinding {
   kind: string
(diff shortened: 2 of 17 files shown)


nadaverell:
## Visual verification

Captured against a local `kind` cluster with synthetic badge-worthy fixtures (a Service whose selector matches no pods, and an Ingress pointing at a missing backend Service).

**Resource list — Ingress badge.** Confirms the plural-URL-kind fix: `Ingress` (a grouped kind) now resolves its canonical kind, so the amber audit badge renders next to `storefront`.

Ingress list badge

**Resource list — Service.** The name-cell badge and the ENDPOINTS column now read the audit `serviceNoMatchingPods` finding: the column shows the orange "No endpoints" pill instead of the old guessed green "Active".

Service endpoints

**Topology.** Worst-severity audit triangle on the `web` Service and `storefront` Ingress nodes; the healthy `api` Deployment / Pod carry no indicator.

Topology audit indicator

PR #1033, opened 2026-06-27
feat(issues): surface live Operational Issues in the resource detail

nadaverell:
## Summary

A resource that is *actually broken* — a Pod in `CrashLoopBackOff`, a Deployment stuck below desired replicas, an Argo app failing to reconcile — only told the full story in the cluster-wide Issues queue. Open that resource's own detail and you got raw status fields, not a diagnosis. This adds a **live Operational Issues section to the resource detail**, so the same runtime diagnosis the Issues queue shows (severity + cause/action) appears in context, at the top of the overview, the moment you open the thing that's failing.

## What changed

**Backend** — `GET /api/issues/resource/{kind}/{namespace}/{name}` (`internal/server/issues_handler.go`). Reuses the existing `RelatedIssues` compose primitive (namespace-scoped), the same drawer-equivalent RBAC pre-flight as resource reads (`preflightResourceGet`), and canonical-kind resolution. Returns the resource's live issues, or empty for a healthy resource.

**Frontend** — a `useResourceIssues` hook and a compact `ResourceIssuesSection` (the diagnosis sibling of the Issues-queue row, without the queue chrome). It mounts through a new **additive, optional** `renderOverviewLead` slot on the pure `WorkloadView`, wired by the web host to render above the renderer in both the compact drawer and the expanded view. Kind-agnostic (`RelatedIssues` covers all kinds); hidden when there are no live issues.

**Dedup without deleting the fallback.** When the section is shown, `PodRenderer` and `WorkloadRenderer` — the kinds the Issues pipeline covers comprehensively — gate out their own status-derived "Issues Detected" banner (via `OperationalIssuesShownContext`, default-false). Their detection logic (`getPodProblems` / `getWorkloadProblems`) is **retained, not removed**: it still fires anywhere the section isn't rendered (library consumers of `@skyhook-io/k8s-ui`; any mount outside radar's `WorkloadView`, which is the only place that sets the context) and as a safety net if the pipeline ever misses something. So a `CrashLoopBackOff` shows once in the richer section, but nothing silently disappears elsewhere.

The shared `ProblemAlerts` component (used **only** by the GitOps renderers, whose Degraded / OutOfSync / revision-mismatch banners the pipeline does not fully emit) deliberately does **not** gate on the context — suppressing those would lose real warnings, which is worse than an occasional duplicate. Kinds with no self-detection (Ingress / Service / HTTPRoute / …) were never duplicated; the section is purely additive there.

**Operator-facing copy.** Since these issues now show in the resource detail, the queue, and MCP, the copy got a quality pass: the card headlines on the plain category label with the raw `reason` as a muted signal (+ a de-emphasized Evidence line so the precise ref/path isn't lost); expert category labels are plain-languaged (Unschedulable → "Can't be scheduled", RBAC forbidden → "Permission denied", …); and the **full dangling-ref family** (pod Secret/ConfigMap/PVC/SA/imagePullSecret, Ingress backend/port/TLS, StatefulSet headless Service, HPA target, webhook backend, Gateway backend/port/ReferenceGrant, PVC StorageClass, RBAC roleRef, Argo Rollout Service, KEDA scaleTargetRef) now carries a structured cause + namespace-qualified next-step action. `reason` stays machine-stable (CEL/category/wire contract unchanged).

## Testing

- `make tsc` ✓ · k8s-ui vitest **737** ✓ (incl. regression tests pinning `ProblemAlerts` visibility under the context and `ResourceIssuesSection`) · Go `internal/server` + `pkg` ✓
- Cross-reviewed by Codex (plan + code, and again after rebasing onto #1028); converged.
- **Visual-tested** on the GitOps demo cluster:

A crashlooping Deployment shows the `CrashLoopBackOff` issue once, with the renderer's own status-problem banner suppressed (no duplicate):

Workload dedup

An OutOfSync Argo app (`guestbook-drift`, live `StuckDriftLoop` issue) shows the Operational Issues section **and** its "out of sync with git" warning banner together — confirming GitOps banners are not suppressed:

GitOps issues + un-suppressed banner

## Notes / follow-up

- Per-drawer `RelatedIssues` is namespace-scoped (same cost model as the main queue); no polling (staleTime-only fetch). A per-scope TTL memo is a possible future optimization.
- Pod/Workload suppression is per-resource all-or-nothing. A cleaner future option is for the GitOps renderers to gate only the *specific* problems the pipeline also emits (rather than always showing them), eliminating the occasional GitOps duplicate without risking hidden warnings.

---

> [!NOTE]
> **Medium Risk**
> New authenticated issues endpoint and UI gating affect what operators see on resource open; detection/message changes are additive but touch broad issue surfacing paths.
> 
> **Overview**
> **Resource detail** now loads **live Operational Issues** at the top of the overview via `GET /api/issues/resource/{kind}/{namespace}/{name}` (`RelatedIssues`, drawer-equivalent RBAC, canonical kind resolution) and a compact **`ResourceIssuesSection`** wired through optional `renderOverviewLead` / `hasOperationalIssues` on `WorkloadView`.
> 
> When that section has rows, **`OperationalIssuesShownContext`** suppresses duplicate **Pod** and **Workload** status problem banners; **GitOps `ProblemAlerts` stay visible** so manual OutOfSync-style warnings are not hidden.
> 
> **Detector copy** is expanded so findings surface operator **`Cause`** + **`Action`**: `withFix` / ref diag helpers across missing-reference detectors, targeted **`Action`** on core problems (rollouts, LoadBalancer, PDB, PVC/PV lost/failed, stuck terminating), **Argo CD** next steps, and **unschedulable** pods via a single-parse **`diagnoseUnschedulable`**. Issue **category labels** in the UI are plain-languaged; tests assert missing-ref **Cause/Action** and unschedulable actions are non-empty.
> 
> **.gitignore** adds local visual-test screenshot dirs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a061dab40596c9af9c28af7e8d88e6a10a4dbcf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
## Visual-test (post-rebase, GitOps demo cluster)

Verified the live Operational Issues surface and the round-2 fix (don't suppress GitOps `ProblemAlerts`).

**GitOps app — fix confirmed.** `guestbook-drift` is OutOfSync with a live issue (`StuckDriftLoop`). The detail shows the **Operational Issues** section *and* the renderer's **"Application is out of sync with git"** warning banner together. Before the fix that banner was suppressed by the issues context (info loss); now it renders.

GitOps issues + un-suppressed banner

**Workload — D1 dedup intact.** A crashlooping Deployment shows the `CrashLoopBackOff` Operational Issue once; the workload renderer's own status-problem banners are suppressed (no duplicate).

Workload dedup

nadaverell:
### Bugbot triage (fix-pr-loop)

All findings addressed. Cursor Bugbot re-anchors unresolved threads to the latest commit, so the three below keep reappearing on HEAD even though they're resolved — recording the disposition here:

| Finding | Verdict | Resolution |
|---|---|---|
| Incomplete plural kind normalization | **Fixed** | `handleResourceIssues` now resolves the kind via discovery (every kind + CRDs), and the partial `apiResourceToKind` seed was removed entirely. Verified live: plural `persistentvolumeclaims` / `applications` resolve and return issues. |
| Empty lead slot adds padding | **Fixed** | The overview-lead wrapper is gated on `hasOperationalIssues`, so healthy resources no longer get an empty padded gap. Verified live (0 empty lead divs on a healthy workload; section still renders on an unhealthy one). |
| Issues query key destabilizes | **Fixed** | `useResourceIssues` now keys on the stable prop kind+group (same inputs as the resource fetch), not the manifest-derived ones, so the key no longer flips on load. (Bugbot dropped this from HEAD.) |
| Cluster issues skip RBAC filters | **Skip (not exploitable)** | `RelatedIssues` exact-matches the target (already authorized by `preflightResourceGet`); groups are owner-collapsed so members share the subject's namespace, and node/PV detections are single-resource — so no returned group can carry a forbidden-namespace member. The per-resource drill-down's gate is the target preflight, by design. |

PR #1050, opened 2026-06-29
feat(workload): expand resource peek into a context-retaining fullscreen overlay with a smooth morph

nadaverell:
## Summary

Expanding a resource detail peek used to **navigate away** to the standalone `/workload` route — whatever you were triaging (the resource list, a topology graph, GitOps tree) vanished, and the drawer→fullscreen transition **stuttered** (the fullscreen layout reflowed live inside a still-growing box while collapse hard-snapped).

Now expand **grows the peek into a fullscreen overlay on the same route**, keeping the view underneath mounted, with a smooth GPU-composited morph. You stay in context: browser Back collapses, Forward re-expands, refresh persists the expanded state. `/workload` remains the canonical standalone route for deep-links and non-list surfaces.

## What changed

**Context-retaining overlay (`App`).** Expand sets `?full=1` on the current route (plus `?tab=yaml` when expanding from the drawer's YAML view) instead of pushing `/workload`. The underlying view — resource list, topology graph, GitOps, Applications — stays mounted beneath the overlay. Collapse, back, and close scrub those params in place. Drilling into a related resource while expanded updates `/resources/{kind}?…&full=1`, so browser Back walks resource→resource still expanded. Closing a peek always drops `full`/`tab` so the next peek opens as a side drawer, never a stale fullscreen.

**Smooth morph (`ResourceDetailDrawer`).** A ~260ms frame-morph + content crossfade: both the drawer and fullscreen layouts mount during the transition, each **pinned to its own measured width** so neither reflows or squashes, and the container width animates while their opacity crossfades (transform/opacity only — no per-frame layout). Symmetric expand and collapse; `prefers-reduced-motion` snaps instantly.

**Pre-mount on hover.** Hovering or pressing the expand control pre-mounts the heavy fullscreen layer invisibly (`inert`, `opacity-0`, pinned to full width) so the click reuses it and the morph's first frames aren't competing with the mount cost. The pre-warm is dropped on pointer-leave, close, reduced-motion, and resource change.

**Keyboard + a11y.** `useSuppressBaseShortcuts` drops view-level shortcuts while the overlay covers them (`global` + `drawer`/Escape still resolve); `WorkloadView`'s new `active` prop gates the hidden layer's shortcuts during the morph so it can't capture Escape/`y`/`e`; the backdrop content is `inert` when expanded so it isn't Tab-reachable.

**Mobile.** No side drawer — resource detail is always fullscreen, with a single close affordance (no collapse-to-drawer control).

**URL hygiene (`ResourcesView` + navigators).** List navigators and kind switches drop stale `full`/`tab`, so an expand flag from one peek never leaks onto the next.

## Testing

- `make tsc` ✓ · `make test` (Go + k8s-ui vitest) ✓
- Live-tested on a `kind` cluster, **desktop + mobile (390px)**, on both `/resources` and a **topology** surface (where the peek isn't URL-backed — the harder case):
  - peek → expand → collapse / back / close; the underlying view stays mounted throughout; the fullscreen workload view renders correctly over the graph
  - browser Back collapses, Forward re-expands; refresh persists `?full=1`
  - a fresh peek after an expand→close cycle opens as a **drawer** (no stale-fullscreen carryover), verified on both desktop and the mobile close path
  - hover **pre-mounts** the fullscreen layer (inert, opacity-0, full width) and **cancels on leave**
  - mobile flips to always-fullscreen with a single close control

## Notes

- **Deep-link asymmetry (intentional):** a `?full=1` link reopens fullscreen-over-list; a `/workload` link opens the standalone fullscreen. Both land somewhere sensible.
- Embedded/library consumers that don't wire the expand flow keep the original drawer behavior — the new props are additive.

### DIFF
--- a/packages/k8s-ui/src/components/resources/ResourcesView.tsx
+++ b/packages/k8s-ui/src/components/resources/ResourcesView.tsx
@@ -2951,6 +2951,11 @@ export function ResourcesView({
       params.set('resource', resourceNs ? `${resourceNs}/${resourceName}` : resourceName)
     } else {
       params.delete('resource')
+      // `full` (over-list fullscreen) and `tab` are resource-scoped — when no
+      // resource is selected they're stale; drop them so they can't leak onto a
+      // later selection (e.g. after a kind switch or closing the drawer).
+      params.delete('full')
+      params.delete('tab')
     }
 
     const newPath = `${basePath}/${kindInfo.name}`
--- a/packages/k8s-ui/src/components/resources/ResourcesView.tsx
+++ b/packages/k8s-ui/src/components/resources/ResourcesView.tsx
@@ -1944,10 +1944,6 @@ interface ResourcesViewProps {
   isBulkRestarting?: boolean
   onBulkScale?: (items: BulkResourceItem[], replicas: number, options?: { onSuccess?: () => void }) => void
   isBulkScaling?: boolean
-  /** When false, this list's keyboard shortcuts are not registered — used when a
-   *  fullscreen detail overlay sits on top of the still-mounted list so its
-   *  row-nav / Escape / action keys don't fire on the hidden background. Default true. */
-  shortcutsActive?: boolean
 }
 
 // Default selected kind
@@ -2068,7 +2064,6 @@ function LastUpdatedLabel({ lastUpdated }: { lastUpdated: Date }) {
 
 export function ResourcesView({
   namespaces, selectedResource, onResourceClick, onResourceClickYaml, onKindChange,
-  shortcutsActive = true,
   apiResources: apiResourcesProp,
   resourceQueries: resourceQueriesProp,
   resourceCounts: resourceCountsProp,
@@ -2514,7 +2509,6 @@ export function ResourcesView({
     category: 'Search',
     scope: 'resources',
     handler: () => searchInputRef.current?.focus(),
-    enabled: shortcutsActive,
   })
 
   // Keyboard navigation: highlighted row state
@@ -2614,8 +2608,8 @@ export function ResourcesView({
     onResourceClick?.(isSelected ? null : stripped)
   }, [onRowSelect, onResourceClick, selectedKind.name, selectedKind.group])
 
-  // Register navigation shortcuts (suppressed while a fullscreen overlay covers the list)
-  useRegisterShortcuts(shortcutsActive ? [
+  // Register navigation shortcuts
+  useRegisterShortcuts([
     {
       id: 'resources-nav-down',
       keys: 'j',
@@ -2788,7 +2782,7 @@ export function ResourcesView({
         else searchInputRef.current?.blur()
       },
     },
-  ] : [])
+  ])
 
   // Refs for accessing filteredResources inside shortcuts (computed later in component)
   const filteredResourceCountRef = useRef(0)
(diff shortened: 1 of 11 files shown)


nadaverell:
Fixed in a27c1248 — rather than only suspending the outgoing layer, both layers are now inactive for the whole crossfade, so neither the outgoing nor the incoming drawer layer dispatches Escape/`y`/`e` mid-animation. Shortcuts resume on the single settled layer once the transition ends.

nadaverell:
Triaged the 5 Bugbot findings — all intentional or known minor edges, no correctness bugs:

1. **Drill wipes query** — building fresh params is *correct* for a cross-kind drill (cloning would carry the source kind's column filters/search onto a different kind's list and break it). The global `namespaces` scope is preserved.
2. **Flag drops before URL** (Low) — requiring `selectedResource` for `drawerExpanded` is deliberate/defensive; dismiss-via-kind-switch works and doesn't leak `full` (verified). A brief morph as the drawer exits is cosmetic.
3. **List shortcuts during collapse morph** — real but minor: a ~360ms window during a user-initiated collapse where the list is the destination anyway. Acceptable; tracked as a possible follow-up to suppress through the full morph.
4. **Detail state lost after morph** — inherent to the crossfade's keyed dual-layer mount (deliberate design); URL-backed state (tab/resource) persists, only transient local UI state resets.
5. **Aborted expand snaps** — no broken state (timers/rAF are cleaned up); a sub-360ms expand→reverse resolving to the final state is acceptable interrupt UX.

CI is green; the conflict-with-main that was blocking the merge-ref (and thus CI) is resolved.

nadaverell:
Triaged the 6 Bugbot findings on the latest commit:

**Fixed (2 new):**
- *Overlay gate ignores drawer visibility* — `expandedView` now keys off `routeSelectedResource` (the actual render gate), so an orphaned/mismatched/inline-workload peek can't inert the shell with no visible drawer.
- *Close triggers collapse morph* — the drawer's `expanded` prop is held through the slide-out, so dismissing an expanded drawer no longer runs a collapse crossfade mid-exit.

**Re-posts, already dispositioned (skip, evidence prior in thread):**
- *Drill navigation wipes query* — intentional: cross-kind drill goes to a different kind's list; cloning would carry incompatible filters. Global `namespaces` is preserved.
- *Detail state lost after morph* — inherent to the crossfade's keyed dual-mount; URL-backed state (tab/resource) persists, only transient local UI resets.
- *Aborted expand skips morph cleanup* — no broken state (timers/rAF cleaned up); sub-360ms reversal resolving to final state is acceptable interrupt UX.
- *Escape during collapse crossfade* — narrow (~360ms), user-initiated, list is the destination; accepted.

nadaverell:
Re-triage of the latest Bugbot pass (it re-posts all findings each run; no new *valid* ones):

- **Close triggers collapse morph** — already fixed: the App holds the drawer's `expanded` prop through the slide-out (`drawerExpandedProp`), so closing an expanded drawer does not flip `expanded`→false and never starts a collapse morph. (Bugbot analyzes ResourceDetailDrawer in isolation and can't see the App-level held value.)
- **YAML lost on drawer expand** — false positive; verified live: `?full=1&tab=yaml` opens the expanded view on the **YAML** tab. The expanded tab is controlled by `activeTab={migratedTab}` (web WorkloadView), which reads `?tab`; Bugbot traced the collapsed `showYaml`/`initialTab` path instead. (Unsaved edits also survive regardless via the session draft.)
- The other 4 (Drill query, Detail-state-lost, Aborted-expand, Escape-during-collapse) are prior re-posts, dispositioned above.

PR #1056, opened 2026-06-29
feat(issues): symptom→root cause pointers + secret-producer link

nadaverell:
## Summary

When an operator lands on a *symptom* in the Issues list ("this Deployment can't schedule", "this HPA stalled"), the next question is always **why**. Radar's causal links were root-anchored only — a failing root (broken PVC, pressured node, down metrics API) was annotated with the symptoms it explains, but the symptom row had no pointer *back* to its cause. This adds that reverse pointer, `incident_parent`, so symptom→root navigation works in one click. It's the highest-value, lowest-risk slice of "incident grouping" — **navigation, not row-hiding** (no demotion; nothing is suppressed or reordered).

It also adds one **new causal link family** so the pointer covers more than the three existing links.

> Framing: these are **root-cause pointers / cause hints**, not visual "incident grouping" (no row compression). True demotion is deliberately deferred (see below).

## What's covered

A symptom row gains `incident_parent` for the causal links whose related issues are genuinely **downstream** of the root:

| Root | → Symptom | Confidence | UI lead-in |
|------|-----------|------------|------------|
| **PVC** pending/lost/resize-failed | pods mounting it (volume-evidenced) | high | "Caused by" |
| **Node** under MemoryPressure/DiskPressure/PIDPressure | pods on it (reason-gated) | medium | "Related" |
| **metrics APIService** unavailable | HPAs starved of that metric family | medium | "Likely cause" |
| **Secret producer** — cert-manager `Certificate` (`spec.secretName`) / external-secrets `ExternalSecret` (`spec.target.name`) — **NEW** | pods referencing the Secret it owns and failing for a config/secret reason | high | "Caused by" |

`selected_backend` (Service with no endpoints) is **excluded** — there the related pods are the *cause*, not a downstream symptom, so reversing it would invert causality.

## How it stays precise (over-linking is the cardinal sin)

- **Causal-direction allow-list**, not confidence-gating: only the four downstream links above record a reverse edge.
- **Best-parent, no guessing**: when several roots name one symptom, a higher confidence tier wins (a declared PVC edge over a co-located node); **distinct roots at the same tier leave the pointer unset** — severity is never used to pick a cause.
- **Whole-row coverage**: a grouped symptom (e.g. a Deployment folding N pods) gets the pointer only when **every** member is explained by the root (`affected ≥ fan-out`). A Deployment whose pods are only partly PVC-blocked, or split across two pressured nodes, is omitted rather than over-claimed. This also subsumes the multi-root ambiguity case.
- **Inherited gates**: the pointer rides the existing forward-link precision (PVC volume-evidence, node reason-gate + dead-kubelet exclusion, HPA metric-family match + missing-request exclusion). The new Secret link requires the symptom to **name the Secret** (so a pod that references it via env but fails on an unrelated volume isn't attributed).
- **Self-edge / cycle guards** (structurally impossible with the current roots, but cheap).
- Edge computation is **uncapped** — independent of the 5-item display cap — so no pointer is dropped and the cap can't hide ambiguity.

## Surfaces

- **Cluster Issues view**: a non-interactive cause-signal in the collapsed row header ("↳ Caused by PersistentVolumeClaim / data") + a clickable **"CAUSED BY"** section (with confidence) in the expanded body. Honest per-link wording (declared edges say "Caused by"; co-located node says "Related"). Cluster-scoped deep-link via the existing `memberRef`.
- **MCP `issues`**: symptom rows carry `incident_parent` (id + ref + confidence + fact_type) so an agent can walk symptom→root directly. Description updated.

## New backend surface
- `issuesapi.IncidentParent` + `Issue.IncidentParent` (additive, omitempty). No `Demote` field — demotion is presentation policy, not part of this contract.
- Provider methods `PodsReferencingSecret` and `PodsDependingOnSecretProducer` (resolves the producer CR's target Secret via the dynamic cache).

## Testing
- `go test ./internal/issues ./internal/mcp ./internal/server` + the `pkg/issuesapi` module — green; `make tsc` green.
- Unit tests pin: PVC (high) / node (medium) reverse pointers; `selected_backend` excluded; best-parent rule (high>medium, distinct same-tier → omit) as a pure unit; **whole-row coverage gate (3/3 links, 2/3 omits)**; secret-producer links only the Secret-naming pod; the root never points at itself.
- **Live-verified on kind — 3 of 4 link families end-to-end:**
  - **PVC** (high): pending PVC + 3-replica Deployment → the `unschedulable` issue gained `incident_parent → PersistentVolumeClaim/data`, rendered as the header chip + "CAUSED BY · HIGH CONFIDENCE" body section (0 console errors).
  - **apiservice→HPA** (medium): metrics-server scaled to 0 → the HPA's `cannot-scale` issue gained `incident_parent → APIService/v1beta1.metrics.k8s.io`.
  - **secret-producer** (high): cert-manager `Certificate` referencing a missing issuer + a pod mounting its Secret → the Deployment's `missing_config_ref` and `volume_mount_failed` issues gained `incident_parent → Certificate/web-tls`; the tightened gate proved correct (the mount failure linked *because* its kubelet message names `secret "web-tls-secret"`, not unconditionally).
- Cross-reviewed with Codex over multiple rounds; every finding (edge/cap coupling, secret gate, whole-row coverage, an unsound provisioning gate, a same-name secret false-positive) is fixed and pinned by tests.
- **Unit-tested-only**: the **node** blast-radius pointer (needs a multi-node cluster + induced MemoryPressure/NotReady), the same honest gap as the forward link.

## Deferred (out of scope, on purpose)
- **Per-resource drawer back-pointer**: the per-resource regroup can't reconstruct whole-row coverage (members share an issue ID), so the pointer ships on the Issues view + MCP only for now.
- **node-provisioning → unschedulable link**: its scheduler-message gate is fundamentally imprecise (cluster-autoscaler's "didn't trigger scale-up" includes taint/affinity reasons); needs the scheduler reason-class threaded onto detections first.
- **Demotion / true incident grouping** (root promoted, symptoms nested/hidden): a separately-validated policy needing relation-specific suppressibility — this PR is the non-destructive navigation foundation it builds on.
- More cross-subject links (cert→Ingress-TLS, admission-webhook fan-in): additive, each will auto-gain a reverse pointer.

---

> [!NOTE]
> **Medium Risk**
> Changes causal attribution and navigation for operators/agents; logic is heavily gated but wrong parent hints could misdirect triage.
> 
> **Overview**
> Symptom issue rows can now carry **`incident_parent`**, a reverse pointer to the root that explains them (id, ref, confidence, fact type), built from existing root→symptom causal links for PVC, node blast radius, metrics APIService→HPA, and a **new** cert-manager **Certificate** / external-secrets **ExternalSecret** → dependent pods link (`secret_not_ready`). Assignment runs only on the **cluster grouped** issues path: it picks a single parent when confidence is unambiguous (high beats medium; tied distinct roots omit the field), requires **whole-row coverage** for grouped symptoms, and records edges **uncapped** separately from the display cap on related issues.
> 
> **CacheProvider** gains **`PodsReferencingSecret`** and **`PodsDependingOnSecretProducer`** to resolve producer CRs to target Secrets and referencing pods. **`foldGroup`** deliberately does not propagate `incident_parent` on per-resource regroup.
> 
> The **Issues** UI shows a collapsed-row cause hint and an expanded **Caused by / Related** section; **MCP** `issues` tool docs describe `incident_parent`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62708596097c2451952cb6a13d1bdf1f4f611a32. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/issues/diagnostic_context.go
+++ b/internal/issues/diagnostic_context.go
@@ -121,6 +121,7 @@ func enrichDiagnosticContext(shaped, flat, grouped []Issue, p Provider) []Issue 
 	}
 
 	out := append([]Issue(nil), shaped...)
+	var incidentEdges []incidentEdge
 	for idx := range out {
 		var b diagnosticContextBuilder
 		i := &out[idx]
@@ -185,15 +186,15 @@ func enrichDiagnosticContext(shaped, flat, grouped []Issue, p Provider) []Issue 
 		}
 
 		if nodeProvider != nil && i.Kind == "Node" && i.Category == issuesapi.CategoryNodeNotReady {
-			addNodeBlastRadiusContext(&b, *i, nodeProvider, flatByResource, groupedByID)
+			addNodeBlastRadiusContext(&b, *i, &incidentEdges, nodeProvider, flatByResource, groupedByID)
 		}
 
 		if pvcProvider != nil && i.Kind == "PersistentVolumeClaim" && pvcRootCategories[i.Category] {
-			addPVCBlastRadiusContext(&b, *i, pvcProvider, flatByResource, groupedByID)
+			addPVCBlastRadiusContext(&b, *i, &incidentEdges, pvcProvider, flatByResource, groupedByID)
 		}
 
 		if metricsAPIFamily(*i) != "" {
-			addAPIServiceHPAContext(&b, *i, flat, groupedByID)
+			addAPIServiceHPAContext(&b, *i, &incidentEdges, flat, groupedByID)
 		}
 
 		if ctx := b.build(); ctx != nil {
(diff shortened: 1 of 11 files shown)


nadaverell:
Fixed in d46ee149c — `incident_parent` is now assigned only on the cluster grouped path; the ungrouped `?view=flat` / per-resource paths (where members share an issue ID and coverage can't be checked) deliberately leave it unset. Pinned by `TestSymptomNamesSecret` + the existing coverage-gate test.

nadaverell:
Resolved (627085960). The finding describes `assignIncidentParents` attaching a parent in flat mode — but it's now gated on `len(grouped) > 0` (`diagnostic_context.go:246`), and the ungrouped paths (`?view=flat`, per-resource regroup) pass `grouped=nil`, so the assignment is **skipped entirely** there. `collectRelated`'s coverage line is moot in that mode because its edges are never consumed. Pinned by `TestIncidentParent_FlatViewNoPointer` (two flat pods sharing an ID → no incident_parent). The bot re-anchored the same BUG_ID statically without tracing the upstream gate.

PR #1082, opened 2026-07-02
feat: enhance namespace handling and resource access in Kubernetes

FinnGrndl:
## Problem

Identities that can list resources in specific namespaces but cannot list namespaces cluster-wide (a common locked-down RBAC setup) only ever got one namespace per resource kind: the RBAC probe stopped at the first grant, so Radar silently showed a fraction of what the identity could actually see.

## What this does

Adds `--namespaces ns1,ns2,ns3` (flag + persisted `namespaces` list in `~/.radar/config.json`) naming the namespaces to probe when cluster-wide listing is denied.

- **Probe collects every grant.** `probeResourceAccess` records all namespaces where a kind is listable (`PermissionCheckResult.ScopeNamespaces`), keeping the first grant as the primary for single-namespace consumers.
- **One informer per granted namespace.** For a kind granted in several namespaces, `pkg/k8score` starts a namespaced informer per grant; typed listers read through a read-only union indexer that fans out across the per-namespace informer stores. Reading the informers' own stores keeps `HasSynced` meaningful and holds each object once. Namespaces are disjoint across the informers, so the union cannot duplicate an object.
- **Launch URL preselects the list.** Browser and desktop launches carry `?namespaces=`, so local sessions open with the configured namespaces selected in the picker.
- **Visibility diagnostics understand the fanout.** A namespace served by the union lister reports as visible, not "unavailable".
- **`--namespace` vs `--namespaces`** resolve with explicit mutual exclusion; `--namespace-scope` still pins to exactly one namespace and rejects multiple.

## Guardrails

- The list is bounded by the probe fanout cap (`--max-scope-candidates`, default 20). Startup fails with a clear error when the list exceeds the cap — including the boundary case where a kubeconfig context namespace distinct from the list consumes a candidate slot — instead of silently probing a subset. Truncation on any remaining path logs which namespaces were dropped.
- Probe deadline expiry keeps every completed outcome (partial result on the short error TTL, full re-probe follows) instead of starting Radar with zero typed informers, and logs which kinds were cut short.

## Scope

Applies to built-in (typed) resource types. Two follow-ups land separately:
- CRD/dynamic-cache multi-namespace parity (CRDs currently use the first granted namespace).
- Server-side picker seeding for auth-mode sessions, rebuilt on top of the pick-mutation primitives from #1115.

## Testing

- Unit coverage: probe multi-grant collection, cached-result deep copies, union indexer read fanout + rejected writes, multi-namespace cache behavior end-to-end against a fake clientset, flag precedence, fanout-cap boundary arithmetic, visibility for primary/secondary/denied namespaces.
- `pkg/k8score` suite green under `-race`; full backend suite green.


nadaverell:
Thanks for the PR! Sorry it took a few days to get to review.

Flagging two things here from initial review pass:
1) I think we should finish with #1115 first, then rebase this on top. They touch same areas, will conflict but should be easy enough to reconcile. Will probably need to change the ns picker logic to seed the ns on startup only, and go with 1115 mechanism for handling deleted ns after initial load.
2) I believe this only adds support for the main typed kinds, not dynamic informers (CRDs). The user experience will be not great this way - if we add --namespaces I think it should support all resources.

nadaverell:
@FinnGrndl thanks a lot for this — the core here (the probe collecting every granted namespace and per-namespace informers behind a union lister) is exactly the right shape for the "can list resources in some namespaces, can't list namespaces" identity, and we want it in.

While this was open, #1115 landed and rewrote the two pick-seeding functions this PR touches (`loadSavedNamespacePreference` / `setActiveNamespaceForUser`), putting the branch into hard conflict. Rather than bounce that back to you, we pushed an update to your branch (maintainer edits — thanks for leaving that enabled). What changed:

- **Merged current `main`, dropping the server-side picker seeding for now.** It conflicts semantically with #1115's pick-mutation primitives (`Delete`-based eviction vs the empty-pick sentinel). We'll rebuild the pre-filled picker on top of those primitives in a follow-up; the `?namespaces=` launch URL already pre-selects for local sessions, so only auth-mode sessions wait for that follow-up.
- **Union lister rework:** the mirror indexer filled via event handlers *after* `HasSynced` flipped (a brief window where a synced kind could list incomplete results) and held every object twice. Listers now read a read-only union over the per-namespace informers' own stores — same behavior, no gap, half the memory.
- **Hardening from review:** probe-deadline expiry now keeps completed outcomes instead of returning empty scopes (and logs which kinds were cut short); `--namespaces` entries past the probe fanout cap now warn and fail fast at startup instead of silently dropping; visibility diagnostics consult the full per-kind namespace set so a granted secondary namespace no longer reports "unavailable".
- **Docs** note the current typed-resources scope — CRD multi-namespace parity is coming next as a separate PR.

Your commits are untouched underneath; the update lands as a merge plus follow-up commits on top (the repo squash-merges anyway). Since this PR rides your fork's `main`, run `git pull` there before branching from it again.

nadaverell (approved):

PR #1110, opened 2026-07-06
audit: feed pod metrics so resourceUtilization actually evaluates

nadaverell:
## Summary

The `resourceUtilization` audit check — the one that flags over-/under-provisioned pods (usage far from requests) — has never run in the fleet audit. The runner built `CheckInput` from the resource cache but never populated `PodMetrics`, so the check silently no-op'd: its efficiency findings never appeared and its coverage denominator was always absent (it just showed up as a missing input). Radar already collects the data — this wires it in.

## What changed

`RunFromCache` now builds `PodMetrics` by joining two sources it already has access to:
- **Usage** from the metrics-history store (`GetAllPodMetricsLatest()`), the same continuously-polled store that backs the Top/metrics views.
- **Requests** summed per pod from the container specs already listed for the audit.

`joinPodMetrics` normalizes both to the millicores/bytes the check compares (store CPU is nanocores), and only includes pods that **both** reported a usage sample and are in the audited scope — a pod with no sample isn't evaluated.

Points worth a reviewer's eye:
- **Native sidecars count.** The request denominator includes `initContainers` with `RestartPolicy=Always` — their usage is part of pod-level metrics, so omitting their requests would inflate the ratio and produce false warnings.
- **Honest when metrics are absent.** If metrics-server hasn't reported (store nil or empty), `PodMetrics` stays nil and the check reports as a missing input rather than a false empty pass — no behavior change on clusters without metrics.
- **Concurrency-safe read.** `GetMetricsHistory()` now reads the store pointer under the same lock `Init`/`Reset` write it with, since audit scans (HTTP/MCP) call it concurrently with a context-switch reset.

## Testing

`go build ./...`, `go test ./internal/audit/ ./pkg/audit/` green. New unit tests cover the usage↔request join (unit conversion, in-scope filtering, nil-when-empty) and the native-sidecar denominator (native sidecar counted, plain init container excluded).

Live-verified across three environments:
- **GKE** (218 pods, metrics-server): `resourceUtilization` evaluates 71 pods (10 passing) → 61 findings; spot-checked a flagged pod's reported usage/requests against `kubectl top` + the pod spec to confirm the unit conversion is correct on real data. The fleet has 9 native-sidecar pods, exercising the init-container denominator path.
- **EKS** (63 pods, ARN context): evaluates 15 pods (4 passing) → 11 findings — confirms the path on a non-GKE provider.
- **kind, no metrics-server**: graceful degradation — `resourceUtilization` stays in `missingInputs`, produces zero findings, and does not panic (identical to prior behavior).

In both metrics-present clusters `podmetrics` is gone from `missingInputs` — the check runs instead of being reported as unavailable.

### DIFF
--- a/internal/audit/runner.go
+++ b/internal/audit/runner.go
@@ -15,8 +15,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/skyhook-io/radar/internal/k8s"
-	"github.com/skyhook-io/radar/pkg/k8score"
 	bp "github.com/skyhook-io/radar/pkg/audit"
+	"github.com/skyhook-io/radar/pkg/k8score"
 )
 
 // RunOptions provides optional data sources for checks that need them.
@@ -31,6 +31,11 @@ func RunFromCache(cache *k8s.ResourceCache, namespaces []string, opts *RunOption
 	if cache == nil {
 		return &bp.ScanResults{Summary: bp.ScanSummary{Categories: map[string]bp.CategorySummary{}}}
 	}
+
+	// Capture the context generation before reading the cache. If a context
+	// switch lands while we're building the input, the metrics join below bails
+	// rather than pairing this cluster's pods with the next cluster's usage.
+	gen := k8s.CurrentOperationGen()
 
 	input := &bp.CheckInput{
 		Pods:                     listNamespaced(cache.Pods(), namespaces),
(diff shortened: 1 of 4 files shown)


nadaverell:
Closing — not merging. `resourceUtilization` in the audit duplicates two surfaces that already tell this story better: **CostView** (per-cluster reserved-vs-used waste, framed in $/hr idle, with sensible tiers) and **RightsizingStrip** (per-workload Prometheus P95, risk measured vs the *limit*, current→recommended). This check is the crude cousin — instantaneous single sample, ratio-to-*request* (wrong risk signal), no absolute-value floor, flat warning severity — so wiring it up produces ~60 undifferentiated per-pod warnings for signal already covered. Fleet-level utilization visibility (a real gap) belongs as a Cost/rightsizing rollup, not per-pod audit findings. Decision from /product-review 2026-07-07.

PR #1130, opened 2026-07-08
timeline: 6a redesign — log-explorer strip, unified drawer, design turns 7–11

roylibman:
## What

Implements the [Timeline Redesign design doc](https://claude.ai/design/p/92ff992e-c389-4661-8d0d-10acd611aa35?file=Timeline+Redesign.dc.html) (final composition 6a through Turn 11) on top of #1111. **Stacked on `radar-timeline-standalone`** — that PR merges first; this diff is only the redesign.

### Strip — new `TimelineStrip` (replaces the two-scale minimap scrubber in local + retained modes)
- **Query vs Window, never confused**: the range picker is the *only* control that changes the query; the blue **window** band is the slice shown in the lanes — always ⊆ the query. Drag to pan, edges to resize, draw on the histogram to place it, −/＋ snaps to round rungs capped at the query span.
- **Query pill speaks the right tense**: live preset queries read relatively (**"Last 1h" / "Last 7d" / "All history"**, exact range on hover) so the label doesn't re-render on every tick; frozen/custom ranges keep explicit timestamps.
- **Custom range = theme-styled datetime pickers**: stamp buttons opening a compact month-grid popover (today/selected states, time field, "Now" shortcut) — replaces both the OS-native widget and the freetext inputs tried earlier. Structural validation (Apply disables when To ≤ From).
- **Honest histogram**: bars positioned by *time* (index positioning had scattered sparse buckets — a bar could sit "under the window" holding another hour's events); in-window bars bright, out-of-window muted; amber warning caps; persistent band range label; corner timestamps; exact "N events in query range" (matches the toolbar's query-scoped chip counts).
- **Pre-data region**: dims the query span from before recording started ("no data before HH:MM") — explains why the list can't scroll further back.

### Swimlanes
- **Two-line lane rows**: the name owns the full first line (360px label column) with the namespace as a quiet subtitle beneath it — so the name stops losing the width fight against metadata and no longer middle-truncates in the common case. Children inherit their group's namespace (no per-row repeat); root-anchored tree rails.
- **Health bar on the lane midline**: the health span runs through the row's center and event dots ride on it — one axis per resource, so the bar unambiguously belongs to its row and the top/bottom of each row is clean gutter (was a separate 7px strip pinned to the row's bottom edge, which straddled the divider to the next row).
- **Continuous zoom**: ctrl/cmd + wheel scales the window smoothly by the wheel delta (device-normalized, floored at 1 min); the +/- buttons keep the round preset stops.
- **Kind chips: neutral + abbreviated** — long kinds get k8s-idiomatic short forms (`Application→APP`, `CustomResourceDefinition→CRD`, `PodDisruptionBudget→PDB`) with a generic CamelCase-initials fallback (`VerticalPodAutoscalerCheckpoint→VPAC`; unmapped CRDs self-shorten, e.g. `ClusterSecretStore→CSS`); full kind always on hover.
- **Color budget (Turn 11)** — color is reserved for status + selection; identity is monochrome. One activity hue (info-blue; shape carries the event class), status is exactly three colors (green/amber/red; rolling/idle map to gray "no signal"), a critical-dominant cluster shows one red **⊘ ×N**, warnings are amber everywhere (red only for actually-broken).
- Aggregate health sweeps on group headers (green/amber weave on member disagreement); every row keeps a baseline; markers clamp at the Now line.
- **Now line**: fixed a 40px geometry drift (stale label-column constant) and a blink on every live data load (marker now pins to the live edge instead of vanishing when `now` transiently passes the latched window).
- **Expand/collapse all**: single morphing toggle in the RESOURCE header (10a — offers collapse when ≥half the groups are open, expand otherwise; `E` shortcut; tooltip names both).
- Toolbar: discrete filter pills with query-scoped counts, labeled "In view" counts, dashed Kinds chip, on-demand legend popover.

### Reusable lane rendering — `LaneTrack`
A single lane's timeline — backdrop grid, centered health bar, and event-dot markers — is factored into one exported **`<LaneTrack>`** primitive. Both the full timeline page and the resource/application detail timeline (below) compose it, so the health-bar + dot rendering lives in exactly one place and can't drift between the two surfaces.

### Workload / application detail timeline
The resource/application detail view's **Timeline** tab is built from the shared timeline pieces and shaped around what an operator opens it to learn:
- **Swimlane only when it earns its space** — when there's more than one resource to lay out over time (a workload + its tree). A single resource (a lone Pod, a ConfigMap) shows just the **event feed**; the swimlane, legend, and zoom are omitted.
- **Lanes surface what matters:** the live **Pods** (where restarts/crashes show) and the **active ReplicaSet** (the one still owning Pods) rank highest; retired ReplicaSets sink to the bottom, and resources with **no events in view** aren't drawn at all — no flat empty graveyard lanes. Up to 6 lanes plus a pinned axis are visible before scrolling.
- **One rendering, shared with the main page:** each lane is a **`<LaneTrack>`** and the axis is the shared **`TimeAxis`**, with the same continuous ctrl/cmd + wheel zoom — so the health-bar/dot/axis rendering can't diverge between the two surfaces. The orphaned old primitives (`HealthSpan`, the duplicate `EventMarker`) are deleted.
- **Feed:** an events table beneath the swimlane; change types are title-cased and the internal `neutral` status is suppressed, while meaningful k8s reasons (`Progressing`, `SuccessfulCreate`) stay.

### Event drawer — one anatomy (Turn 9)
Rail + detail for single events *and* clusters: a single event is a rail of one plus its ±15-min correlated neighbors; the header echoes the selected dot with ‹ › stepping; property-grid detail with diff and a View-resource action; opens at 40 vh with a draggable top edge, no backdrop (timeline stays interactive).

### List view
Query-bounded fetch (limit 2000), two-way strip sync (scrolling moves the band; dragging the band scrolls the list), view switches preserve the window. Default query = last 1 h. Kind-less events (e.g. GKE `resource-tracker`) get a neutral `EVENT` chip.

### Known deviations from the design (deliberate)
- The all-history **minimap** is removed per the design's Turn 3 decision — the range picker owns cross-history navigation.
- Drawer action row ships **View resource** only; *Logs at this time* / *YAML* need `NavigateToResource` to grow a tab/target param (public `@skyhook-io/radar-app` surface — separate change).
- The **Now line/label stays purple** — a sixth hue Turn 11 doesn't account for; needs a design call (brand vs. gray).
- `DateTimeField` lives in `TimelineStrip.tsx` for now — graduate to `components/ui/` when another view needs date input.
- The old `TimelineScrubber` (and its 86 tests) is kept but unconsumed, for reference during review — delete in a follow-up.

## Testing
- `tsc --noEmit` clean; **1,181 k8s-ui vitest** green (adds `LaneTrack` continuous-zoom / wheel-factor pins alongside the existing `TimelineStrip` window⊆query, time-positioned-bucket, query-pill, drawer-anatomy, chip-abbreviation, and morphing-toggle suites).
- Exercised end-to-end against k3d + GKE (nonprod/prod): picker + calendar, window drag/draw/resize, live mode across data loads, list two-way sync, drawer stepping/resize, pre-data region, bulk expand/collapse, color budget.
- New surfaces verified in the running UI with DOM hit-tests, not just screenshots: two-line rows + midline health bar on the main timeline (dots co-centered with the bar to the pixel); the detail-view Timeline tab rendering the shared `LaneTrack`, its pinned axis aligned to the lane tracks with a scrollbar present, and continuous wheel zoom landing on off-preset widths (e.g. 1h → 47m); a single-resource view (ConfigMap) collapsing to feed-only; and a Deployment surfacing its live Pods + active ReplicaSet instead of retired ones.


nadaverell:
image
unnecessarily big. even if many events are clustered it doesn't look good:
image

still mostly empty whitespace. the height should be what makes sense for a single event (comfortably, not squeezed, but one event)

nadaverell:
image
duplicate clickable targets to open the resource.
use the standard radar resource link badge like in other places:
image

nadaverell:
image
black is jarring here, as is white in darkmode

a bit too much whitespace above:
image

nadaverell:
image
delete icon without text is confusing, even if tooltip explains. this needs text next to it

nadaverell:
scrollwheel zoom - revert back to anchoring on right, not middle. so that a user can "zoom to view farther back or focus  on more recent". anchoring on middle is very confusing

nadaverell:
Not a fan of hiding the legend altogether behind a button. narrow strip like the previous live design is better IMO

nadaverell:
timeline mode -> swtich to list mode -> mouse back -> infinite loop stuck

nadaverell:
image
refresh button not relevant here. adopt the live indicator from other tabs like topology:
image

nadaverell:
image

1) don't use browser-native "title" showing these tooltips, always our custom Tooltip component
2) avoid entirely tooltips that don't actually show any more info
3) I don't think it makes sense to mix these two controls. "View" doesn't hint at them. Should just be two controls.

Tooltip component here as well:
image

Do a review pass to catch all these instances in the PR(s) - there are many more

roylibman:
> image unnecessarily big. even if many events are clustered it doesn't look good: image
> still mostly empty whitespace. the height should be what makes sense for a single event (comfortably, not squeezed, but one event)

noted, I'll reduce it

nadaverell:
image
not super clear what these buttons do - what is being filtered exactly?

roylibman:
> image black is jarring here, as is white in darkmode
> a bit too much whitespace above: image

I'll try other colors, but I'm not sure we'll have solution for the space that we need to save for it, it will have to show above or below.

roylibman:
> image delete icon without text is confusing, even if tooltip explains. this needs text next to it

I prefer to reduce the amount of text, I'll try to think of a creative alternative

nadaverell:
> > image black is jarring here, as is white in darkmode
> > a bit too much whitespace above: image
> 
> I'll try other colors, but I'm not sure we'll have solution for the space that we need to save for it, it will have to show above or below.

not sure it's needed at all... could also appear below, when not stretching all the way to the sides.

the "530 events in query range" could move somewhere else if needed.

roylibman:
> Not a fan of hiding the legend altogether behind a button. narrow strip like the previous live design is better IMO

I'm highly against it. The page is very noisy and overwhelming as it is. There are too many different icons to keep up with what each means. I don't see people looking back and forth from the legend to the chart all the time. Most users will likely just hover.

roylibman:
> > > image black is jarring here, as is white in darkmode
> > > a bit too much whitespace above: image
> > 
> > 
> > I'll try other colors, but I'm not sure we'll have solution for the space that we need to save for it, it will have to show above or below.
> 
> not sure it's needed at all... could also appear below, when not stretching all the way to the sides.
> 
> the "530 events in query range" could move somewhere else if needed.



> image not super clear what these buttons do - what is being filtered exactly?

they filter events from the list view, and from the timeline view it removes resources that didn't have events during this time. how would you otherwise suggest to apply these filters of locating unhealthy resources?

nadaverell:
> > > > image black is jarring here, as is white in darkmode
> > > > a bit too much whitespace above: image
> > > 
> > > 
> > > I'll try other colors, but I'm not sure we'll have solution for the space that we need to save for it, it will have to show above or below.
> > 
> > 
> > not sure it's needed at all... could also appear below, when not stretching all the way to the sides.
> > the "530 events in query range" could move somewhere else if needed.
> 
> > image not super clear what these buttons do - what is being filtered exactly?
> 
> they filter events from the list view, and from the timeline view it removes resources that didn't have events during this time. how would you otherwise suggest to apply these filters of locating unhealthy resources?

As a use who knows k8s I honestly don't know what to expect will happen when I click - what is the relationship between "k8s events" and the other filters? is "unhealthy" the object or the event?  currently unhealthy objects? if event - what is an "unhealthy" event - not well specified. if object - right now unhealthy? was at some point? but then this is different from "warnings" because an object can't be in "warnings" state...

This is not intuitive at all...

Tell me what you're trying to accomplish with these buttons and I could try and suggest an alternative.

roylibman:
> 3\. I don't think it makes sense to mix these two controls. "View" doesn't hint at them. Should just be two controls.

I think it's best to keep it as so since it takes too much real-estate otherwise (which isn't worth it in my view)

roylibman:
@nadaverell looking good! lets show it some sunlight

PR #1189, opened 2026-07-15
Per-kind progressive readiness during initial sync

### LINKED ISSUE #1149: Background sync
## Problem statement

I am using the quick install setup. The initial sync may take a long time and there is no way to "skip" it.

## Proposed solution

I would like the option to move to a specific tab the relevant sync for the tab will take precedence.
For example, during the initial sync I move to resources/pods I would this view to be loaded first and not wait for the full sync to start working.

nadaverell:
Fixes #1149.

## What

On large clusters the initial informer sync takes minutes, and until now the entire UI sat behind a splash screen with every read returning 503 until the critical set — including pods, typically the slowest LIST — finished. Radar now becomes usable **per kind, as each informer completes**: navigate to Deployments while Pods is still listing and the full Deployments view works; the Pods view shows a loading state and appears the moment its data is complete.

**Honest bound:** the pods view itself can't appear faster than the pods LIST — the win is every other view, plus a legible app with visible per-kind progress instead of an opaque splash.

## How it works

**Readiness, not reordering.** Informers already start in parallel; there is no queue to jump. The change is exposing per-kind sync state and gating reads on it:

- `k8score`: `OnInformersStarted` hands out the cache mid-Phase-1; `KindReadinessFor` classifies a kind `ready / pending / failed / unavailable` from live informer state; `GetSyncSnapshot` reports per-kind progress cheaply (no lister walks — the connection endpoint polls sub-second during sync).
- `internal/k8s`: a **generation-guarded** `syncingCache` handle alongside the untouched `resourceCache` singleton (its non-nil ⇒ synced invariant holds everywhere else). A context switch mid-construction invalidates every publish/promotion from the stale build — this also closes a pre-existing race where a construction finishing after a switch could republish the old cluster's cache.
- `server`: the resource list/get handlers gate per kind — `503 {error_code: kind_sync_pending}` while an informer syncs, `kind_sync_failed` when it never will — and serve from whichever cache exists. **This also fixes a shipped false-empty bug:** post-connect, promoted criticals and `isEnabled`-only listers (ReplicaSets, HPAs, ServiceAccounts…) could render a partial store as a complete list. Dynamic/CRD reads keep the connected gate (no dynamic cache exists mid-sync). Single-GET serves without relationship enrichment while the topology cache doesn't exist. A disconnected cluster still 503s everywhere — cache handles stand in for connectedness only during `connecting`.
- `web`: the splash is replaced by the app shell as soon as informers start. Resource views work per kind — a pending kind stays in its loading state (**an unsynced kind never renders as an empty list**) and populates automatically. Views needing the full dataset (Home, Topology, Timeline, …) show a per-kind progress panel with links to ready kinds. The header carries "Loading cluster data — x of y ready". The large-list guard holds while counts are unavailable, and warm shared query caches (embedded multi-cluster mounts) are dropped before the shell can show another cluster's data.

`RADAR_DEBUG_SYNC_DELAY=pods=120s` (dev seam) simulates a slow kind for testing/demoing the window.

## Testing

- New deterministic tests: k8score progressive readiness via blocked-LIST fake-client reactors (pending → ready, deferred-timeout → failed), generation-guard publish/promote/clear vs reset, alias→informer-key mapping, debug-delay parsing. Full `go test ./...` + k8score module + `tsc` + frontend build green.
- Visual-tested against a live GKE cluster with the delay seam: progress panel with per-kind links → full Deployments list (92 items) during sync → Pods loading state → automatic transition to the complete dashboard on connect. Fast-cluster path regression-checked (no delay: straight to connected, no syncStatus in payload).

## Deliberate scope cuts

- Aggregate views stay gated until connected — partial topology/audit would mislead.
- `/resource-counts` isn't progressive yet, so guarded kinds (Pod/Event/ReplicaSet/EndpointSlice) hold their large-list guard until connected — the safe direction; progressive counts is a natural follow-up.
- Events sync in the background with no deadline by design; its view polls while open rather than erroring (previously it error'd out after 3 retries mid-sync).

---

> [!NOTE]
> **Medium Risk**
> Changes cluster connection gating, cache singleton lifecycle on context switches, and what resource APIs return during sync—high user impact but guarded by generation checks and explicit readiness errors instead of silent wrong data.
> 
> **Overview**
> Large clusters used to block the whole UI on a connecting splash until critical informers (often pods) finished listing. This PR exposes **per-kind sync state** so resource list/get handlers and the web shell can open **one kind at a time** while others are still pending.
> 
> **Backend:** `k8score` adds `OnInformersStarted`, `KindReadinessFor` (ready/pending/failed/unavailable), `GetSyncSnapshot`, and optional `DebugSyncDelays`. `internal/k8s` publishes a generation-guarded mid-sync `syncingCache` before Phase-1 completes; `ResetResourceCache` bumps `cacheGeneration` so stale constructions cannot republish an old cluster’s cache after a context switch. Resource handlers use `gateResourceRead` / `requireConnectedOrSyncing` and return `503` with `kind_sync_pending` or `kind_sync_failed` instead of serving partial stores as empty lists. Connection status includes `syncStatus` while `connecting`.
> 
> **Frontend:** When `syncStatus` is present, the app shell replaces the splash; resource views load per kind (pending kinds keep polling); Home/Topology/etc. show `SyncProgressPanel` until connected. React Query retries on `kind_sync_pending`; shared query caches are cleared when progressive sync starts to avoid cross-cluster stale data.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eedcbcce8886ef0a4320daf968b96e3d17c6d550. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/k8s/cache.go
+++ b/internal/k8s/cache.go
@@ -457,7 +457,6 @@ func InitResourceCache(ctx context.Context) error {
 			return
 		}
 
-		initialSyncComplete = core.IsSyncComplete()
 		// Only the no-enabled-resources path (OnInformersStarted never ran)
 		// reaches here with a nil embedded pointer; when the callback did run
 		// it already set the same core, and rewriting it would race with
@@ -471,6 +470,7 @@ func InitResourceCache(ctx context.Context) error {
 			initErr = fmt.Errorf("cluster changed during cache initialization")
 			return
 		}
+		initialSyncComplete = core.IsSyncComplete()
 	})
 	return initErr
 }
--- a/internal/k8s/cache.go
+++ b/internal/k8s/cache.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
 	"strings"
 	"sync"
 	"time"
@@ -304,6 +305,17 @@ var (
 	resourceCache *ResourceCache
 	cacheOnce     = new(sync.Once)
 	cacheMu       sync.Mutex
+	// syncingCache is the mid-Phase-1 handle published by OnInformersStarted:
+	// structurally complete but not synced — every read through it MUST gate on
+	// KindReadinessFor. Cleared on promotion (resourceCache assigned), on init
+	// failure, and on reset. Guarded by cacheMu.
+	syncingCache *ResourceCache
+	// cacheGeneration invalidates in-flight cache constructions: a context
+	// switch bumps it (via ResetResourceCache), and any publish/promote/clear
+	// from an older generation becomes a no-op — without this, an old
+	// cluster's construction finishing late could republish that cluster's
+	// cache over the new one. Guarded by cacheMu.
+	cacheGeneration uint64
 )
 
 // tombstones retains recently-seen resource enrichment (owner/labels/createdAt)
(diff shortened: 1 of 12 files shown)


hisco:
@nadaverell I did a rebase of this branch, please consider if this branch should be merged

PR #1201, opened 2026-07-17
Extend per-issue change correlation to warning-severity issues

nadaverell:
## Why

In degraded-not-down incidents the active fault often manifests at **warning** severity — e.g. a Service selector matching no pods right after a selector change. Per-issue change correlation previously skipped everything below critical (`attachIssueChangeCorrelation`), so the actual fault arrived with no `correlated_changes` evidence while chronic criticals carried explicit markers. In SREGym benchmark transcripts this evidence asymmetry repeatedly steered agents toward the loudest chronic critical (or an unrelated entry in the global `recent_changes` feed) instead of the changed warning-severity subject.

## What

- Correlation runs **two passes sharing the existing cap (10)**: all criticals first, then warnings. Cap priority is independent of response sort order — a warning can never cost a critical its lookup slot.
- **Group-aware correlation, both directions**: the eligibility gate (`TrackedKindForGroup`) treats a CRD issue whose kind collides with a tracked kind (Knative Service vs core Service) as untracked, and candidate queries drop timeline events recorded from a different API group than the one the feed tracks for that kind — so neither side of a same-name kind collision can absorb the other's changes. A drift-guard test pins the kind→group map to exactly the feed's tracked set in both directions (Secret stays out: its updates are delete-only in the feed, so marker-eligibility would produce false claims).
- **Truthful saturation under crowding**: saturation keys on the raw pre-filter candidate count, so mismatched-group events filling the bounded query window read as "unknown", never as a false `no_recent_changes`.
- Same subject-scoped, deterministic, non-causal machinery as before: no demotion, no reordering, no new heuristics; the existing truthful-window semantics apply unchanged. Contract docs updated throughout (`CorrelationTruncated`, `no_recent_changes` covers tracked changes on the subject only, `issues` tool description).

## Tests

Warning correlation + unchanged-warning markers; criticals-priority-under-cap; warning-only truncation; both group-collision directions (Knative issue vs core events, core issue vs CRD events); CRD-crowding saturation; Secret non-eligibility; worst-case payload bound (~43KB measured, 128KB guard). Full `make test` green.

---

> [!NOTE]
> **Medium Risk**
> Changes triage evidence agents rely on (correlation markers and saturation semantics) and adds group filtering on timeline queries; behavior is heavily tested but wrong group mapping could silently disable correlation for a kind.
> 
> **Overview**
> **Per-issue change correlation** on single-namespace `issues` responses now covers **warning** as well as **critical** issues, using the same 10-issue cap with **criticals always processed first** so response ordering cannot starve critical lookups. Logic is refactored into `correlateIssue`; MCP/API docs clarify that `no_recent_changes` is scoped to **tracked** subject changes, not proof of a chronic incident.
> 
> **Group-aware correctness** replaces plain `TrackedKind`: `TrackedKindForGroup` skips CRD issues whose kind collides with a tracked core kind (e.g. Knative vs core `Service`), and timeline candidate queries **filter events by API group** so same-named resources cannot share history. **Saturation** now uses the **raw pre-filter** candidate count so mismatched-group churn cannot yield a false `no_recent_changes`. A `trackedKindGroups` map (drift-tested) keeps marker eligibility aligned with what the feed actually records—**Secret** issues stay ineligible because updates are not tracked.
> 
> Tests cover warning markers, cap priority, group collisions both ways, CRD crowding → unknown, Secret non-eligibility, and a worst-case JSON payload bound (~128KB).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b88aaa45277d31fe1cbceb858c3b413b04f93aae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/mcp/issue_correlation.go
+++ b/internal/mcp/issue_correlation.go
@@ -129,41 +129,10 @@ func correlateIssue(ctx context.Context, iss *issuesapi.Issue, window time.Durat
 		}
 		return
 	}
-	changes = annotateOnsetOrder(changes, iss.FirstSeen)
 	if len(changes) > correlationChangeCap {
 		changes = changes[:correlationChangeCap]
 	}
 	iss.CorrelatedChanges = changes
-}
-
-// onsetSlop absorbs detection lag and clock skew between a change landing
-// and the resulting condition transition: a change within this margin after
-// first_seen may still explain the onset, so it is not flagged.
-const onsetSlop = 2 * time.Minute
-
-// annotateOnsetOrder marks changes that postdate the issue's onset
-// (first_seen + slop) with AfterIssueOnset — deterministic causal-direction
-// evidence: such a change cannot explain why the issue began, though it may
-// explain later behavior. Pre-onset changes are ordered first so the cap
-// never drops a possible cause in favor of a post-onset edit. No filtering:
-// the consumer weighs.
-func annotateOnsetOrder(changes []issuesapi.RecentChange, firstSeen time.Time) []issuesapi.RecentChange {
-	if firstSeen.IsZero() {
-		return changes
-	}
-	cutoff := firstSeen.Add(onsetSlop)
-	pre := make([]issuesapi.RecentChange, 0, len(changes))
-	post := make([]issuesapi.RecentChange, 0)
-	for _, c := range changes {
-		ts, err := time.Parse(time.RFC3339, c.Timestamp)
-		if err != nil || !ts.After(cutoff) {
-			pre = append(pre, c) // unparseable ⇒ unknown ⇒ no claim
-			continue
-		}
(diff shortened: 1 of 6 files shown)


nadaverell:
Both Bugbot findings addressed: (1) Secret removed from the marker-eligibility map — Secret updates are delete-only in the feed, so Secret-kind issues now carry no markers (regression test added; the drift guard now pins the map to exactly TrackedKind's set in both directions). (2) The saturation-ignores-filtered-events finding was reviewed against commit 5653422f and had already been fixed in d0419ad7 (saturation keys on raw pre-filter candidate counts, with a crowding regression test).

PR #1202, opened 2026-07-17
Dashboard warningGroups: full dedup window with lastSeen + involved objects, recency-ordered

nadaverell:
## Why

`get_dashboard`'s `topWarnings` rows carried only `{reason, message, count}`, sorted by lifetime count. Two observed failure modes (SREGym benchmark transcripts):

1. A noisy-but-resolved warning group dominated the list over the live incident (count-first ordering).
2. A consumer had **no way to tell a stale warning from a current one** — a long-recovered BackOff read as live crash behavior, costing an otherwise-correct diagnosis.

## What

- Each row now carries `lastSeen` plus the distinct involved objects as structured refs `{kind, group, namespace, name}` (capped at 3, most recent first, with `objectCount`/`objectsTruncated` for the uncapped distinct total, counted before the cap) — the ref shape feeds directly into `get_resource`'s inputs, with API group included because bare kinds collide (core vs Knative Service).
- **No row selection — the full dedup window is emitted** (up to 20 groups, recency-ordered), and the field is renamed `topWarnings` → `warningGroups` since nothing is "top" anymore. An earlier revision selected 5 rows (3 newest + 2 highest-count union); design consult concluded any pick-N-of-20 heuristic creates unrecoverable omissions — an agent can underweight a returned row but cannot recover an absent one, and both observed failures were selection failures. With `lastSeen` on every row, stale rows are self-labeling and skimmable rather than misleading. Worst-case cost is ~20 rows only on clusters that actually have 20 distinct warning groups.
- **Deterministic ordering end to end**: the full comparator (recency, count, reason, message, type — type is a grouping key so it must be a comparator key) applies inside dedup before its 20-group cap, so equal-timestamp survival no longer depends on informer iteration order; representative message on exact ties is deterministic.
- **Object identity keys on `InvolvedObject.UID`** when present (falling back to kind/group/namespace/name), with a known apiVersion carried across merged sightings — emitters populate apiVersion inconsistently, and raw-string keying double-counted one object seen through two emitters. Emitted refs then collapse successive incarnations (same kind/group/namespace/name across different UIDs, e.g. a recreated StatefulSet pod) into one subject, so objectCount never counts two UIDs a consumer cannot tell apart.
- **Event recency/count honor events.k8s.io `Series` fields** (`Series.LastObservedTime`/`Series.Count`) with legacy fallback — previously an actively repeating series warning read as a stale count-1 one-off. Benefits all `DeduplicateEvents` consumers.
- Object collection is a **dashboard-scoped opt-in** (`DeduplicateEventsWithObjects`); the shared `DeduplicateEvents` wire shape is byte-identical for its other consumers (get_events, diagnose, resource includes), pinned by a shape test.
- `get_dashboard` tool description separates the three signals: `lastSeen` for liveness, `count` for cumulative occurrence volume, `objectCount` for breadth (`objects` is a capped sample of 3).

## Tests

Dashboard-level recency+objects assertions; full-window emission (7 groups → 7 rows, recency-ordered, oldest high-count storm present); distinct-object collection across pod-hash-normalized groups; UID merge across missing apiVersion; ref-validity guard (kind+name required); Series honor; cap-boundary determinism under shuffled input incl. type-only ties; group derivation; unchanged legacy wire shape. Full `make test` green.

Known pre-existing issue deliberately not expanded here (tracked as follow-up): the shared dedup helper's internal 20-group cap conflicts with `get_events`' documented limit=100 and `cluster://events`' cap=50.

---

> [!NOTE]
> **Medium Risk**
> Breaking MCP JSON field rename (`topWarnings` → `warningGroups`) and shared event dedup behavior changes (Series timestamps, sort order) affect every `DeduplicateEvents` consumer, though the legacy wire shape is explicitly preserved.
> 
> **Overview**
> **`get_dashboard`** replaces **`topWarnings`** with **`warningGroups`**: each row adds **`lastSeen`**, involved-resource refs (**`objects`**, up to 3, most recent first), **`objectCount`**, and **`objectsTruncated`**, with **`group`** derived from **`apiVersion`** for **`get_resource`**. Groups are ordered by **recency** (not lifetime **`count`**), and the dashboard emits the **full dedup window** (up to 20 groups) instead of a top-5 count-first slice.
> 
> Event dedup in **`pkg/ai/context`** is refactored behind **`deduplicateEventGroups`**. **`DeduplicateEventsWithObjects`** is dashboard-only; plain **`DeduplicateEvents`** keeps the same JSON shape. All dedup consumers get **`Series.LastObservedTime`** / **`Series.Count`**, deterministic tie-break sorting before the 20-group cap, and richer involved-object identity (UID, API group, incarnation collapse).
> 
> Tests cover dashboard ordering, full-window emission, and dedup/object edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba2552711e4618ea94fe122945d95c26f50187c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/mcp/dashboard_warnings_test.go
+++ b/internal/mcp/dashboard_warnings_test.go
@@ -2,11 +2,13 @@ package mcp
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/fake"
 
 	"github.com/skyhook-io/radar/internal/k8s"
@@ -25,11 +27,12 @@ func warningEvent(name, reason, message string, count int32, last time.Time, obj
 	}
 }
 
-// topWarnings must order by recency (not lifetime count) and carry the facts
-// a consumer needs to judge each row: lastSeen and the involved objects. The
-// old count-first ordering promoted a long-stale high-count group over the
-// live incident, and the old shape gave no way to tell the two apart.
-func TestBuildDashboard_TopWarnings_RecencyAndObjects(t *testing.T) {
+// warningGroups must order by recency (not lifetime count) and carry the
+// facts a consumer needs to judge each row: lastSeen and the involved
+// objects. The old count-first ordering promoted a long-stale high-count
+// group over the live incident, and the old shape gave no way to tell the
+// two apart.
+func TestBuildDashboard_WarningGroups_RecencyAndObjects(t *testing.T) {
 	defer k8s.ResetTestState()
 
 	ns := "shop"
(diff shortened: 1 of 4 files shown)


nadaverell:
Addressed second-round review findings: Type added as final comparator key (with cap-boundary determinism test), objects[] switched to structured refs with API group, involved-object guard now requires both kind and name. One reviewer finding deliberately deferred: the shared dedup helper's internal 20-group cap conflicts with get_events' documented limit=100 and cluster://events' cap=50 — that defect predates this PR and changing it alters an unrelated tool's behavior, so it's tracked as a follow-up instead of expanding this PR's scope.

nadaverell:
Fixed the object-identity finding: the distinct-object map now keys on {kind, apiGroup, namespace, name} (raw apiVersion excluded — emitters populate it inconsistently and version-within-group never distinguishes objects), keeping the most recent sighting's full ref for emission. Regression test added for the "" vs "v1" variant case.

PR #1217, opened 2026-07-19
feat(traffic): add Grafana Beyla as a traffic source (#1205)

### LINKED ISSUE #1205: feat(traffic): add Grafana Beyla as a traffic source
# Spec: Grafana Beyla Traffic Source Integration

## Problem Statement

Radar's traffic visualization currently supports three sources: Hubble (Cilium), Istio (service mesh), and Caretta (TCP-only eBPF via Prometheus). Users without Cilium or Istio get only TCP-level topology from Caretta — no HTTP methods, status codes, latency, or gRPC visibility. There is no way to observe L7 traffic in clusters that don't run a service mesh or Cilium CNI.

## Solution

Add Grafana Beyla as a fourth traffic source, enabling L7 HTTP/gRPC visibility via eBPF auto-instrumentation without requiring a service mesh. Beyla is deployed as an Alloy + Beyla DaemonSet (recommended) or standalone Beyla. Radar detects Beyla's Prometheus metrics and renders L7-enriched traffic topology alongside existing sources.

## User Stories

1. As a Radar user without Cilium or Istio, I want to see HTTP methods and status codes on traffic edges, so that I can understand application-level communication patterns
2. As a Radar user without Cilium or Istio, I want to see gRPC service calls in the traffic map, so that I can diagnose microservice dependencies
3. As a Radar user with Beyla deployed, I want Radar to auto-detect Beyla and activate it as a traffic source, so that I don't need manual configuration
4. As a Radar user with Alloy + Beyla deployed, I want Radar to detect Beyla running inside Alloy, so that the recommended deployment works out of the box
5. As a Radar user with standalone Beyla deployed, I want Radar to detect standalone Beyla pods, so that both deployment models are supported
6. As a Radar user, I want the traffic source priority to remain stable (Hubble > Istio > Caretta > Beyla), so that existing sources aren't displaced by the new integration
7. As a Radar user with Beyla deployed, I want to see HTTP request rates and error rates on traffic edges, so that I can identify problematic service pairs
8. As a Radar user, I want Beyla traffic data to merge L4 (TCP) and L7 (HTTP) information into unified flow objects, so that the topology view shows both connection-level and request-level data
9. As a Radar user, I want namespace filtering to work with Beyla metrics, so that I can focus on a specific namespace's traffic
10. As a Radar user without the recommended Beyla `attributes.select` configuration, I want Radar to fall back to owner-level names (Deployment, DaemonSet), so that basic traffic visibility works without extra configuration
11. As a Radar user who has configured Beyla's extended attributes (`attributes.select`), I want Radar to show pod-level names, ports, and transport protocols, so that I get more granular visibility
12. As a Radar user, I want the Beyla detection to check for the `beyla_network_flow_bytes_total` metric in Prometheus, so that Radar confirms metrics are actually flowing before activating
13. As a Radar user, I want Beyla detection to check for Alloy or Beyla pods as a fallback when Prometheus is unreachable, so that detection still works during Prometheus maintenance
14. As a Radar user, I want Beyla version information to be extracted from the `beyla_build_info` metric or pod image tags, so that I can verify which Beyla version is deployed
15. As a Radar user, I want Beyla traffic to stream at the same 10-second polling interval as Caretta and Istio, so that all sources have consistent update behavior
16. As a Radar user, I want Beyla detection to return `Native: false`, so that the UI correctly shows Beyla as a user-deployed tool (not a built-in cluster feature)
17. As a Radar user, I want the `job` label regex in Beyla PromQL queries to match both standalone Beyla and Alloy deployments (`job=~".*beyla.*|.*alloy.*"`), so that both deployment models produce metrics
18. As a Radar user, I want Beyla's L7 metrics (`http_request_duration_milliseconds_count`) to be merged with L4 flow data, so that HTTP request information enriches the existing TCP flow topology
19. As a Radar user, I want Beyla flows to be keyed by `(srcNs, srcName, dstNs, dstName, dstPort)` for L4 and `(srcNs, srcName, dstNs, dstName, httpMethod, httpRoute)` for L7, so that the merge logic correctly matches L7 data to the right flow
20. As a Radar user, I want the recommendation system to suggest Alloy + Beyla for clusters without Cilium or Istio, so that I know what to install for L7 visibility
21. As a Radar user, I want Beyla detection to handle the case where the Prometheus metric exists but has zero data points, so that detection doesn't false-positive on empty metric series
22. As a Radar user, I want Beyla kind mapping to translate `k8s_src_owner_type` values (`Deployment`, `DaemonSet`, `ReplicaSet`, `StatefulSet`, `Pod`) to Radar's `Endpoint.Kind` vocabulary, so that the topology graph renders correct node types
23. As a Radar user, I want Beyla transport values (`TCP`, `UDP`) to map to Radar's protocol strings (`tcp`, `udp`), so that protocol display is consistent across sources
24. As a Radar user, I want Beyla external endpoints (no namespace) to be tagged as `Kind: "External"`, so that external traffic is visually distinguished from internal service traffic
25. As a Radar user, I want the Beyla Connect method to delegate to the shared Prometheus client's `EnsureConnected()`, so that Prometheus discovery and port-forwarding are handled consistently across all Prometheus-based sources

### Source priority ordering

Beyla is appended after Caretta, not inserted between Istio and Caretta:
```
["hubble", "istio", "caretta", "beyla"]
```

This avoids disrupting existing priority behavior. Caretta stays in its current position; Beyla is the lowest-priority fallback.

### BeylaSource struct

BeylaSource holds only a K8s client reference and a mutex — no persistent connections. All Prometheus interaction goes through the shared `internal/prometheus.Client` singleton (same pattern as IstioSource). This means BeylaSource doesn't own its own Prometheus discovery or port-forward; it piggybacks on the infrastructure Caretta/Istio already use.

### Detection strategy

Two-phase detection:
1. **Metric probe (primary):** Query `count(beyla_network_flow_bytes_total)` via the shared Prometheus client. If non-empty, Beyla is available.
2. **Pod label fallback (secondary):** If Prometheus is unreachable, check for pods with labels `app.kubernetes.io/name=alloy` or `app.kubernetes.io/name=beyla`. Count running pods.

Version extraction: Try `beyla_build_info` metric first, fall back to pod image tag.

Detection returns `Native: false` — Beyla is not built into any cloud provider's CNI.

### GetFlows queries

**L4 network flows:**
```promql
sum by (
  k8s_src_owner_name, k8s_src_namespace, k8s_src_owner_type,
  k8s_dst_owner_name, k8s_dst_namespace, k8s_dst_owner_type,
  dst_port, transport
) (
  rate(beyla_network_flow_bytes_total{job=~".*beyla.*|.*alloy.*"}[5m])
)
```

**L7 HTTP requests:**
```promql
sum by (
  k8s_src_owner_name, k8s_src_namespace,
  k8s_dst_owner_name, k8s_dst_namespace,
  http_method, http_route, http_status_code
) (
  rate(http_request_duration_milliseconds_count{job=~".*beyla.*|.*alloy.*"}[5m])
)
```

Namespace filter: append `and (k8s_src_namespace="<ns>" or k8s_dst_namespace="<ns>")` when `FlowOptions.Namespace` is set.

### L4+L7 merge logic

1. Execute L4 query → build map of `Flow` keyed by `(srcNs, srcName, dstNs, dstName, dstPort)`
2. Execute L7 query → for each result, find matching L4 flow and enrich with `L7Protocol`, `HTTPMethod`, `HTTPPath`, `HTTPStatus`
3. L7-only flows (no L4 match) → still emit with `Connections: 0`, `BytesSent: 0`

### Field mapping — owner-level (default)

When user hasn't configured extended attributes:

| Beyla label | Flow field |
|---|---|
| `k8s_src_owner_name` | `Source.Name`, `Source.Workload` |
| `k8s_src_namespace` | `Source.Namespace` |
| `k8s_src_owner_type` | `Source.Kind` (via `mapBeylaKind`) |
| `k8s_dst_owner_name` | `Destination.Name`, `Destination.Workload` |
| `k8s_dst_namespace` | `Destination.Namespace` |
| `k8s_dst_owner_type` | `Destination.Kind` (via `mapBeylaKind`) |
| `dst_port` | `Flow.Port`, `Destination.Port` |
| `transport` | `Flow.Protocol` |
| rate value | `Flow.Connections` |

### Field mapping — extended (with `attributes.select`)

When user has configured the recommended attributes:

| Beyla label | Flow field |
|---|---|
| `k8s_src_name` | `Source.Name`, `Source.Workload` |
| `k8s_src_namespace` | `Source.Namespace` |
| `k8s_src_type` | `Source.Kind` (via `mapBeylaKind`) |
| `k8s_dst_name` | `Destination.Name`, `Destination.Workload` |
| `k8s_dst_namespace` | `Destination.Namespace` |
| `k8s_dst_type` | `Destination.Kind` (via `mapBeylaKind`) |
| `dst_port` | `Flow.Port`, `Destination.Port` |
| `transport` | `Flow.Protocol` |
| rate value | `Flow.Connections` |
| `http_method` | `Flow.HTTPMethod` |
| `http_route` | `Flow.HTTPPath` |
| `http_status_code` | `Flow.HTTPStatus` (parsed to int) |

Always set: `Verdict: "forwarded"`, `LastSeen: time.Now()`

### Kind mapping

```go
func mapBeylaKind(beylaType string) string {
    switch strings.ToLower(beylaType) {
    case "pod":
        return "Pod"
    case "deployment", "replicaset", "statefulset", "daemonset":
        return "Workload"
    case "service":
        return "Service"
    default:
        return "Pod"
    }
}
```

### Manager registration

In `InitializeWithConfig()`: register `manager.sources["beyla"] = NewBeylaSource(client)`.

In `DetectSources()`: add `"beyla"` to `sourceOrder` after `"caretta"`.

In `generateRecommendation()`: when no source is available, suggest Alloy + Beyla for clusters without Cilium/Istio.

### Connect method

Delegates to `internal/prometheus.Client.EnsureConnected()` — no dedicated connection logic. BeylaSource doesn't maintain its own connection state.

### StreamFlows

10-second polling ticker, buffered channel (capacity 100), drop on full. Same pattern as Caretta and Istio.

## Testing Decisions

Tests should verify external behavior (detection results, flow construction, priority ordering) not internal implementation details (PromQL string formatting, HTTP request construction).

### Modules to test

- `internal/traffic/beyla.go` — `BeylaSource` detection and flow construction
- `internal/traffic/manager.go` — priority ordering includes Beyla

### Prior art

`internal/traffic/caretta_test.go` — the existing test file uses `httptest.NewServer` to mock Prometheus responses and verifies header propagation. The Beyla tests should follow the same pattern: mock HTTP server returning canned Prometheus responses, verify flow output matches expected structures.

### Test cases

| Test | What it verifies |
|---|---|
| `TestBeylaSource_Detect_MetricProbe` | Prometheus returns `beyla_network_flow_bytes_total` → detection succeeds |
| `TestBeylaSource_Detect_LabelFallback_Alloy` | Prometheus fails, Alloy pods exist → detection succeeds |
| `TestBeylaSource_Detect_LabelFallback_Beyla` | Prometheus fails, standalone Beyla pods exist → detection succeeds |
| `TestBeylaSource_Detect_NotAvailable` | No metric, no pods → detection fails gracefully |
| `TestBeylaSource_GetFlows_OwnerLevel` | Owner-level labels (default) → correct Flow fields |
| `TestBeylaSource_GetFlows_ExtendedLabels` | Extended labels → correct Flow fields with HTTP data |
| `TestBeylaSource_GetFlows_L4PlusL7` | L4 TCP flows enriched with L7 HTTP data |
| `TestBeylaSource_GetFlows_NamespaceFilter` | Namespace filter applied to PromQL queries |
| `TestBeylaSource_GetFlows_FallbackToOwner` | Extended query returns empty → falls back to owner-level |
| `TestBeylaSource_MapBeylaKind` | Kind string mapping (`Deployment` → `Workload`, etc.) |
| `TestBeylaSource_MapBeylaTransport` | Transport string mapping (`TCP` → `tcp`, etc.) |
| `TestManager_DetectSources_IncludesBeyla` | `DetectSources()` includes "beyla" in output |
| `TestManager_SourceOrder_BeylaLast` | Priority ordering is `["hubble", "istio", "caretta", "beyla"]` |

## Out of Scope

- **OTLP export support:** Prometheus-only for MVP. OTLP receiver infrastructure is a separate feature.
- **Beyla eBPF kernel version requirements:** Document but don't enforce (Linux 5.8+ with BTF).
- **Alloy auto-discovery differentiation:** Radar detects both Alloy and standalone Beyla pods but doesn't differentiate them in the UI. Future work could show deployment model.
- **Beyla-specific Helm chart installation:** The recommendation system will point to docs but won't include a pre-built Helm chart for Alloy + Beyla (unlike Caretta which has `carettaHelmChart()`).
- **Per-resource drift detection:** This is a TrafficSource concern, not a resource-renderer concern.

## Further Notes

- Beyla is pre-1.0. Metric schema stability is a known risk. The `job` regex (`job=~".*beyla.*|.*alloy.*"`) and metric names (`beyla_network_flow_bytes_total`, `http_request_duration_milliseconds_count`) are based on current documentation and may change. Mitigation: pin to documented names, add metric existence check in detection.

- The shared `internal/prometheus.Client` singleton means BeylaSource doesn't need its own Prometheus discovery or port-forward. However, if the shared client's discovery fails (e.g., no Prometheus in the cluster), BeylaSource's `GetFlows()` will return an empty response with a warning — same behavior as IstioSource.

- The `attributes.select` configuration is user-side (Beyla/Alloy config). Radar documents the recommended configuration but doesn't enforce it. Without it, Radar degrades gracefully to owner-level names.

- External endpoints (no namespace) are tagged as `Kind: "External"` to match the pattern used by Caretta and Istio sources.

rafaelpissolatto:
## Description

Adds **Grafana Beyla** as a fourth traffic source, alongside Hubble, Istio and Caretta.

Beyla derives network flows from eBPF without requiring a service mesh or CNI-level
integration, which makes it the first traffic source that works on a plain cluster.
Radar reads its metrics from any Prometheus-compatible endpoint - Prometheus, Mimir,
or whatever Alloy remote-writes to - rather than talking to Beyla directly.

## Implementation

- `BeylaSource` implementing the `TrafficSource` interface: `Connect`, `Detect`,
  `GetFlows`, `StreamFlows`, `Close`
- Detection probes `beyla_network_flow_bytes_total`, with a pod-label fallback that
  recognises both Alloy-embedded and standalone Beyla deployments
- **L4** flows built from `beyla_network_flow_bytes_total`, keyed on owner-level
  attributes (`k8s_src_owner_name` / `k8s_dst_owner_name`) so edges land on
  Deployments rather than individual pods
- **L7** enrichment (HTTP method/route/status) comes from
  `http_server_request_duration_seconds`, matched onto L4 flows by **destination
  only**. That metric is recorded server-side and carries no source-side labels at
  all, so pairing by (src, dst) the way L4 does isn't possible. The destination
  workload is resolved from Beyla's own `k8s_owner_name` label - the same value the
  L4 metric exposes as `k8s_dst_owner_name` - rather than a hand-built
  Deployment/ReplicaSet/StatefulSet/DaemonSet fallback chain, so both metrics agree
  on the same destination identity.
- A destination's `RequestRate` is split across its L4 edges weighted by byte share
  (an even split when no bytes are available), rather than copied onto each edge
  verbatim - copying it would overcount by the number of edges.
- Enrichment is skipped when a destination has more than one distinct port, since
  the L7 metric has no port label to say which one is actually HTTP (e.g. an app
  port alongside a raw-TCP DB port on the same workload). The port count is taken
  from *every* port Beyla reported for that destination, not just the ports that
  survived into the result - an L4 series with an unresolved source name (e.g. an
  external caller Beyla can't name) is dropped from the output but still counted
  here, so a destination whose real HTTP port drops out doesn't look falsely
  single-port and get HTTP metadata attached to an unrelated surviving port.
- `--beyla-job-selector` flag overrides the default PromQL `job` label matcher
  (`job=~".*beyla.*|.*alloy.*"`) for clusters running Alloy/Beyla under a
  non-standard job name
- Manager wiring: registration and priority ordering - Beyla is tried last, so an
  existing Hubble or Istio install still wins

## Notes for reviewers

- Namespace filtering on the **L4** query is emitted as two OR'd selectors
  (`sum(...{src="ns"}) or sum(...{dst="ns"})`). PromQL cannot express "src OR dst
  namespace" inside a single selector - a bare matcher after `and` is a parse
  error, not just a stylistic issue. This matches the idiom already used in
  `istio.go`. The **L7** query doesn't need this: since the metric has no source
  side, it filters on a single `k8s_namespace_name` matcher.
- `l4Key` includes `transport` in addition to `dst_port` - two series to the same
  (src, dst, port) but different transport (e.g. DNS on port 53, TCP vs UDP) would
  otherwise collide in `l4Map` and one would silently overwrite the other.
- `Connections` is used downstream in `pkg/traffic/aggregation.go` as an
  aggregation weight (`agg.Connections`, and the weight for L7-protocol voting), so
  it's kept explicitly non-zero on every flow. It's not the same thing as
  `RequestRate`: `BytesSent` (L4) and `Connections` (both) are derived from the
  Prometheus rate by multiplying back up by the 5m window; `RequestRate` (L7) is
  kept as the rate itself and used as-is downstream.
- Series with an empty source or destination name are dropped rather than emitted
  as anonymous nodes the UI cannot resolve or navigate to.

## Type of change

Purely additive - the source is only used when detected, and detection is
last in priority order, so existing Hubble/Istio/Caretta users see no change.

## How has this been tested?

- [x] Tested locally with minikube/kind (and deployed Grafana-Beyla as Daemonset)

radar_with_beyla
radar_with_beyla_2

#### Unit tests

18 test functions covering metric-probe and pod-label detection, owner-level and
extended-label flows, the L4+L7 merge (destination-only matching, the multi-port
skip guard, TCP/UDP key collision, and HTTP metadata misattachment when the real
HTTP port has an unresolved source name), namespace filtering, kind/transport
mapping, and PromQL shape assertions for both L4 and L7 queries.

#### Live validation - L4 (original testing session)

Validated end-to-end on a `kind` cluster running Beyla (DaemonSet, `hostPID` +
`hostNetwork`), Alloy scraping via ServiceMonitor, and Mimir as the query backend:

| Check | Result |
|---|---|
| `Detect` | `available=true`, "Beyla detected via Prometheus metrics" |
| `GetFlows` | 418 flows |
| Namespace filter | `demo-backend` → 24, `demo-frontend` → 223, no cross-namespace leakage |

Representative flows:

```
demo-frontend/frontend (Workload) -> demo-backend/backend (Workload)  port=80 tcp  bytes=1649148
demo-frontend/frontend (Workload) -> kube-system/coredns  (Workload)  port=53 udp  bytes=784348
demo-frontend/frontend (Workload) -> kube-system/kube-dns (Service)   port=53 udp  bytes=587614
```

At this point the L7 path was still unvalidated: Beyla's application-level probes
failed to load on the test kernel (WSL2) with `BPF program is too large. Processed
1000001 insn`, so no HTTP metrics existed to exercise the merge against, and the
metric name (`http_request_duration_milliseconds_count`, guessed rather than
confirmed) turned out to be wrong - see below.

A build-tagged live test (`//go:build livebeyla`) is included so this is
reproducible; it is excluded from normal builds and CI.

#### Live validation - L7 (follow-up session, different cluster)

The wrong metric name was caught in review and confirmed against Beyla's own
`docs/sources/metrics.md`: Beyla exports `http_server_request_duration_seconds`
(OTel-aligned, seconds not ms), with `http_request_method` / `http_route` /
`http_response_status_code` / `k8s_owner_name` labels - not the network-flow-style
labels this PR originally queried. The L7 merge was redesigned around
destination-only matching to fit the metric's actual shape (see Implementation).

Re-ran the `livebeyla` suite against a live Mimir + Beyla v3.28.0 install (a
different cluster than the one above, kernel not affected by the WSL2 BPF-size
issue): `demo-frontend/frontend -> demo-backend/backend` came back correctly
enriched with `l7=HTTP GET /missing`, and the live metric/label names matched
exactly what the code now queries. `Detect`, `GetFlows`, and namespace filtering
all passed against real data.

One live-only finding: this cluster's Beyla config deliberately drops `dst.port`
from `beyla_network_flow_bytes`'s attribute selection (ephemeral UDP ports were
blowing past Mimir's per-series cardinality limit) - a legitimate, common
operational trade-off. Under that config every L4 flow reports `port=0`, so the
multi-port skip guard can never trigger there - every destination looks trivially
single-port. This degrades gracefully (no crash, no wrong data on *this* cluster),
but it does mean clusters that disable per-port cardinality on L4 lose the guard's
protection: if a destination genuinely serves both an HTTP port and a non-HTTP
port, there's no signal left to tell them apart. Querying the live metric directly
did show a `server_port` label on `http_server_request_duration_seconds_count`
itself, so a future fix could match L7 to L4 by exact port when both metrics carry
one - that's not implemented here, filed as a known limitation rather than
blocking.

## Checklist

## Related issues

Part-Of and Closes #1205

---

> [!NOTE]
> **Low Risk**
> Additive traffic integration with last-priority selection; existing Hubble/Istio/Caretta behavior is unchanged unless Beyla is the only detected source.
> 
> **Overview**
> Adds **Grafana Beyla** (standalone or via Alloy) as a fourth Traffic view source, reading eBPF network and HTTP metrics from Prometheus instead of talking to Beyla directly.
> 
> A new `BeylaSource` implements the existing `TrafficSource` contract: detection probes `beyla_network_flow_bytes_total` (with Alloy/Beyla pod-label fallback), builds L4 edges from owner-level labels, and enriches them from `http_server_request_duration_seconds_count` matched by **destination only** (the HTTP metric has no caller labels). L7 metadata is skipped when a workload exposes more than one L4 port, and request rates are split across edges by byte share to avoid overcounting.
> 
> Wiring includes manager registration, detection order **Hubble → Istio → Caretta → Beyla** (Beyla only wins when nothing else is available), Prometheus `Connect` handling, and **`--beyla-job-selector`** to override the default `job=~".*beyla.*|.*alloy.*"` PromQL scope. README Traffic docs are updated accordingly; unit tests cover detection, merge edge cases, and PromQL shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a0d03555eaf43f0b1b25fb661d634bc4db89d77. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/traffic/beyla.go
+++ b/internal/traffic/beyla.go
@@ -1,0 +1,543 @@
+package traffic
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/skyhook-io/radar/internal/portforward"
+	promclient "github.com/skyhook-io/radar/internal/prometheus"
+	"github.com/skyhook-io/radar/pkg/prom"
+)
+
+const (
+	// defaultBeylaJobSelector is used unless overridden via SetBeylaJobSelector
+	// (wired to --beyla-job-selector) for clusters running Alloy/Beyla under a
+	// non-default Prometheus job name.
+	defaultBeylaJobSelector = `job=~".*beyla.*|.*alloy.*"`
+	// Rate window in the PromQL queries; used to turn per-second rates back
+	// into absolute counts for the window.
+	beylaRateWindowSeconds = 300
+)
+
+// beylaJobSelector returns the PromQL job-label matcher fragment (e.g.
+// `job=~".*beyla.*"`) used to scope Beyla queries, honoring any operator
+// override.
+func beylaJobSelector() string {
+	if selector := BeylaJobSelector(); selector != "" {
+		return selector
+	}
+	return defaultBeylaJobSelector
+}
+
+type promQueryFunc func(ctx context.Context, query string) (*prom.QueryResult, error)
+
+// BeylaSource implements TrafficSource for Grafana Beyla via Prometheus metrics.
+type BeylaSource struct {
+	k8sClient kubernetes.Interface
+	queryFn   promQueryFunc
+}
+
+// NewBeylaSource creates a new Beyla traffic source wired to the shared Prometheus client.
+func NewBeylaSource(client kubernetes.Interface) *BeylaSource {
+	s := &BeylaSource{k8sClient: client}
+	s.queryFn = s.defaultQuery
+	return s
+}
+
+func (s *BeylaSource) Name() string { return "beyla" }
+
+func (s *BeylaSource) defaultQuery(ctx context.Context, query string) (*prom.QueryResult, error) {
+	client := promclient.GetClient()
+	if client == nil {
+		return nil, fmt.Errorf("prometheus client not initialized")
+	}
+	return client.Query(ctx, query)
+}
+
+func (s *BeylaSource) query(ctx context.Context, query string) (*prom.QueryResult, error) {
+	return s.queryFn(ctx, query)
+}
+
+// Connect delegates to the shared Prometheus client's EnsureConnected.
+func (s *BeylaSource) Connect(ctx context.Context, contextName string) (*portforward.ConnectionInfo, error) {
+	client := promclient.GetClient()
+	if client == nil {
+		return &portforward.ConnectionInfo{Connected: false, Error: "Prometheus client not initialized"}, nil
+	}
+	_, _, err := client.EnsureConnected(ctx)
+	if err != nil {
+		return &portforward.ConnectionInfo{Connected: false, Error: fmt.Sprintf("Failed to connect to Prometheus: %v", err)}, nil
+	}
+	status := client.GetStatus()
+	info := &portforward.ConnectionInfo{Connected: true, Address: status.Address, ContextName: contextName}
+	if status.Service != nil {
+		info.Namespace = status.Service.Namespace
+		info.ServiceName = status.Service.Name
+	}
+	return info, nil
+}
+
+func (s *BeylaSource) Close() error { return nil }
+
+func (s *BeylaSource) Detect(ctx context.Context) (*DetectionResult, error) {
+	result := &DetectionResult{Available: false}
+
+	// Phase 1: metric probe via Prometheus. Scoped to the same jobs the flow
+	// queries read, so detection can't succeed on metrics GetFlows won't see.
+	qr, err := s.query(ctx, fmt.Sprintf(`count(beyla_network_flow_bytes_total{%s})`, beylaJobSelector()))
+	if err == nil && qr != nil && len(qr.Series) > 0 {
+		result.Available = true
+		result.Native = false
+		result.Message = "Beyla detected via Prometheus metrics"
+		result.Version = s.detectVersion(ctx)
+		return result, nil
+	}
+
+	// Phase 2: pod label fallback
+	if pods := s.countBeylaPods(ctx); pods > 0 {
+		result.Available = true
+		result.Native = false
+		result.Message = fmt.Sprintf("Beyla detected via %d running pod(s) (Alloy or standalone)", pods)
+		return result, nil
+	}
+
+	result.Message = "Beyla not detected. Install Alloy + Beyla for L7 traffic visibility."
+	return result, nil
+}
+
+func (s *BeylaSource) detectVersion(ctx context.Context) string {
+	qr, err := s.query(ctx, `beyla_build_info`)
+	if err != nil {
+		return ""
+	}
+	for _, series := range qr.Series {
+		if v := series.Labels["version"]; v != "" {
+			return v
+		}
+		if v := series.Labels["beyla_version"]; v != "" {
+			return v
+		}
+	}
+	return ""
+}
+
+func (s *BeylaSource) countBeylaPods(ctx context.Context) int {
+	count := 0
+	for _, label := range []string{"app.kubernetes.io/name=alloy", "app.kubernetes.io/name=beyla"} {
+		pods, err := s.k8sClient.CoreV1().Pods("").List(ctx, metav1.ListOptions{LabelSelector: label})
+		if err != nil {
+			log.Printf("[beyla] Failed to list pods matching %s: %v", label, err)
+			continue
+		}
+		for i := range pods.Items {
+			if pods.Items[i].Status.Phase == corev1.PodRunning {
+				count++
+			}
+		}
+	}
+	return count
+}
+
+// l4Key uniquely identifies an L4 flow for dedup. Must include every label
+// beylaL4GroupBy groups by (dst_port, transport) — otherwise distinct series
+// (e.g. TCP and UDP to the same endpoint/port) collide and one is dropped.
+type l4Key struct {
+	srcNs, srcName string
+	dstNs, dstName string
+	dstPort        int
+	protocol       string
+}
+
+// dstKey identifies a destination workload only. Beyla's HTTP server-duration
+// metric is recorded server-side and carries no source labels at all — it
+// can't be paired by (src, dst) the way the network-flow metric can — so L7
+// results are matched onto L4 flows by destination alone.
+type dstKey struct {
+	dstNs, dstName string
+}
+
+func (s *BeylaSource) GetFlows(ctx context.Context, opts FlowOptions) (*FlowsResponse, error) {
+	flows, err := s.getFlowsInternal(ctx, opts)
+	if err != nil {
+		log.Printf("[beyla] Error fetching flows: %v", err)
+		return &FlowsResponse{Source: "beyla", Timestamp: time.Now(), Flows: []Flow{},
+			Warning: fmt.Sprintf("Failed to query Beyla metrics: %v", err)}, nil
+	}
+	return &FlowsResponse{Source: "beyla", Timestamp: time.Now(), Flows: flows}, nil
+}
+
+func (s *BeylaSource) getFlowsInternal(ctx context.Context, opts FlowOptions) ([]Flow, error) {
+	l4Flows, dstPorts, err := s.queryL4(ctx, opts)
+	if err != nil {
+		return nil, fmt.Errorf("L4 query: %w", err)
+	}
+
+	l7Flows, err := s.queryL7(ctx, opts)
+	if err != nil {
+		log.Printf("[beyla] L7 query failed (continuing with L4 only): %v", err)
+		l7Flows = nil
+	}
+
+	l4Map := make(map[l4Key]*Flow, len(l4Flows))
+	byDst := make(map[dstKey][]*Flow, len(l4Flows))
+	for i := range l4Flows {
+		f := &l4Flows[i]
+		l4Map[l4FlowKey(f)] = f
+		dst := dstKey{f.Destination.Namespace, f.Destination.Name}
+		byDst[dst] = append(byDst[dst], f)
+	}
+
+	// Merge L7 into L4 by destination only: the HTTP server-duration metric is
+	// recorded on the serving side and has no source labels, so there's no way
+	// to attribute a request to a particular caller. A destination may have
+	// several L4 edges (different callers/ports); each gets a share of the
+	// destination's total HTTP metadata. L7 series with no matching L4
+	// destination are dropped — without a source there's no edge to draw.
+	for _, l7 := range busiestL7PerDst(l7Flows) {
+		dst := dstKey{l7.Destination.Namespace, l7.Destination.Name}
+		existing, ok := byDst[dst]
+		if !ok {
+			continue
+		}
+		// The HTTP metric has no port label, so if this destination fans out
+		// over more than one distinct L4 port there's no way to tell which one
+		// is actually HTTP (e.g. an app port alongside a raw-TCP DB port) —
+		// skip rather than mislabel every port as HTTP. Checked against
+		// dstPorts (every port Beyla reported for this destination) rather
+		// than len(existing): a port can be dropped from existing/l4Map for
+		// having an unresolved (e.g. external) source name while still being
+		// the real HTTP port, and existing's remaining single port would
+		// otherwise look unambiguous.
+		if len(dstPorts[dst]) != 1 {
+			continue
+		}
+		// l7.RequestRate describes the whole destination, so it must be split
+		// across its L4 edges rather than copied onto each one — otherwise
+		// aggregation overcounts by the number of edges. Weight the split by
+		// each edge's L4 byte volume as the best available proxy for its share
+		// of the destination's request traffic.
+		var totalBytes int64
+		for _, f := range existing {
+			totalBytes += f.BytesSent + f.BytesRecv
+		}
+		for _, f := range existing {
+			f.L7Protocol = l7.L7Protocol
+			f.HTTPMethod = l7.HTTPMethod
+			f.HTTPPath = l7.HTTPPath
+			f.HTTPStatus = l7.HTTPStatus
+			if totalBytes > 0 {
+				share := float64(f.BytesSent+f.BytesRecv) / float64(totalBytes)
+				f.RequestRate = l7.RequestRate * share
+			} else {
+				f.RequestRate = l7.RequestRate / float64(len(existing))
+			}
+		}
+	}
+
+	result := make([]Flow, 0, len(l4Map))
+	for _, f := range l4Map {
+		result = append(result, *f)
+	}
+	return result, nil
+}
+
+func busiestL7PerDst(l7Flows []Flow) []Flow {
+	best := make(map[dstKey]Flow, len(l7Flows))
+	topRate := make(map[dstKey]float64, len(l7Flows))
+	for _, f := range l7Flows {
+		dst := dstKey{f.Destination.Namespace, f.Destination.Name}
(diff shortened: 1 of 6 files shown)


nadaverell:
Hey @rafaelpissolatto just checking if you need help pushing this through - would be nice to get Beyla support merged into Radar. Thanks for working on it!

internal/traffic/beyla.go
```diff
@@ -0,0 +1,419 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/skyhook-io/radar/internal/portforward"
+	promclient "github.com/skyhook-io/radar/internal/prometheus"
+	"github.com/skyhook-io/radar/pkg/prom"
+)
+
+const (
+	beylaJobRegex = `job=~".*beyla.*|.*alloy.*"`
```
rafaelpissolatto: For the reviewers: Wondering if we should make this configurable?

rafaelpissolatto: Already made this as part of this initial feature

internal/traffic/beyla.go
```diff
@@ -0,0 +1,526 @@
...
+		if !ok {
+			continue
+		}
+		// The HTTP metric has no port label, so if this destination fans out
+		// over more than one distinct L4 port there's no way to tell which one
+		// is actually HTTP (e.g. an app port alongside a raw-TCP DB port) —
+		// skip rather than mislabel every port as HTTP.
+		if distinctPorts(existing) != 1 {
+			continue
+		}
```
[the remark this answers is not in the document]
rafaelpissolatto: @cursor review

rafaelpissolatto:
cursor review

rafaelpissolatto:
@nadaverell the PR is ready for review; please take a look when you have chance. Thanks

PR #1253, opened 2026-07-22
feat(web): add PWA support for browser installability

bannert1337:
Adds Progressive Web App support so browsers can advertise Radar as installable.

### Changes
- Add `vite-plugin-pwa` dependency and `VitePWA` plugin with web manifest
- Generate PWA icons (192x192, 512x512, 512x512 maskable) from existing `favicon.svg`
- Configure Workbox with `NetworkFirst` for `/api/*` routes — always live K8s data when online, last-known-good when offline
- Add `PWAUpdatePrompt` component using `useRegisterSW()` for user-controlled service worker updates
- Wire `PWAUpdatePrompt` into `App.tsx` alongside existing `UpdateNotification`

### Service Worker Strategy
| Traffic | Strategy |
|---|---|
| Static assets (JS/CSS, hash-versioned) | CacheFirst (precache) |
| API GET (`/api/*`) | NetworkFirst (10s timeout, 5min TTL) |
| API mutations | NetworkOnly |

### Manifest
- `display: standalone`, `theme_color: #0f172a` (matches Radar's dark theme)
- Icons at 192x192, 512x512, and 512x512 maskable (Android adaptive)
- PWA features only active in standalone build; no impact on library consumers

### Verified
- Manifest loads correctly in DevTools → Application tab
- Service worker registers on `localhost` with `SW_DEV=true`
- Tested in Floorp (Firefox-based) and Helium browsers

---

> [!NOTE]
> **Medium Risk**
> Service worker caching of `/api/*` can serve stale cluster data offline or after network blips; users must understand Reload applies a new SW. PWA registration affects all standalone builds, not only “installed” users.
> 
> **Overview**
> Adds **Progressive Web App** support to the Radar web app so browsers can treat it as installable and ship updates through a service worker.
> 
> **Build & config:** `vite-plugin-pwa` is wired in `vite.config.ts` with `registerType: 'prompt'`, a **Radar** web manifest (`standalone`, dark theme colors, 192/512 icons), Workbox precaching for static assets, and **NetworkFirst** runtime caching for `GET /api/*` (10s network timeout, 5‑minute cache). Local SW testing is gated on `SW_DEV=true`. TypeScript picks up PWA React types via `tsconfig.json`.
> 
> **UI:** New **`PWAUpdatePrompt`** uses `useRegisterSW()` to show a bottom-right toast when a new SW is waiting; **Reload** calls `updateServiceWorker(true)`, **Dismiss** clears the flag. It is mounted in **`App.tsx`** next to the existing **`UpdateNotification`** (desktop/binary updates), so both paths can coexist.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8acc0ff940b8957b2852f0c1016dacb3226d0699. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hisco:
Thanks for this, and sorry for the slow reply.

Before we decide, can you tell us a bit about how you use Radar? Specifically: how you run it, which browser you would install it into, and what you were hoping to get from having it installed. We are trying to work out who this helps most, and we would rather ask than assume.

The reason we are unsure: Radar runs three ways. As a binary you start yourself, as our desktop app, and in a cluster behind a web address. For the first group, an installed window has no address bar, so if the binary is not running there is nothing on screen to explain why. Those users also already have our desktop app. For the third group there is a real gap, and an installed icon would help, but only as an icon.

If your setup is the third one, or something we have not thought of, that would change the picture.

Two notes on the patch regardless of what we decide. `start_url` and `scope` are set to `/`, which breaks when Radar is served under a path prefix with `--base-path`. Relative values (`./`) stay correct there. And `web/package.json` replaces the `vite` devDependency rather than adding alongside it, which would break the build.

PR #1265, opened 2026-07-24
feat(ai): add support for antigravity, opencode, and pi CLI integrations

Paslestrange:
This PR adds auto-detection, configuration setup, and active diagnostic runner drivers for three additional AI agent CLI tools as requested in discussion #1252:

1. **Antigravity** (checks for the `agy` / `antigravity` binaries, configures `~/.gemini/antigravity-cli/config.json`)
2. **OpenCode** (checks for the `opencode` binary, configures `~/.config/opencode/opencode.json`)
3. **Pi CLI** (checks for the `pi` binary, configures `~/.pi/agent/mcp.json`)

### Key Changes
* **Backend CLI Drivers (`internal/ai/agent_antigravity.go`, `internal/ai/agent_opencode.go`, `internal/ai/agent_pi.go`):** Implemented the `Agent` interface drivers for these CLIs, mapping target prompt/remediation variables, session continuation flag mappings, model selectors, and standard output log stream parsing.
* **Agent Resolution (`internal/ai/agent.go`, `internal/ai/diagnoser.go`):** Added the new drivers to the `resolveAgent` switch-case and added them to `agentCLICandidates` to mark them as fully `Supported` by Radar's diagnosis executor.
* **Auto-Detection (`internal/ai/detect.go`):** Added the new agents to `knownAgents` and resolved binary names dynamically (such as mapping `antigravity` to `agy` dynamically via LookPath).
* **Frontend Setup UI (`web/src/components/home/MCPSetupDialog.tsx`):** Consolidated configuration generation helper methods and created accurate MCP configs matching each agent's native format and schema directory path structure.

Addresses discussion #1252.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new full-local agent path that spawns the OpenCode CLI and writes per-run MCP config; mistakes in command/stream handling could break investigations or mis-report tool activity.
> 
> **Overview**
> Adds **OpenCode** as a first-class local investigation agent alongside Claude, Codex, and Cursor.
> 
> A new `opencodeAgent` driver runs `opencode run` with JSON streaming, optional `--session` / `--model`, and prepends the system prompt into the positional message. Each turn writes a workspace `opencode.json` that points Radar’s MCP as a remote server, and parses stdout JSON (plus fallback text) into thinking/step events and a resumable session id. OpenCode is wired through agent resolution, PATH detection, supported CLI candidates, **full-local-only** execution profiles, and `opencode:full-local` consent versioning.
> 
> The diagnose settings UI gets OpenCode-specific model hints, and the MCP setup dialog adds copy-paste config for `~/.config/opencode/opencode.json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c93365187b1261e2b6d300b218589dab088f8d3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/ai/agent.go
+++ b/internal/ai/agent.go
@@ -63,6 +63,12 @@ func resolveAgent(bin string) Agent {
 		return &cursorAgent{bin: bin}
 	case strings.Contains(base, "codex"):
 		return &codexAgent{bin: bin}
+	case strings.Contains(base, "agy") || strings.Contains(base, "antigravity"):
+		return &antigravityAgent{bin: bin}
+	case strings.Contains(base, "opencode"):
+		return &opencodeAgent{bin: bin}
+	case strings.Contains(base, "pi"):
+		return &piAgent{bin: bin}
 	default:
 		return &claudeAgent{bin: bin}
 	}
--- a/internal/ai/agent.go
+++ b/internal/ai/agent.go
@@ -63,11 +63,11 @@ func resolveAgent(bin string) Agent {
 		return &cursorAgent{bin: bin}
 	case strings.Contains(base, "codex"):
 		return &codexAgent{bin: bin}
-	case strings.Contains(base, "agy") || strings.Contains(base, "antigravity"):
+	case base == "agy" || base == "antigravity" || strings.HasPrefix(base, "agy-") || strings.HasPrefix(base, "antigravity-"):
 		return &antigravityAgent{bin: bin}
-	case strings.Contains(base, "opencode"):
+	case base == "opencode" || strings.HasPrefix(base, "opencode-"):
 		return &opencodeAgent{bin: bin}
-	case strings.Contains(base, "pi"):
+	case base == "pi" || base == "pi-coding-agent":
 		return &piAgent{bin: bin}
 	default:
 		return &claudeAgent{bin: bin}
--- a/internal/ai/agent_antigravity.go
+++ b/internal/ai/agent_antigravity.go
@@ -5,15 +5,18 @@ import (
 	"context"
 	"io"
 	"os/exec"
+	"regexp"
 	"strings"
 )
+
+var antigravityConvRe = regexp.MustCompile(`(?i)Conversation ID:\s*([a-zA-Z0-9-]+)`)
 
 // antigravityAgent drives the Antigravity CLI (`agy`).
 type antigravityAgent struct{ bin string }
 
 func (a *antigravityAgent) Name() string { return "antigravity" }
 
-func (a *antigravityAgent) SigninCmd() string { return "agy update" }
+func (a *antigravityAgent) SigninCmd() string { return "agy auth login" }
 
 func (a *antigravityAgent) command(ctx context.Context, s turnSpec) (*exec.Cmd, func(), error) {
 	args := []string{}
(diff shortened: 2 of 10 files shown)


nadaverell (commented):
Thanks again for pushing these integrations forward. We’ve now merged #1273, which adds the execution-profile and consent framework we wanted in place before adding more active diagnosis runners. Could you rebase this PR on `main` and adapt each new CLI to that framework?

For each agent, please declare only the execution profiles the CLI can honestly enforce. “Radar safeguards” requires verified isolation from the user’s other tools and MCP servers; otherwise expose only “Your <agent> setup,” validate that profile in the driver, and preserve the agent’s normal configuration/environment with the corresponding warning. In particular, Antigravity’s unconditional `--dangerously-skip-permissions` should not be presented as safeguarded unless the CLI also provides verified controls that disable its built-in and external tooling.

Please also address the current functional gaps before marking the agents supported: OpenCode should obtain its resumable `sessionID` from structured JSON output, and Pi’s Radar MCP access needs a verified/configured MCP extension rather than only writing a file Pi does not natively load.

Because these CLIs have independent config, tool-loading, streaming, and session semantics, please consider splitting them into one PR per agent if validating all three together becomes unwieldy. That isn’t required if you can verify each one thoroughly in this PR, but smaller PRs would let us review and land the proven integrations independently rather than blocking the whole contribution on the least-certain CLI.

The framework should keep each adaptation fairly contained: profile declaration, driver validation/behavior, and consent copy/tests for that agent. Happy to help clarify the intended contract.

nadaverell:
thanks for rebasing and taking another pass at this. i took another look and i think there are still a few things to sort out before this is ready:

- the new consent surfaces aren’t in `aiConsentVersions`, so i think all 3 integrations currently fail at the consent step (and `make test` catches this)
- pi still seems to depend on the separate `pi-mcp-extension`, but we only write its config and don’t install/load/check for the extension
- i think the antigravity MCP paths/schema don’t match its current docs, and afaict the CLI still doesn’t expose a conversation id in print mode, so follow-up turns may not work
- the “full local” behavior is a bit inconsistent: we scrub most of the user’s env, while the UI says it uses their normal setup. opencode also uses `--auto`, while the UI suggests the CLI’s sandbox still applies
- this branch is behind main again, and the new agent catalog added on main probably needs updating too

maybe it makes sense to split these up? opencode looks closest to working end-to-end, while pi and antigravity seem to need a bit more integration work / upstream verification.

for each one, could you include a quick real smoke test covering MCP discovery, initial diagnosis, capturing the session id, a resumed follow-up, and an apply turn? then add tests around whatever args/config/output format that test confirms.

happy to help with the small radar-side cleanup once those behaviors are verified, but i’m a little wary of us filling in the missing pieces based only on the CLI docs.

nadaverell:
Hey @Paslestrange you wanna continue this, or prefer that I pick it up and finish the implementation on top of your PR?

Paslestrange:
Hi @nadaverell, thanks for getting back to the topic. I was out for quite a while. Will get back to the topic soon.

nadaverell:
@Paslestrange great, welcome back :)

Paslestrange:
@nadaverell,
 
I took your suggestion to focus on opencode to keep things clean and easy

- Execution Profile & Consent: Registered opencode (`opencode:full-local`) under the execution profile and consent framework, so the consent step passes cleanly.
- Session & Output Parsing: Updated the driver to handle opencode's structured JSON output (`--format json --auto`), ensuring session IDs and token streams are captured reliably for follow-up turns.
- UI & Docs: Added opencode setup instructions in the MCP setup dialog and aligned the model selector hints.

Settings:
Screenshot 2026-09-18 at 11 54 25

Investigation: 
Screenshot 2026-09-18 at 11 53 51


Let me know if this looks good to you or if you'd like any other tweaks!

nadaverell (approved):
Thanks for the PR! LGTM, made some small changes to complete the integration, docs mention + tests. merging!

Paslestrange:
Thanks! Looking forward to the new version!

PR #1291, opened 2026-07-29
fix(resources): don't show false memory/CPU % on multi-container pods with partial limits

### LINKED ISSUE #1249: Memory usage tooltip mixes metrics from different containers in a multi-container pod (wrong % shown)
In the pod resource table, hovering over the memory column for a multi-container pod shows a tooltip combining `Usage` from one container with `Request`/`Limit` from a *different* container in the same pod — resulting in a wildly incorrect percentage.

**Example:** `kafka-0` in namespace `kafka` (StatefulSet `kafka`, 2 containers: `kafka` and `kafka-exporter`)

- Tooltip shows: `Usage 5.4Gi / Request 128Mi / Limit 256Mi → 2144%`
- Container `kafka` has **no** resources set (`resources: {}` in the pod spec) — it runs with `KAFKA_HEAP_OPTS=-Xms8G -Xmx8G`, so ~5.4Gi usage is expected and normal for this container.
- Container `kafka-exporter` has `requests.memory: 128Mi` / `limits.memory: 256Mi` — these are its own values, not the `kafka` container's.
- The UI is evidently summing/matching `usage` from one container's cgroup with `request`/`limit` from the sibling container, rather than computing per-container (or a correctly aggregated pod-level) usage vs. request/limit.

**Expected behavior:** Either:
1. Show per-container breakdowns (Usage/Request/Limit all from the same container), or
2. If aggregating at the pod level, sum usage across *all* containers and sum requests/limits across *all* containers (and clearly label it as pod-aggregate) — not usage from one container against limits from another.

**Impact:** This produces alarming false positives (here, a fake "2144% / near OOM" warning) on pods that are actually healthy, which can trigger unnecessary incident response or alert fatigue.

Image

hisco:
## Problem

The pod resource-table memory/CPU cells sum usage across all of a pod's containers, but divide that usage by a limit summed from only the containers that actually set one. On a multi-container pod where some containers omit a request/limit, this produces a misleading percentage - e.g. `404Mi / 256Mi 158% — at risk of OOM kill` - comparing the whole pod's usage against a limit that bounds just one sidecar. The pod is not actually at risk.

## Fix

The percentage (and the OOM guidance) is now shown only when **every running container** actually sets that request/limit. "Running containers" means regular containers plus native sidecars (initContainers with `restartPolicy: Always`, which consume steady memory).

When not every container sets the field, the cell falls back to plain usage with a structured tooltip that:
- keeps the Usage / Request / Limit grid,
- marks the request/limit as partial `(N/M containers)`,
- omits the percentage and OOM guidance, which would be misleading.

## Scope

Only affects pods where not every running container sets the field in question. Normal pods - where all containers set the request/limit - render exactly as before.

## Verification

- `make tsc` passes.
- Reproduced live on a kind cluster with a two-container pod (one container with no memory limit, one limited sidecar): before the fix the cell showed a false over-limit percentage; after the fix it shows plain usage with the partial-containers tooltip and no percentage.

Fixes #1249

https://claude.ai/code/session_01XgG563HEjaSEr2B3cdsn7q

---

> [!NOTE]
> **Low Risk**
> UI-only change to pod table cell rendering and tooltips; no auth, data, or API impact. Risk is limited to how resource metrics are displayed for edge-case multi-container pods.
> 
> **Overview**
> Fixes misleading **OOM / over-limit** percentages on pod rows when usage is summed across all containers but limits/requests are only summed from containers that define them.
> 
> Pod **CPU** and **memory** cells now treat a summed limit or request as a valid denominator only when **every running container** sets that field—regular containers plus native sidecars (`initContainers` with `restartPolicy: Always`). When limits are complete, behavior matches the previous limit-based bar and tooltip. When only requests are complete, the bar uses request as the denominator with a **partial** tooltip if limits are incomplete. Otherwise the cell shows **plain usage** with a tooltip that lists usage/request/limit, annotates partial values as `(N/M containers)`, and **omits** the percentage and OOM guidance.
> 
> Adds `runningContainerFieldCount` and `buildPartialResourceTooltip` to drive that logic. Pods where all containers set both fields are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f564967216ce24bd06492fe33ec9a435a9eb120e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/k8s-ui/src/components/resources/ResourcesView.tsx
+++ b/packages/k8s-ui/src/components/resources/ResourcesView.tsx
@@ -5850,6 +5850,80 @@ function GenericCell({ resource, column }: { resource: any; column: string }) {
 // KIND-SPECIFIC CELL RENDERERS
 // ============================================================================
 
+// A pod's summed request/limit is a meaningful ceiling for its summed usage
+// ONLY when every running container sets that field. If one container has no
+// memory limit (e.g. a JVM broker sized by -Xmx) while a sidecar sets a small
+// one, pod-usage / summed-limit compares the whole pod's usage against a limit
+// that bounds just the sidecar — reading as a false "over limit / OOM risk".
+// Native sidecars live in initContainers with restartPolicy: Always and consume
+// steady memory, so they count toward the running set too.
+function runningContainerFieldCount(
+  resource: any,
+  field: 'requests' | 'limits',
+  resourceName: 'cpu' | 'memory',
+): { withField: number; total: number } {
+  const regular = Array.isArray(resource?.spec?.containers) ? resource.spec.containers : []
+  const nativeSidecars = (Array.isArray(resource?.spec?.initContainers) ? resource.spec.initContainers : [])
+    .filter((c: any) => c?.restartPolicy === 'Always')
+  const running = [...regular, ...nativeSidecars]
+  const withField = running.filter((c: any) => Boolean(c?.resources?.[field]?.[resourceName])).length
+  return { withField, total: running.length }
+}
+
+function allRunningContainersHave(
+  resource: any,
+  field: 'requests' | 'limits',
+  resourceName: 'cpu' | 'memory',
+): boolean {
+  const { withField, total } = runningContainerFieldCount(resource, field, resourceName)
+  return total > 0 && withField === total
+}
+
+// Tooltip for the case where the pod's summed request/limit does NOT bound the
+// whole pod (some containers set neither). Keeps the same Usage/Request/Limit
+// grid as buildResourceTooltip — but shows the request/limit as partial (N of M
+// containers) and omits the % / OOM guidance, which would be misleading here.
+function buildPartialResourceTooltip(
+  usage: number,
+  request: number,
+  limit: number,
+  requestCount: { withField: number; total: number },
+  limitCount: { withField: number; total: number },
+  formatFn: (n: number) => string,
+) {
+  return (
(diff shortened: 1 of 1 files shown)


nadaverell:
Blocking: native sidecars still produce a falsely trusted percentage.

runningContainerFieldCount includes spec.initContainers with restartPolicy Always, so a pod with a regular container plus a native sidecar, both fully limited, is considered complete. But handleTopPods and topPodFromObject sum requests and limits only over pod.Spec.Containers, while GetAllPodMetricsLatest sums every container reported by metrics-server. The sidecar usage is therefore included in the numerator while its limit is omitted from the denominator.

Example: an app using 900Mi of a 1Gi limit plus a native sidecar using 200Mi of a 256Mi limit is displayed as 1.1Gi / 1Gi = 107%, with a false OOM-risk warning, even though neither container is near its own limit. The new completeness check certifies this value because it counts the sidecar while the producer does not.

Please make the backend producer and UI completeness check use the same effective-container set, share that logic across both producer paths, and add a native-sidecar regression test. Computing coverage alongside the backend sum would prevent the table, top_resources, MCP output, and workload rollups from diverging again.

hisco:
Superseded by #1298 — the per-container redesign fixes the same false over-limit bug at the root (native sidecars included in the sums) and reworks the display per container.

PR #1298, opened 2026-07-29
feat(resources): per-container memory/CPU display in the Pods table

### LINKED ISSUE #1249: Memory usage tooltip mixes metrics from different containers in a multi-container pod (wrong % shown)

hisco:
## Problem

The Pods table computed memory/CPU % from pod-level sums: usage summed across **all** containers (including native sidecars), but request/limit summed over `spec.Containers` **only**. On multi-container pods with heterogeneous or missing limits — a broker with no limit next to a small sidecar, or a native sidecar (initContainer with `restartPolicy: Always`) that sets its own limit — this produced false over-limit percentages and "at risk of OOM kill" warnings on healthy pods.

## What changed

A per-container model instead of pod-level sums.

**Backend**
- A shared union-sum helper sums request/limit over regular containers **plus native sidecars**, applied to both the table (`handleTopPods`) and agent/MCP (`topPodFromObject`) paths — so the pod-level % is honest everywhere, including library consumers that read the pod-level fields.
- The table endpoint additionally returns a per-container metrics array (usage/request/limit per container); omitted when a pod has a single running container.
- Existing pod-level fields are kept (additive, non-breaking).

**Frontend**
- The memory/CPU cell shows the **most-constrained container**: usage vs its own limit (can go red), or vs its request in a neutral color when there is no limit, or plain usage when neither is set. A faint tag marks unbounded containers.
- The hover tooltip breaks every container down against its own request/limit.
- The column sorts by that per-container pressure.
- The pod detail drawer resolves native-sidecar specs so their charts show limit lines.

## Scope / compatibility

- Additive API field; existing pod-level fields retained — no breaking change for `@skyhook-io/radar-app` / `@skyhook-io/k8s-ui` consumers.
- Supersedes #1291 (a bounded patch of the same bug).

## Verification

- `make tsc` and `make build` pass.
- Verified live on a kind cluster with multi-container pods: a native-sidecar pod that previously read a false ~126% now reads its most-constrained container honestly; requests-only pods show utilization vs request in a neutral color; pods with no limits show plain usage.

Follow-up in this PR: unit tests for the union-sum/per-container builder and the frontend constraint picker.

Fixes #1249

https://claude.ai/code/session_01XgG563HEjaSEr2B3cdsn7q

---

> [!NOTE]
> **Medium Risk**
> Changes how pod resource percentages and top-metrics request/limit totals are computed across API and UI; behavior shifts for sidecar-heavy pods but is covered by new unit tests and is intended to fix incorrect alerts rather than introduce new enforcement paths.
> 
> **Overview**
> Fixes **misaligned pod-level CPU/memory %** when usage included native sidecars but request/limit sums only counted `spec.containers`, which produced false over-limit readings on multi-container pods.
> 
> **Backend:** `SumRunningContainerResources` and `BuildPodContainerMetrics` treat **regular containers plus native sidecars** (`initContainers` with `restartPolicy: Always`) as the running set; one-shot init containers and stale usage-only names are excluded. The top-pods handler and `topPodFromObject` use the union sum for pod-level fields; the bulk endpoint adds optional **`containers`** (usage + request/limit per name) when more than one running container exists. `GetAllPodContainerMetricsLatest` supplies per-container usage from the metrics history store.
> 
> **Frontend:** CPU/memory cells use **`podAggregate` / `readContainer`** so the bar reflects summed limits when every container is limited, neutral bars vs summed requests when there are no limits, or plain usage with tags like **unbounded** / **no limit** when limits are partial or missing. Tooltips list each container against its own yardstick. **PodRenderer** resolves sidecar specs from init containers for chart limit lines.
> 
> Types and API clients gain **`ContainerResourceMetrics`**; existing pod-level fields remain additive.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f315d65024cefb7e69a8d22015c7cff7cf3609d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/k8s/metrics_history.go
+++ b/internal/k8s/metrics_history.go
@@ -13,6 +13,7 @@ type ContainerMetricsHistory = k8score.ContainerMetricsHistory
 type PodMetricsHistory = k8score.PodMetricsHistory
 type NodeMetricsHistory = k8score.NodeMetricsHistory
 type TopPodMetrics = k8score.TopPodMetrics
+type ContainerResourceMetrics = k8score.ContainerResourceMetrics
 type TopNodeMetrics = k8score.TopNodeMetrics
 type MetricsCollectionHealth = k8score.MetricsCollectionHealth
 type MetricsSourceHealth = k8score.MetricsSourceHealth
--- a/internal/k8s/per_container_resources_test.go
+++ b/internal/k8s/per_container_resources_test.go
@@ -1,0 +1,185 @@
+package k8s
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// resReq builds a ResourceRequirements from CPU/memory request+limit strings.
+// Empty strings are omitted so callers can model "no request" or "no limit".
+func resReq(cpuReq, cpuLim, memReq, memLim string) corev1.ResourceRequirements {
+	req := corev1.ResourceList{}
+	lim := corev1.ResourceList{}
+	if cpuReq != "" {
+		req[corev1.ResourceCPU] = resource.MustParse(cpuReq)
+	}
+	if memReq != "" {
+		req[corev1.ResourceMemory] = resource.MustParse(memReq)
+	}
+	if cpuLim != "" {
+		lim[corev1.ResourceCPU] = resource.MustParse(cpuLim)
+	}
+	if memLim != "" {
+		lim[corev1.ResourceMemory] = resource.MustParse(memLim)
+	}
+	return corev1.ResourceRequirements{Requests: req, Limits: lim}
+}
+
+func alwaysRestart() *corev1.ContainerRestartPolicy {
+	p := corev1.ContainerRestartPolicyAlways
+	return &p
+}
+
+// mixedPod: one regular container, one native sidecar (initContainer with
+// restartPolicy Always), one regular init container (no restartPolicy).
+func mixedPod() *corev1.Pod {
+	return &corev1.Pod{
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{Name: "app", Resources: resReq("50m", "100m", "64Mi", "128Mi")},
+			},
+			InitContainers: []corev1.Container{
+				// native sidecar — counted
+				{Name: "istio-proxy", RestartPolicy: alwaysRestart(), Resources: resReq("100m", "200m", "128Mi", "256Mi")},
+				// regular init container — excluded
+				{Name: "setup", Resources: resReq("500m", "500m", "512Mi", "512Mi")},
+			},
+		},
+	}
+}
+
+func TestSumRunningContainerResources(t *testing.T) {
+	got := SumRunningContainerResources(mixedPod())
+
+	// CPU nanocores = MilliValue * 1_000_000. app 50/100m + sidecar 100/200m.
+	wantCPUReq := int64((50 + 100) * 1_000_000)
+	wantCPULim := int64((100 + 200) * 1_000_000)
+	// Memory bytes. app 64/128Mi + sidecar 128/256Mi.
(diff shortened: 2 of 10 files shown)


nadaverell (approved):
LGTM. Didn't test out the core logic but visually looks nice.

nadaverell:
## Cross-review follow-up — compact cell redesigned (headline = pod aggregate)

A self-review + adversarial cross-model pass (Codex) confirmed two real problems with the compact CPU/Memory cell as it stood on this branch; the cell contract changes as a result. Fix lands in the next push.

### Confirmed findings

1. **A limited sidecar could hide the pod's dominant consumer.** The cell headlined the most-constrained *limited* container (unlabeled). Common failure shape: injected sidecars (istio-proxy) carry limits while app containers often don't — so a pod with an app using 8Gi unbounded plus a sidecar at 50Mi/512Mi rendered as `50Mi / 512Mi · 1 unbounded`, and in pods with 4+ containers the unbounded consumer could fall past the tooltip's 3-row cap too.
2. **CPU/Memory sorting became a no-op on clusters without requests/limits.** Sort used most-constrained percentage, which is 0 for pods with nothing set — every pod tied, and a runaway unlimited pod sorted *below* an idle limited one. Direct regression from raw-usage sorting for a common cluster shape (dev/kind).

### New cell contract

The headline is always **total pod usage**; per-container pressure is the detail layer (tooltip, worst-first):

- **Every container has a limit** → utilization bar (red-capable) against the summed limit — the sum is a real ceiling — with the summed-request marker.
- **No limits, some requests** → quiet bar against the summed request (no ceiling → never red).
- **Partial limit coverage** → plain usage + `N unbounded` tag. A partial limit sum is not a ceiling; no bar pretends it is.
- **Nothing set** → plain usage + `no limit` tag.
- **Sorting** is raw total usage again — always meaningful, matches the headline.

### Reviewed & skipped (with rationale)

- *"Latest container metrics can stay stale for the pod's lifetime"* — pre-existing store semantics, not introduced here: `GetAllPodMetricsLatest` has always summed the last sample of every container buffer. This branch narrows the exposure by dropping spec-absent names from the breakdown. Follow-up candidate: timestamp-based pruning in the metrics store.
- *"Active init/ephemeral container usage discarded"* — real but transient (init phase) or niche (ephemeral debug). Including it correctly needs pod-status plumbing to tell active from stale, and doing it naively would reintroduce the stale-completed-init inflation this branch just fixed. Known limitation.

Issue #80, opened 2026-02-05
Add support for database storage (mysql, psql)


greenaar:
## Problem statement

It's not hard to break sqlite. I've done it twice recently with radar, once when doing a helm update.

## Proposed solution

I'd like to see support for either mysql or postgres.  I'm not anticipating any built-in installation, but the ability to point this at a database and have it handle schema and migrations would be good.

## Alternatives considered

I've tried 'memory' mode -- it works, but it can be slow, and it's not resilient against pod restarts.

nadaverell:
Took a very long time for this... but finally we have postgres as well 🎉🐘
@greenaar and everyone else who waited for this @cosmincatalin @Conor-Behard333 @syscod3 @viethoang1211 
Check it out in latest release and let us know if anything is not working as intended.

PR #1314, opened 2026-08-01
feat(timeline): add PostgreSQL timeline storage

syscod3:
## Description

Radar can now persist timeline history in an externally managed PostgreSQL database, so events survive pod restarts and Helm upgrades without requiring Radar to install or operate a database.

The PostgreSQL adapter implements the existing timeline store contract, applies versioned transactional migrations under an advisory lock, validates the resulting schema at startup, and preserves the existing SQLite and memory behavior. The chart accepts only a same-namespace Secret reference; the DSN is read from `RADAR_TIMELINE_POSTGRES_DSN` and is never persisted to config or exposed as a CLI argument.

This implements the PostgreSQL portion of https://github.com/skyhook-io/radar/issues/80. MySQL support and SQLite history import are intentionally out of scope.

## Type of change

## How has this been tested?

- [x] Tested PostgreSQL integration against PostgreSQL 17
- [x] Ran `go test ./...` with `RADAR_TEST_POSTGRES_DSN`
- [x] Ran `go test ./...` in `pkg/`
- [x] Ran Helm lint, Helm unit tests, and chart rendering checks
- [x] Ran frontend typecheck, tests, lint, and build
- [x] Ran shared `k8s-ui` tests

## Checklist

## Related issues

Implements the PostgreSQL portion of https://github.com/skyhook-io/radar/issues/80.

---
### Personal Notes
I was in need of a more robust db for persistence and I decided to fork the db interface, this is being used on my personal cluster for some days and the code is heavily AI assisted, so I expect to be properly grilled on this PR, please bear with me.

Thank you.

---

> [!NOTE]
> **Medium Risk**
> New persistence path touches startup, migrations, and shared DB semantics for multi-replica deploys; misconfiguration or migration issues can block Radar boot, though credentials stay out of config files.
> 
> **Overview**
> Adds **PostgreSQL** as a third timeline backend alongside memory and SQLite, so cluster change history can live in an external database instead of a single PVC—useful for restarts, rolling updates, and multi-replica setups.
> 
> The new **`PostgresStore`** (`pgx`) implements the existing store contract: embedded migrations under an advisory lock, schema validation at startup, append/query/grouping behavior aligned with SQLite (including K8s event upserts and seq paging), age-based retention cleanup, and **no** SQLite-style max-size pruning. **`BuildTimelineStoreConfig`** now returns errors for unknown backends or postgres without a DSN; explorer/desktop read the DSN only from **`RADAR_TIMELINE_POSTGRES_DSN`** (not `config.json`). Failed postgres init is **fatal** (no memory fallback), and timeline init failures now fail subsystem startup.
> 
> **Helm** gains `timeline.storage=postgres`, `timeline.postgres.existingSecret` / `secretKey`, template guards when postgres is chosen without a secret, deployment env injection for the DSN, and unittest/CI coverage including a Postgres 17 service job for `./internal/timeline` tests. Docs and chart README describe operator-managed secrets and credential rotation (pod restart required).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 873ab97de4e157ed483ee7ff2c5efce9b7a2f697. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hisco:
Thank you for this, and sorry it sat for a while.

We reviewed it properly, in the spirit you asked for. The design held up: the
advisory-locked migration runner with a post-migration schema validation is
better than the ad-hoc path it sits beside, `resource_key bytea` is the only
correct choice given the NUL separator in the seen-resource key, and the
credential handling - env-only DSN, redaction over both DSN forms, and a
reflection test asserting the field cannot be serialized - is the strongest
part of the change.

We also found real problems, mostly in one place. The store implemented its own
tests but not the full `EventStore` contract: `Query` ignored `UntilSeq` and
`SequenceOrder`, `Stats` never reported the sequence bounds or the eviction
flag, and `timestamptz` truncated the nanosecond timestamps the timeline
carries. The shared conformance suite catches all four, which is why wiring the
new store into it was the right call on your part even though it is what turned
the job red.

One detail worth passing on: those failures are platform-dependent. On macOS you
would have seen three; on Linux there are seven. `time.Now()` is
microsecond-resolution on darwin, so it hides the truncation entirely. That is
almost certainly why the branch looked green to you locally.

We could not push to your branch - the fork is org-owned, so maintainer edits
do not apply - and it had diverged 121 commits. So we rebased your commits onto
`main` and pushed the fixes on top. Your commits are unchanged and you are
credited as co-author on the merge. It landed in #1632.

Two things we dropped: the in-cluster reconnect commit, because `main` had
already solved it a different way and it became unreachable after the rebase;
and the multi-replica recommendation in the chart docs, because a shared
timeline does not make Radar multi-replica while informer caches and SSE
streams are still per-pod.

Closing this in favour of #1632. Thanks again - this was a substantial piece of
work and it is in the product.

PR #1349, opened 2026-08-05
Stop resource-metrics Prometheus discovery from binding to caretta-vm

hisco:
## Problem

The reverse of #1347. Radar's resource-metrics module (per-pod CPU/memory charts) could bind to the traffic source's `caretta-vm`, which serves only Caretta's eBPF link metrics and commonly lacks cluster workload metrics -> charts go blank.

Two independent paths allowed it:

1. **Blind cross-owner port-forward reuse.** `discover()` reused any owner's managed forward via `GetAddress` **before** enumerating candidates, so it could adopt the traffic source's `caretta-vm` forward. The generic `/api/v1/query?query=up` probe passes against `caretta-vm`, so the wrong backend was accepted silently.
2. **Candidate ranking.** `caretta-vm` sat in the primary well-known list, so it outranked any Prometheus discoverable only via the dynamic cluster scan (non-standard name/namespace).

## Fix

1. **Target-aware reuse.** Reuse now happens **per candidate** via `portforward.GetAddressForService` (added in #1347) inside the priority-ordered port-forward fallback loop, so a forward is adopted only when it targets the candidate being tried. The now-unused `GetAddress` is removed. The reuse path also clears `discoveryService` on a superseded commit, matching the port-forward path.
2. **Fallback tier.** `caretta-vm` moves out of the primary well-known list into `fallbackWellKnownLocations`: fetched early (to seed dedup) but appended **last**, and suppressed by `namespace/name` in the dynamic scan so it can't resurface ranked-by-score. A real Prometheus always wins; `caretta-vm` is kept as a candidate only so a `caretta-vm`-only cluster still works. The fallback tier also survives a Services `list` failure (get-but-not-list RBAC).

## Tests

- Target-aware reuse, both directions (`internal/portforward`).
- `caretta-vm` ranks after a dynamic-only Prometheus; remains the sole candidate when it is the only backend; survives a `list` failure; stays single and last with a non-standard service port (`pkg/prom`).
- All tests pass with `-race` across `internal/portforward`, `internal/prometheus`, `internal/traffic`, and the `pkg/prom` module.

## Validation

Reproduced end-to-end on a kind cluster with **real** kube-prometheus-stack + real eBPF Caretta (149 `caretta_links_observed` series in caretta-vm): the pre-fix binary binds Caretta to the general Prometheus and returns **0 flows**; this fix opens a dedicated `caretta-vm` forward and returns real flows, leaving the resource-metrics Prometheus forward untouched.

## Follow-up to #1347

Builds on the `GetAddressForService` primitive introduced in #1347 (now merged). Rebased onto `main`.

---

> [!NOTE]
> **Medium Risk**
> Changes Prometheus discovery and shared port-forward selection paths used for resource metrics; wrong backend choice previously caused blank charts rather than crashes, but regressions could still affect metrics availability.
> 
> **Overview**
> Fixes resource-metrics (CPU/memory charts) silently binding to **caretta-vm**, which lacks cluster workload metrics.
> 
> **Port-forward reuse:** Removes blind pre-enumeration reuse via `GetAddress`. Reuse now runs **per candidate** in the port-forward loop using `GetAddressForService`, so only a tunnel to that namespace/service is adopted. Superseded reuse clears `discoveryService` without stopping peer-owned forwards; cancelled context bails before `Start` would tear down a reused tunnel.
> 
> **Candidate ranking:** Moves `caretta/caretta-vm` from the primary well-known list into a **fallback tier** (fetched early for dedup, appended last). Dynamic discovery suppresses fallback services by namespace/name so caretta-vm cannot outrank a dynamically found Prometheus; fallback still returns when list RBAC fails or caretta-vm is the only backend.
> 
> **Tests:** Port-forward target-awareness and caretta-vm ordering edge cases (`internal/portforward`, `pkg/prom`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d9e6f7fc69c371a2e09378c2b49b8f656ff8bbe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/portforward/portforward.go
+++ b/internal/portforward/portforward.go
@@ -8,8 +8,8 @@
 // starting or stopping one owner's forward never tears down the other's. (A
 // single shared forward previously let them clobber each other whenever they
 // wanted different services.) Owners may still read each other's forward
-// address (GetAddress) to reuse a compatible endpoint, but only ever stop their
-// own.
+// address (GetAddressForService) to reuse a forward that targets the same
+// service, but only ever stop their own.
 package portforward
 
 import (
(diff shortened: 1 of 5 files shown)


nadaverell (approved):
LGTM, but worth a quick test together with the other PR for sanity check

PR #1368, opened 2026-08-06
mcp: repair mistaken tool argument names, and report accepted ones on schema errors

nadaverell:
## Summary

Every radar MCP tool schema is `additionalProperties: false`, so an agent that guesses **one** argument name wrong gets a hard rejection — and the SDK's error lists the properties it did *not* expect without ever naming the ones it does. Agents rarely retry a schema error, so a single wrong guess silently costs an entire investigation, and it surfaces as a bad answer rather than a failed call.

This was hit for real. An agent called `get_subject_permissions` with `{"subject": …, "subjectKind": …}` — the nouns that tool's own description used — instead of `{"name": …, "kind": …}`:

```
validating "arguments": validating root: unexpected additional properties ["subject" "subjectKind"]
```

The tool never ran. The agent went on to report *"the ServiceAccount has no RBAC RoleBindings at all"* for a ServiceAccount whose `ClusterRoleBinding` existed and was merely missing the `patch` verb — a wrong root cause, from a tool that works correctly when called correctly.

## What changed

**1. Schema errors name the accepted arguments**, split required/optional:

```
validating "arguments": validating root: unexpected additional properties ["whoIsThis"]

get_subject_permissions accepts: kind, name (required); namespace, verb, resource,
group, resource_namespace, subresource, resource_name (optional). Retry with these names.
```

This is the load-bearing half: it applies to all **28/28** tools and covers argument mistakes that can't be predicted in advance. It keys off argument names the middleware could not resolve — plus required arguments detected as missing — rather than matching the SDK's rendered message, so an upstream wording change cannot silently disable it and a tool's own domain error is never mistaken for a schema failure.

**2. Argument repair before validation**, from two sources:

- *Orthographic* (global): snake_case ↔ camelCase of the same word. Always safe — same word, same meaning.
- *Semantic* (one tool): `subject`/`subjectKind` → `name`/`kind` on `get_subject_permissions`.

A name is rewritten **only** when the target is a real parameter of that tool, the caller did not already supply it, and — for semantic aliases — the value still means the same thing under the target field. That last check is subject-kind aware: ServiceAccount names are DNS subdomains, so `prod/cleanup` is refused; User and Group names are opaque identities where `system:authenticated` and `system:serviceaccount:ns:sa` are correct and are accepted. Without that, repair could turn a rejected call into a confidently wrong lookup — the same false negative this PR removes.

Repairs are logged, so a call that only worked because of repair is visible rather than looking like the agent got it right.

**3. The description that caused the misuse.** `get_subject_permissions` spoke of "subjects" throughout while its parameters are `kind`/`name`. It now names the parameters explicitly and states the name must be unqualified — fixing the bait at source rather than only compensating for it.

**4. `manage_gitops` parameters renamed to snake_case.** It was the only tool publishing camelCase (`dryRun`, `applyOnly`, `syncOptions`, `historyId`) in an otherwise snake_case server — and its own description already documented `dry_run`, so the schema contradicted the docs. radar now publishes snake_case with no exceptions, matching Anthropic's agent tooling (`file_path`, `old_string`, `run_in_background`) and kubectl's flags (`--dry-run`, `--all-namespaces`).

**This rename is not breaking in practice**: orthographic repair maps camelCase to snake_case, so existing callers sending `dryRun` or `historyId` keep working and are logged as repaired. Verified against a running server — both spellings reach the handler and return identical results. `TestParameterNamingStaysConsistent` now has zero grandfathered exceptions.

`addTool()` replaces `mcp.AddTool()` at all 28 registration sites, recording each tool's accepted/required argument names from the handler's input struct. `TestRegistryMatchesPublishedSchema` verifies that reflection against the schema the SDK actually publishes, for both the full and read-only servers — if they ever diverge, repair would stop recognising a real parameter and the help would name arguments that don't exist.

### Reviewer notes

- **The semantic alias list is one tool long, by design.** Adding to it needs evidence of an agent actually getting a name wrong, not a hunch — `pod` and `workload` were considered and rejected because `prod/api-0` and `deployment/api` are plausible values that are not bare names.
- Docs updated in a companion branch (`radar-docs`: `docs/mcp-gitops-snake-case-params`).

## Testing

- `go test ./...` — 32 packages pass · `go vet ./...` clean · `go test -race ./internal/mcp` clean · `gofmt` clean
- `-count=2`, `-count=3` and `-shuffle=on` all pass (the sentinel test mutates a process-global registry and restores it)
- Safety properties are tested, not restated: qualified values refused per subject kind, bare names still repaired, aliases never clobber a correctly-supplied value, non-object args and unregistered tools pass through, domain errors left unannotated
- Three tests drive **real** `tools/call` requests through the SDK over an in-memory transport — an unknown argument, a missing required argument, and a repaired call whose sentinel handler asserts it received the canonical names — so an integration break can't hide behind a hand-written error string
- Verified end-to-end against a running radar on a live cluster: the previously-rejected `subject`/`subjectKind` call returns the full bindings array; the published schema now has zero camelCase parameters; old camelCase `manage_gitops` callers still succeed via repair
- visual-test: skipped (Go-only, no rendered surface)

---

> [!NOTE]
> **Medium Risk**
> Middleware runs on every tools/call and semantic aliasing could mis-route values if guards regress; manage_gitops schema rename is mitigated by camelCase repair but still changes the published contract.
> 
> **Overview**
> Agents often hard-fail MCP tool calls when a single argument name is wrong (`additionalProperties: false`), which can derail investigations. This PR adds **receiving middleware** that rewrites common mistakes before validation and **appends accepted parameter names** when validation still fails.
> 
> **Repair** maps global snake_case ↔ camelCase (e.g. `dryRun` → `dry_run`) and a curated alias on `get_subject_permissions` (`subject`/`subjectKind` → `name`/`kind`), only when the target is a real parameter and semantic moves pass subject-kind-aware value checks so qualified identities are not silently misrouted. **`addTool`** replaces direct SDK registration at all tool sites to record accepted/required names from handler structs for help text and repair.
> 
> **`manage_gitops`** JSON fields are aligned to snake_case (`dry_run`, `history_id`, `sync_options`, etc.); docs and validation error strings match. **`get_subject_permissions`** description now documents `kind`/`name` explicitly instead of “subject” wording that invited the benchmark failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f10634fe8d99c5fd37ea4471561a6738b6ae3fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
Docs updated in skyhook-dev/radar-docs#70 (`manage_gitops` parameter table).

PR #1372, opened 2026-08-06
fix(navigation): stop double-pluralizing CRD plurals before discovery lands (RAD-330)

nadaverell:
Fixes [RAD-330](https://linear.app/skyhook-io/issue/RAD-330). Pre-existing; found during RAD-314 (Velero) and RAD-321 (Kyverno) live verification.

## The bug

`kindToPlural` already has a double-pluralization guard — it returns the input unchanged when it is a *known* plural. But "known" means `BUILTIN_PLURAL_TO_KIND` plus the map that `initNavigationMap` builds from API discovery, and on a **cold direct-URL load** the discovery response is still in flight when the detail drawer runs `kindToPlural` on the URL's plural slug. Any CRD plural absent from the small static map falls through to `englishPlural`, which pluralizes it a second time.

The drawer then fetches a nonexistent kind, and self-heals on the retry once discovery lands — which is why this has survived as a transient console error rather than a broken page. In-app navigation never hits it because the map is already warm.

## Root cause

Not a missing guard — an incomplete plural set at the moment the guard runs. So decide it structurally, with no round-trip required:

```ts
if (kind === kindLower && isEnglishPlural(kindLower)) return kindLower
```

Both halves are load-bearing:

- **`kind === kindLower`** — Kubernetes Kinds are PascalCase (`Schedule`), API resource names are lowercase (`schedules`). An all-lowercase input is a resource name.
- **`isEnglishPlural`** — a lowercase input can still be a *singular*. `pkg/topology/relationships.go`'s `normalizeKind` returns its input unchanged when a kind resolves through neither its static map nor discovery, so a `ResourceRef.Kind` can arrive as `validatingpolicy` or `certificaterequest`. The case check alone would have stopped pluralizing those — a regression this predicate prevents. It also keeps the singulars that merely *end* in 's' (`ingress`, `nodeclass`, `storageclass`, `ec2nodeclass`) pluralizing correctly, which a bare `endsWith('s')` would not.

`isEnglishPlural(n)` is `englishPlural(englishSingular(n)) === n` — a noun is plural iff the pluralizer reproduces it from its own singular. `englishSingular` is the new inverse of `englishPlural`; `pluralToKind` had hand-inlined the identical de-pluralization rules and now delegates to it so the two can't drift.

Ordering is preserved: the builtin-plural check, the discovered kind→plural lookup, and the lowercase alias map (`pvc`, `podgroup`, `horizontalpodautoscaler`) all run *before* the new branch.

## Verification

**Before** — cold load of `/resources/schedules?resource=demo/daily-backup` against a cluster with a `schedules.velero.io` CRD:

```
GET /api/resources/scheduleses/demo/daily-backup                 => 400 Bad Request
GET /api/resources/scheduleses/demo/daily-backup/cascade-preview => 200
GET /api/resources/schedules/demo/daily-backup                   => 200   (self-heal)
```

**After** — same cold load, same cluster:

```
GET /api/resources/schedules/demo/daily-backup                   => 200
GET /api/resources/schedules/demo/daily-backup/cascade-preview   => 200
```

`validatingpolicies` (cluster-scoped, `?resource=name` with no namespace) verified the same way: `validatingpolicieses` gone, no 4xx. `/resources/ingresses` re-checked as a core-kind regression guard.

## Tests

`navigation.test.ts`, under a cold (reset) map — the first two groups fail without this change with exactly `expected 'scheduleses' to be 'schedules'`:

- unknown lowercase plurals returned unchanged (`schedules`, `validatingpolicies`, `virtualservices`, `clusters`, `backups`)
- the cold answer equals the post-`initNavigationMap` answer for the same inputs
- singular PascalCase CRD kinds still pluralize
- **lowercase singular** kinds still pluralize (`certificaterequest`, `validatingpolicy`, `ingress`, `nodeclass`, `ec2nodeclass`, `storageclass`) — the guard-tightening case

`pluralize.test.ts` covers `englishSingular` and `isEnglishPlural` directly, including a round-trip property over `englishPlural`.

## Considered and skipped

RAD-330 asks for "a test that loads a CRD detail by cold direct URL". The Playwright e2e harness (`cmd/testserver`) is backed by a typed fake clientset with no dynamic client or discovery, so it cannot serve a CRD at all — adding that is a substantial change to shared test infrastructure. The unit tests pin the exact regression and the live-cluster trace above covers the cold-URL path.

---

> [!NOTE]
> **Medium Risk**
> Changes how resource API paths are built during the pre-discovery window; behavior is heuristic with known edge cases but is heavily tested and scoped to navigation pluralization.
> 
> **Overview**
> Fixes **cold direct-URL** loads where `kindToPlural` ran on a CRD plural slug (e.g. `schedules`) before API discovery filled the map, so `englishPlural` turned it into `scheduleses` and the first detail fetch 4xx’d until retry.
> 
> **`kindToPlural`** now returns lowercase inputs unchanged when `isEnglishPlural` says they’re already a regular plural (`kind === kindLower` + round-trip check), while still pluralizing lowercase **singulars** like `ingress`, `nodeclass`, and `certificaterequest`.
> 
> **`pluralize.ts`** adds `englishSingular` and `isEnglishPlural`; **`pluralToKind`** uses `englishSingular` instead of duplicated de-plural rules. Unit tests cover the cold-map CRD cases and the new helpers (including documented `*se` / lone-`s` limits).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3c4a8896d118d3f3ff018e179a9446911c84659. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/k8s-ui/src/utils/navigation.test.ts
+++ b/packages/k8s-ui/src/utils/navigation.test.ts
@@ -1,4 +1,5 @@
 import { describe, test, expect, afterEach } from 'vitest'
+import { englishPlural } from './pluralize'
 import { kindToPlural, pluralToKind, refToSelectedResource, initNavigationMap, resetNavigationMap, laneId, laneResourceKey, groupQualifiesLaneId, parseLaneId } from './navigation'
 
 afterEach(() => {
@@ -64,6 +65,66 @@ describe('kindToPlural', () => {
     expect(kindToPlural('HorizontalPodAutoscaler')).toBe('horizontalpodautoscalers')
     expect(kindToPlural('pvc')).toBe('persistentvolumeclaims')
     expect(kindToPlural('PodGroup')).toBe('pods')
+  })
+
+  // A cold direct-URL load runs kindToPlural against the URL's plural slug one
+  // round-trip BEFORE initNavigationMap lands, so the discovered-plural guard
+  // can't fire for a CRD absent from BUILTIN_PLURAL_TO_KIND.
+  describe('CRD plurals before the discovery map arrives', () => {
+    test('idempotent on unknown lowercase plurals', () => {
+      expect(kindToPlural('schedules')).toBe('schedules')
+      expect(kindToPlural('validatingpolicies')).toBe('validatingpolicies')
+      expect(kindToPlural('virtualservices')).toBe('virtualservices')
+      expect(kindToPlural('clusters')).toBe('clusters')
+      expect(kindToPlural('backups')).toBe('backups')
+    })
+
+    // Plurals of `*se` singulars — Flux's HelmRelease and coordination.k8s.io's
+    // Lease are both kinds Radar handles, and both reach this cold path.
+    test('idempotent on plurals of *se singulars', () => {
+      expect(kindToPlural('helmreleases')).toBe('helmreleases')
+      expect(kindToPlural('leases')).toBe('leases')
+      expect(kindToPlural('databases')).toBe('databases')
+    })
+
+    test('still pluralizes singular PascalCase CRD kinds', () => {
+      expect(kindToPlural('Schedule')).toBe('schedules')
+      expect(kindToPlural('ValidatingPolicy')).toBe('validatingpolicies')
+      expect(kindToPlural('VirtualService')).toBe('virtualservices')
+      expect(kindToPlural('Cluster')).toBe('clusters')
+    })
+
+    // Lowercase does NOT imply plural. WorkloadViewRoute passes the URL segment
+    // through verbatim, so /workload/deployment/ns/name yields "deployment";
+    // pkg/topology's normalizeKind likewise returns its input unchanged when a
+    // kind resolves through neither its static map nor discovery, so a
+    // ResourceRef can carry "certificaterequest". Both must still pluralize —
+    // including the singulars that already end in 's'.
+    test('still pluralizes lowercase singular kinds', () => {
+      expect(kindToPlural('deployment')).toBe('deployments')
+      expect(kindToPlural('pod')).toBe('pods')
+      expect(kindToPlural('cronjob')).toBe('cronjobs')
+      expect(kindToPlural('certificaterequest')).toBe('certificaterequests')
+      expect(kindToPlural('validatingpolicy')).toBe('validatingpolicies')
+      expect(kindToPlural('ingress')).toBe('ingresses')
+      expect(kindToPlural('nodeclass')).toBe('nodeclasses')
+      expect(kindToPlural('ec2nodeclass')).toBe('ec2nodeclasses')
+      expect(kindToPlural('storageclass')).toBe('storageclasses')
+    })
+
+    test('leaves the empty-string result unchanged', () => {
+      expect(kindToPlural('')).toBe(englishPlural(''))
+    })
+
+    test('agrees with the post-discovery answer', () => {
+      const inputs = ['schedules', 'validatingpolicies', 'Schedule', 'ValidatingPolicy']
+      const cold = inputs.map(kindToPlural)
+      initNavigationMap([
+        { group: 'velero.io', version: 'v1', kind: 'Schedule', name: 'schedules', namespaced: true, isCrd: true, verbs: [] },
+        { group: 'kyverno.io', version: 'v1alpha1', kind: 'ValidatingPolicy', name: 'validatingpolicies', namespaced: false, isCrd: true, verbs: [] },
+      ])
+      expect(inputs.map(kindToPlural)).toEqual(cold)
+    })
   })
 })
 
--- a/packages/k8s-ui/src/utils/navigation.ts
+++ b/packages/k8s-ui/src/utils/navigation.ts
@@ -1,5 +1,5 @@
 import type { SelectedResource, ResourceRef, APIResource } from '../types/core'
-import { englishPlural } from './pluralize'
+import { englishPlural, englishSingular, isEnglishPlural } from './pluralize'
 
 /**
  * Canonical callback type for navigating to a resource.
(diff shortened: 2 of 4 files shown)


nadaverell:
### Cross-review (Codex, `019fd928`) — findings triaged

Codex reviewed this branch and flagged one **P1**, against the earlier snapshot `80540e2f` where the guard was just `if (kind === kindLower) return kindLower`:

> **Lowercase-singular direct URLs silently lose workload behavior.** `WorkloadViewRoute` takes the URL segment verbatim and passes it as `kind`, then the wrapper calls `kindToPlural(kindProp)`. For `/workload/deployment/default/api`, the old cold result was `deployments`; the new result is `deployment`. The primary fetch still succeeds because the server deliberately accepts both singular and plural, but pod loading is gated specifically on plural names, so `supportsWorkloadPods` becomes false and `useWorkloadPods` is disabled.

**Accepted — and already closed.** This is the same class I found independently (`pkg/topology`'s `normalizeKind` can hand back a lowercase singular `ResourceRef.Kind`), which is why the guard was tightened to `kind === kindLower && isEnglishPlural(kindLower)` before Codex reported. Codex notes its verdict applies to `80540e2f`, not the current commit.

`/workload/deployment` was a route I had **not** traced, so I added it to the test as a named case. Verified against the shipped guard with a cold map:

```
"deployment"         -> "deployments"       "ingress"            -> "ingresses"
"pod"                -> "pods"              "certificaterequest" -> "certificaterequests"
"cronjob"            -> "cronjobs"          ""                   -> "s"
```

Codex's other points:

- **§3 empty string** — no longer applies. `isEnglishPlural('')` is false, so `''` still returns `'s'` exactly as before. Pinned by a test asserting `kindToPlural('') === englishPlural('')`.
- **§4 branch placement** — confirmed correct; the authoritative lookups (builtin plurals, discovered map, aliases) all still run first.
- **§5 residual hypotheses** — the exhaustive call-site trace (81 production invocations across 36 files) found no *other* confirmed lowercase-singular producer. Its `parseNodeID` hypothesis is the path I'd already identified and is covered by the `certificaterequest` test case. The remaining two (raw nested `data: any` refs, apply-result Kinds) are unconfirmed and, being lowercase singulars, are handled correctly by the shipped predicate regardless.

Codex's suggested alternative — keep `kindToPlural` naive and disambiguate at each route boundary — was considered and not taken: `SelectedResource.kind` is polymorphic by design (plural URL slug from `/resources/:kind`, singular Kind from search hits and `ResourceRef`s) and reaches this function from 81 sites, so pushing the decision outward means auditing and guarding all of them rather than one.

nadaverell:
### Bugbot triage — 1 false positive, 1 real-but-unreachable

Both findings target `isEnglishPlural`. I ran them rather than reasoned about them.

---

#### 1. "Plural guard misses `-ses` resource names" — **false positive, no change**

The premise is right and the conclusion is wrong. `englishSingular` *does* over-strip a `*se` singular — but `englishPlural` inverts that exact error, so the round trip `isEnglishPlural` actually depends on still holds:

```
leases        -> englishSingular -> "leas"        -> englishPlural -> "leases"        -> true
helmreleases  -> englishSingular -> "helmreleas"  -> englishPlural -> "helmreleases"  -> true
```

`'leas'` ends in `s`, so `englishPlural` appends `es` — cancelling the `es` that was stripped. Only the round trip is contractual; the intermediate stem is not, and `isEnglishPlural` never exposes it.

Measured on this branch with a cold map:

| input | `englishSingular` | re-pluralized | `isEnglishPlural` | `kindToPlural` |
|---|---|---|---|---|
| `leases` | `leas` | `leases` | `true` | `leases` |
| `helmreleases` | `helmreleas` | `helmreleases` | `true` | `helmreleases` |
| `releases` | `releas` | `releases` | `true` | `releases` |
| `licenses` | `licens` | `licenses` | `true` | `licenses` |
| `databases` | `databas` | `databases` | `true` | `databases` |

No `helmreleaseses`. Both named kinds — Flux `HelmRelease` and `coordination.k8s.io/Lease` — are handled correctly today.

Since Bugbot named them specifically, they're now pinned by tests in both files so this doesn't get re-raised: `englishSingular` "over-strips *se stems but stays round-trip exact", `isEnglishPlural` "recognizes plurals of *se singulars", and `kindToPlural` "idempotent on plurals of *se singulars".

---

#### 2. "Singulars ending in `s` misdetected" — **real, but unreachable. Won't fix, pinned as a known limit**

This one is correct. `isEnglishPlural('status')` is `true`, so `kindToPlural('status')` returns `status` rather than `statuses`. `status`, `alias`, and `canvas` all reproduce.

It is not fixable by rule: `status → statu → status` and `clusters → cluster → clusters` are structurally identical. Distinguishing them needs a dictionary, not a suffix rule. So the question is whether any real input has the shape.

**Enumerated Radar's entire kind vocabulary** — 103 distinct singular PascalCase Kinds drawn from `pkg/topology`'s `normalizeKind` map values, the `NodeKind` union, `ResourceRendererDispatch`'s `KNOWN_KINDS`, and `BUILTIN_PLURAL_TO_KIND`'s values — and filtered to those that actually reach the guard (excluding the builtin plurals and aliases that short-circuit above it):

```
Kinds reaching the guard whose lowercase form is misread as plural:
  (none)
```

The two producers of lowercase singulars that Codex's trace identified are both clear:

- `/workload/:kind` is restricted to `FULLSCREEN_RESOURCE_KINDS`, so its singulars are `pod`, `deployment`, `statefulset`, `daemonset`, `job`, `cronjob`, `node` — verified all seven pluralize correctly.
- `pkg/topology`'s `normalizeKind` discovery-miss fallback emits from `buildNodeID`'s lowercase-singular vocabulary, which is inside the 103 above.

And the worst case is bounded even hypothetically. The discovered-map lookup runs *before* this guard, so a `Status`-shaped CRD self-heals the instant discovery lands:

```
cold  status -> status
warm  status -> statuses     (after initNavigationMap)
warm  Status -> statuses
```

That makes it a transient cold-window mistake — the same severity as the pre-existing RAD-330 bug this PR fixes, for a kind shape that doesn't exist in Radar's vocabulary. Trading it for a hand-maintained exception list would add a real maintenance burden to remove a hypothetical.

Pinned as an explicit `isEnglishPlural` test — "cannot distinguish singulars ending in a lone s" — with the reasoning in a comment, so the boundary is documented rather than merely unknown, and a future change to these rules surfaces it instead of moving it silently.

---

**CI note:** the `Backend` failure on this PR is `TestParameterNamingStaysConsistent` (`diagnose.inCluster` camelCase vs snake_case), a `main` breakage unrelated to this change — this PR touches only `packages/k8s-ui/src/utils/`. Fixed by `e4e950aa` on #1373; this branch goes green once that reaches `main`.

PR #1393, opened 2026-08-09
feat(scripts): kyverno-demo cluster for policy UI testing

nadaverell:
Scripts the Kyverno fixture set from #1371 so it survives the cluster it was built on. Follows `scripts/gitops-demo.sh` — same subcommand shape, same fixture-numbering convention, same status output.

`make kyverno-demo` · `make kyverno-demo-down` · `make kyverno-demo-status`

## Why script this one

Several of #1371's most load-bearing verifications came from cluster states that aren't reproducible by guesswork — a policy that declares `Deny` and blocks nothing, one with no `validationActions` at all, a report family that's served but empty. Without a script, testing policy UI means testing against whatever cluster you happen to be pointed at.

It's also the shape RAD-322 needs: admission-denial attribution requires live admission and real policies, which is what this bootstraps.

## What's in the fixtures

The set encodes the states that are hard to get right, not a minimal smoke test.

**The four enforcement-posture cases**, which are why Radar computes an effective posture instead of echoing `spec.validationActions`:

| Policy | Renders | Why |
|---|---|---|
| `require-run-as-nonroot` | `Deny` | Deny + admission on — genuinely blocks |
| `audit-require-labels` | `Audit + Warn` | Deliberate; declared and effective agree |
| `background-only-no-latest-tag` | `Background only` | **Declares Deny, blocks nothing** |
| `implicit-deny-configmap-labels` | `Deny` | **No `validationActions` at all** — Kyverno treats this as Deny, undocumented upstream |

Both surprising ones are verifiable by hand, and the README shows the commands. Confirmed on a fresh cluster from this script:

```
$ kubectl run probe --image=nginx:latest -n default
pod/probe created                                    # declares Deny, doesn't block

$ kubectl create configmap probe --from-literal=a=b -n policy-demo
admission webhook "vpol.validate.kyverno.svc-fail" denied the request:
  Policy implicit-deny-configmap-labels failed                # no validationActions, blocks
```

**Everything else worth a fixture:** a `MutatingPolicy` with `mutateExisting` and admission off (the case that used to render "Inactive" while rewriting the cluster — note its default is `false`, unlike admission/background); an `ImageValidatingPolicy` with cosign keyless attestors and SBOM/SLSA attestations; a `GeneratingPolicy` whose evaluation block is a different shape again; and namespaced twins.

**Both API families**, because family selection, the sidebar grouping and the legacy-vs-modern renderer split only have meaning during migration. `PolicyException` is the sharp case — same Kind, same plural, two groups, different specs — so there's one of each.

**Two DeletingPolicies on six-hourly and every-minute crons**, so within two minutes of bootstrap the table shows `Never run` beside a recent timestamp. That pairing is what the Last Run column exists to surface, and it happens on its own rather than needing a hand-patched status:

```
NAME                     LASTRUN
cleanup-completed-jobs   <none>
probe-every-minute       2026-08-09T08:16:09Z
```

**Real workloads, one passing and one failing**, so the background scanner produces ~40 reports across four distinct `results[].source` values — the engine-taxonomy case made visible:

```
37  KyvernoValidatingPolicy
29  KyvernoMutatingPolicy
25  kyverno
 7  KyvernoGeneratingPolicy
```

One engine, four producer strings.

## Two scenario subcommands

`openreports` reconfigures the chart so reports land in `openreports.io` while the `wgpolicyk8s.io` CRDs stay **served but empty** — the selection case where taking the first *served* group shows zero findings on a cluster full of them. Verified on this script's cluster: `wgpolicyk8s.io: 0`, `openreports.io: 37`.

`modern-only` removes the legacy CRDs to reproduce the Kyverno 1.20 API surface.

## The gotcha, stated up front

`modern-only` prints a warning before it runs and the README leads with it:

> **Kyverno 1.18.2's admission controller crashloops without the legacy CRDs.** It sanity-checks for `clusterpolicies.kyverno.io` and `policies.kyverno.io` at startup and exits when they're absent.

That's upstream behaviour, not a broken cluster. The reports controller stays healthy and the reports survive, which is what the detection gate needs. Anyone who hits this without warning will think they broke something.

## Testing

Ran `up` on a clean machine and it **failed** — which is why it's worth saying. `helm --wait` returns once the Deployments report ready, but the admission webhook isn't accepting connections yet, so the first policy apply died with `failed calling webhook "validate-policy.kyverno.svc": connection refused`.

Fixed with a readiness gate that waits for the Service endpoints and then round-trips a real policy through a server-side dry-run, because endpoint readiness alone still races the TLS cert. Fixture applies also retry, since the webhook can blip under load and a half-applied fixture set is worse than a slow one.

After the fix, verified end-to-end on a fresh cluster:

- `up` from nothing → all four controllers healthy, all fixtures applied, 40 reports
- `up` again on the existing cluster → idempotent, webhook ready first attempt
- `status` → full inventory including the source breakdown above
- `openreports` → 0 vs 37 as described
- Both hand-verification commands in the README behave as documented
- `bash -n` clean; `help` renders; unknown subcommand exits 1

**Not tested end-to-end: `modern-only`.** It intentionally breaks the cluster and costs a full reset to recover, and its behaviour was already established empirically during #1371 — that's where the crashloop and the surviving-reports observations come from. Flagging rather than implying full coverage.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to local dev scripts, fixtures, and documentation; no application runtime or auth paths are modified.
> 
> **Overview**
> Adds a **Kyverno demo bootstrap** (same shape as `gitops-demo`) so policy UI work can run against a predictable **kind** cluster instead of whatever kube context is active.
> 
> **`scripts/kyverno-demo.sh`** provisions Kyverno 1.18.2 (pinned chart), applies numbered fixtures under `scripts/kyverno-demo/`, waits for PolicyReports, and prints inventory via `status`. It includes a **webhook readiness gate** (endpoints + server-side dry-run policy) and **retried fixture applies** so `up` does not fail when Helm reports ready before admission is actually accepting traffic.
> 
> **Makefile** exposes `kyverno-demo`, `kyverno-demo-down`, and `kyverno-demo-status`. The fixture set targets Radar policy rendering: the four **effective enforcement-posture** cases, modern policy kinds (image/mutate/generate/delete + namespaced twins), **legacy + modern** APIs including dual-shape **PolicyException**, workloads that seed ~40 reports, plus scenario toggles **`openreports`** (served-but-empty `wgpolicyk8s.io` vs populated `openreports.io`) and **`modern-only`** (legacy CRD removal / 1.20 surface simulation, with documented admission-controller crashloop caveat). **`scripts/kyverno-demo/README.md`** documents the coverage matrix and hand-verification commands.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff0835d7292901843770809cbb54e07bbe171188. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,6 +174,21 @@ gitops-demo-status:
 gitops-demo-drift:
 	./scripts/gitops-demo.sh drift
 
+# Bootstrap a kind cluster pre-loaded with Kyverno 1.18.2 and curated policy
+# fixtures (both API families, the four enforcement-posture cases, dual-API
+# PolicyExceptions, and ~40 real PolicyReports across several sources).
+# Useful for visual-testing policy UI changes against realistic state.
+# See scripts/kyverno-demo/README.md for the coverage matrix — and read the
+# 1.20-simulation warning before running `kyverno-demo.sh modern-only`.
+kyverno-demo:
+	./scripts/kyverno-demo.sh up
+
+kyverno-demo-down:
+	./scripts/kyverno-demo.sh down
+
+kyverno-demo-status:
+	./scripts/kyverno-demo.sh status
+
 # Bootstrap a kind cluster pre-loaded with curated Crossplane fixtures
 # (core + provider-kubernetes + function-patch-and-transform + XRD/Composition/XRs).
 # Useful for visual-testing Crossplane UI changes against realistic state.
--- a/scripts/kyverno-demo.sh
+++ b/scripts/kyverno-demo.sh
@@ -1,0 +1,359 @@
+#!/usr/bin/env bash
+# Bootstrap a kind cluster pre-populated with a curated set of Kyverno
+# scenarios for visual-testing the policy UI. Idempotent — can be re-run to
+# refresh state or apply fixture updates without recreating the cluster.
+#
+# Subcommands:
+#   up            Create cluster (if missing), install Kyverno, apply fixtures.
+#   down          Delete the kind cluster.
+#   reset         down + up.
+#   status        Inventory the policies, reports and report API groups.
+#   openreports   Switch report output to openreports.io, leaving the
+#                 wgpolicyk8s.io CRDs served but EMPTY. This is the
+#                 report-family selection case: a naive "first served wins"
+#                 picks the empty API and shows zero findings on a cluster
+#                 full of them.
+#   modern-only   Remove the legacy kyverno.io policy CRDs to reproduce the
+#                 Kyverno 1.20 API surface. READ THE WARNING IT PRINTS.
+#   help          Show this message.
+#
+# Prerequisites:
+#   - kind         https://kind.sigs.k8s.io/
+#   - kubectl
+#   - helm
+#
+# Set CLUSTER_NAME=foo to use a different cluster (default: radar-kyverno-demo).
+#
+# See scripts/kyverno-demo/README.md for the coverage matrix and the
+# 1.20-simulation gotcha.
+
+set -euo pipefail
+
+CLUSTER_NAME="${CLUSTER_NAME:-radar-kyverno-demo}"
+KUBECTL_CTX="kind-${CLUSTER_NAME}"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+FIXTURES_DIR="${SCRIPT_DIR}/kyverno-demo"
+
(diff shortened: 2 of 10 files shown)


nadaverell:
Superseded by #1402, which carries the Kyverno demo plus the context-safety, scenario assertions, review fixes, and end-to-end verification. Keeping this branch intact for history.

PR #1435, opened 2026-08-14
feat(ai): drive GitHub Copilot CLI for investigations

rafaelherik:
## Description

Adds GitHub Copilot CLI (`copilot`) as a fourth investigations backend alongside
Claude Code, Codex, and Cursor, at full parity: both execution profiles, model and
reasoning-effort control, session resume, and the "continue in your terminal" hand-off.

Safeguarded is the tightest containment of the four backends, because Copilot can
filter the model's entire tool surface: `--available-tools` leaves only Radar's MCP
tools visible (no shell, no file edits, no web access), `--allow-tool=radar` approves
them without `--allow-all-tools`, and the built-in GitHub MCP server plus every server
in the user's own config is disabled by name. Copilot has no single "ignore user
config" flag, so that set is enumerated with a cached `copilot mcp list` probe; an
unreadable probe is an error rather than an empty list, so "we couldn't tell" never
silently downgrades the profile.

Both profiles pass `--no-remote-export`. Copilot otherwise publishes the session to
GitHub's web and mobile surfaces, and these transcripts carry cluster data.

**Guards a silent failure mode.** When Radar's MCP server doesn't attach — an org
policy disabling third-party MCP, or a connect failure — Copilot still exits 0 with
empty stderr and the model still answers, having read nothing from the cluster. The
verdict parses cleanly, so the existing `structured()` check would have let it through.
`Diagnosis.mcpErrText` carries that signal and fails the turn before that check: a
well-formed answer is what makes the state dangerous, not what excuses it.

Reasoning-effort validation becomes per-agent (`ai.ReasoningEfforts`): Copilot accepts
`none..max` where Codex accepts `minimal..high`, so one agent's vocabulary can no
longer be handed to another.

## Type of change

- [ ] Documentation update

### Kind Cluster

- Init a kind cluster and installed grafana stack using a wrong storage class
- Started the first investigation session with Radar safeguards
- Started the second investigation with my personal configuration

### Remote Cluster

- Deployed AKS Cluster
- Started Radar Local instance
- Same Grafana Stack Scenario but against different cluster

PS: The test was focused on the application behavior, UI behavior, communication with local agent and safeguards preventing the model to skip safeguards and try to run dangerous commands. The investigation produced a similar result compared with a Claude Code Session. 

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds a new agent execution path with strong safeguarded containment and MCP-attach validation, but mistakes here could allow false “healthy” verdicts without cluster access or leak investigation data to GitHub export surfaces.
> 
> **Overview**
> Adds **GitHub Copilot CLI** (`copilot`) as a fourth investigations backend, wired through detection, routing, consent, API validation, CLI help text, and the diagnose UI (catalog, safeguards copy, reasoning-effort menus, session hand-off).
> 
> The new **`copilotAgent`** runs headless `copilot -p` with JSON streaming. In **safeguarded** mode it restricts the model to Radar MCP tools (`--available-tools`, `--allow-tool=radar`), redirects **`COPILOT_HOME`** to `~/.radar/copilot-home`, runs in an empty temp cwd, disables built-in MCPs and custom instructions, and always passes **`--no-remote` / `--no-remote-export`** so cluster transcripts are not published to GitHub.
> 
> **MCP attach failures** are treated as turn failures: Copilot can exit 0 with a plausible verdict when Radar’s MCP never connected. **`mcpErrText`** on `Diagnosis` captures that signal and **`DiagnoseStream`** fails the turn before accepting a structured verdict.
> 
> Reasoning effort is now **per-agent** (`ReasoningEfforts` / `SupportsEffort`); the API validates effort against the selected agent (Copilot’s wider `none…max` range vs Codex).
> 
> Agent env scrubbing is centralized in **`minimalEnv` / `filterEnv`** (platform base keys on Windows), reused by Codex, Copilot, and Claude’s scrubbed path.
> 
> Minor dev tweak: `.air.toml` uses `entrypoint` instead of `bin` and drops the default kubeconfig flag from the live-reload command.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d47b4f852bb5adcc84b5cdc30d0edcce4bff1d25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/ai/agent_copilot.go
+++ b/internal/ai/agent_copilot.go
@@ -10,10 +10,8 @@ import (
 	"log"
 	"os"
 	"os/exec"
-	"sort"
+	"path/filepath"
 	"strings"
-	"sync"
-	"time"
 )
 
 // copilotMCPServer is the name Radar's MCP server is registered under. Copilot
@@ -30,9 +28,13 @@ const copilotMCPServer = "radar"
 //     once the tools are gone);
 //   - --allow-tool <server> approves those tools without --allow-all-tools, so
 //     nothing else is auto-approved;
-//   - --disable-builtin-mcps drops github-mcp-server, and every server in the
-//     user's own config is disabled by name (see resolveUserMCPServers) — Copilot
-//     has no single "ignore user config" flag, so the set is enumerated;
+//   - COPILOT_HOME points at a Radar-owned directory, so NOTHING of the user's own
+//     Copilot setup loads: not their MCP servers ($COPILOT_HOME/mcp-config.json),
+//     not their settings, and — decisively — not their user-level hooks, which run
+//     shell in -p mode and which no flag can disable (the documented
+//     disableAllHooks setting does not take effect from a workspace in 1.0.80);
+//   - --disable-builtin-mcps drops github-mcp-server, which is built in rather
+//     than user config and so survives the COPILOT_HOME redirect;
 //   - cluster WRITE access is gated by the read-only MCP MOUNT, the same
 //     server-side gate the other backends rely on.
 //
@@ -41,9 +43,9 @@ const copilotMCPServer = "radar"
 type copilotAgent struct {
 	bin string
 
-	serversMu    sync.Mutex
-	serversKnown bool
-	userServers  []string
+	// homeDir resolves the user's home. Overridden in tests so a safeguarded run
+	// never writes into the developer's real ~/.radar.
+	homeDir func() (string, error)
 }
 
 func (a *copilotAgent) Name() string { return "copilot" }
@@ -89,6 +91,10 @@ func (a *copilotAgent) command(ctx context.Context, s turnSpec) (*exec.Cmd, func
 		// The transcript contains cluster data — never publish it to GitHub's web
 		// and mobile surfaces, in either profile.
 		"--no-remote", "--no-remote-export",
+		// Copilot self-updates by default and its CI-environment detection can't
+		// fire under a minimized env: pin the binary for the duration of the run
+		// rather than let it swap itself mid-investigation.
+		"--no-auto-update",
 	}
 
 	if s.profile == ExecutionProfileSafeguarded {
@@ -100,13 +106,6 @@ func (a *copilotAgent) command(ctx context.Context, s turnSpec) (*exec.Cmd, func
 			"--disable-builtin-mcps",
 			"--no-custom-instructions",
 		)
-		servers, err := a.resolveUserMCPServers(ctx)
-		if err != nil {
-			return nil, nil, err
-		}
-		for _, name := range servers {
-			args = append(args, "--disable-mcp-server", name)
-		}
 	} else {
 		// Full-local: the user's own tools, MCP servers and instructions are live.
 		// Cluster writes stay gated by the read-only MCP mount.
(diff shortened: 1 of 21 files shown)


internal/ai/agent_copilot.go
```diff
@@ -0,0 +1,416 @@
...
+	if s.model != "" {
+		args = append(args, "--model", s.model)
+	}
+	// Default to medium rather than Copilot's bare default, matching the other
+	// backends: it gives a solid investigation without an unbounded token bill.
+	effort := s.effort
+	if effort == "" {
+		effort = "medium"
+	}
+	args = append(args, "--effort", effort)
```
nadaverell: Breaks on accounts whose default model is `auto` (see review) — only pass effort when chosen. `TestCopilotEffortDefault` + the "Default (medium)" menu copy need the matching update.
```suggestion
	if s.effort != "" {
		args = append(args, "--effort", s.effort)
	}
```

internal/ai/agent_copilot.go
```diff
@@ -0,0 +1,416 @@
...
+		// times out. The agent can still ask in prose, which is what Radar wants.
+		"--no-ask-user",
+		// Reasoning summaries are off by default; without them the stream carries
+		// only tool calls and the final message, and the UI shows no thinking.
+		"--enable-reasoning-summaries",
+		// Keep diagnostics out of the JSONL on stdout.
+		"--log-level", "none",
+		// The transcript contains cluster data — never publish it to GitHub's web
+		// and mobile surfaces, in either profile.
+		"--no-remote", "--no-remote-export",
```
nadaverell, line 93: Copilot self-updates by default and the CI env detection won't trigger under the minimized env — pin the version for the run.
```suggestion
		"--no-remote", "--no-remote-export",
		"--no-auto-update",
```

web/src/components/diagnose/parts.tsx
```diff
@@ -818,13 +834,29 @@ export function ConsentCard({
...
                   Review the agent&apos;s documented restrictions before
                   continuing.
                 </>
               ),
+              ...(agent === "copilot"
+                ? [
+                    <>
+                      The session is never exported to GitHub&apos;s web or
+                      mobile surfaces, so the transcript and the cluster data in
+                      it stay on this machine.
```
nadaverell: "stay on this machine" overstates it — the conversation still goes to the model provider. The flag only kills the GitHub web/mobile surface.
```suggestion
                      The session is never exported to GitHub&apos;s web or
                      mobile surfaces; it exists only here and in your local
                      Copilot session store.
```

internal/ai/agent_copilot.go
```diff
@@ -0,0 +1,416 @@
...
+
+// resolveUserMCPServers lists the MCP servers Copilot would otherwise load and
+// returns every one that isn't Radar's, so a safeguarded turn can disable them by
+// name. Copilot has no --strict-mcp-config / --ignore-user-config equivalent, so
+// the set has to be enumerated. Probed once and cached, like the Cursor trust flag.
+//
+// An unreadable or timed-out probe is an ERROR, not an empty list: silently
+// treating "we couldn't tell" as "the user has no servers" would quietly downgrade
+// safeguarded to a run with the user's other servers attached.
+func (a *copilotAgent) resolveUserMCPServers(ctx context.Context) ([]string, error) {
```
nadaverell: With the `COPILOT_HOME` redirect (see review) this whole probe can go away.

nadaverell (changes requested):
thanks - great PR!
 
a few comments to lock it down properly and integrate cleanly with copilot supported options and flags. tested with copilot 1.0.80, most things check out, hit two issues:

(note - partly AI assisted review but driven and checked by me)

**1. User hooks still run in safeguarded mode.** Hooks from `~/.copilot/hooks/` fire in `-p` mode (verified with a sentinel `preToolUse` hook - it ran shell from a safeguarded-shaped invocation). No flag disables them, and the documented `disableAllHooks` in workdir `.github/copilot/settings.json` doesn't work in v1.0.80 (tested). What does work: pointing `COPILOT_HOME` at a Radar-owned dir for safeguarded runs. Auth (via `GH_TOKEN`) and `--resume` both work under the redirect - and it makes the PR simpler: the user's MCP servers never load, so `resolveUserMCPServers` and the `--disable-mcp-server` enumeration can be deleted. `launch.ts` would prefix `COPILOT_HOME=…` on the hand-off command. One thing to confirm on your setup, since I only tested env-token auth: that a keychain `copilot login` survives the redirect.

**2. `--effort medium` breaks on `auto`-model accounts.** `copilot -p "hi" --effort medium` → `Error: Model "auto" does not support reasoning effort configuration` (exit 1). Free tier defaults to `auto`, so every default-settings run fails there. Fix inline: only pass `--effort` when the user picked one.

Added also a few smaller comments as suggestions, take it or leave it - happy to also push followups if you prefer. Either way really appreciate the PR!

nadaverell:
Hey @rafaelherik would you like to pick this up and finish the implementation? I can take it from here if you prefer, on top of your PR.

PR #1451, opened 2026-08-18
fix(nodes): parse pod capacity as a Quantity, not a raw integer (#1441)

### LINKED ISSUE #1441: Nodes list parses pod capacity `1k` as 1, showing 216/1 = 21600%
## Describe the bug

Node pod capacity is read as a raw number, but `status.capacity.pods` and `status.allocatable.pods` are `resource.Quantity` values, whose **canonical serialization collapses trailing zeros into SI suffixes**. A node with `--max-pods=1000` reports `"1k"`, not `"1000"`.

The Nodes list appears to parse that as `1`, so a node running 216 pods renders as **`216 / 1` — `21600%`** with the usage bar pegged.

Any node whose max-pods is a multiple of 1000 hits this. It is purely a display defect (the scheduler and API are unaffected), but the percentage is alarming enough to look like a real capacity emergency, which is how we found it mid-maintenance-window.

## To reproduce

1. Have a node with a pod capacity that canonicalizes to an SI suffix — e.g. kubelet `--max-pods=1000`
2. Confirm what the API returns:
   ```console
   $ kubectl get node <node> -o jsonpath='{.status.capacity.pods}'
   1k
   ```
3. Open the **Nodes** list view in Radar
4. The PODS column shows `216 / 1` and `21600%`

## Expected behavior

Pod capacity parsed as a Quantity, so the node shows `216 / 1000` and `21.6%`.

## Screenshots

Nodes list on a k3s node with `--max-pods=1000`:

> `gemi-coder` — Ready — control-plane — CPU `11148m / 3200…` 35% — MEMORY `115.4Gi / 235…` 49% — **PODS `216 / 1`  21600%**

CPU and memory in the same row are correct; only PODS is wrong.

## Confirmation that the value is genuinely 1000

```console
$ kubectl get node gemi-coder -o json | jq -c '{capacityPods: .status.capacity.pods, allocatablePods: .status.allocatable.pods}'
{"capacityPods":"1k","allocatablePods":"1k"}

$ kubectl get pods -A --field-selector spec.nodeName=gemi-coder --no-headers | wc -l
217
```

217 pods are scheduled and running, which would be impossible against a capacity of 1.

## Likely root cause

`resource.Quantity` canonicalization applies SI suffixes in groups of three zeros, so the surfaces that need suffix-aware parsing are:

| max-pods | serialized as |
|---|---|
| 1000 | `1k` |
| 2000 | `2k` |
| 1020 | `1020` |
| 990 | `990` |

Pointers from a quick read of the tree (not exhaustive — I did not find the Nodes-list component that computes the percentage):

- `packages/k8s-ui/src/utils/format.ts` has `parseMemoryToBytes()` doing correct binary/decimal suffix handling and `parseCPUToNanocores()` handling `n`/`m`, but there is **no equivalent parser for plain counts**, so a `Quantity` like `1k` has nothing suffix-aware to go through.
- `packages/k8s-ui/src/components/resources/renderers/NodeRenderer.tsx:140-141` passes `capacity.pods` / `allocatable.pods` straight through unparsed. The detail view renders the string as-is (displays `1k`, which is at least not *wrong*); the list view's percentage math is where it becomes `1`.

The same class of bug would affect any other integer-count Quantity field rendered numerically — e.g. extended resources and device-plugin counts (`nvidia.com/gpu: 1k` is unlikely, but attachable-volumes and custom resources can exceed 999).

A shared `parseQuantityToNumber()` in `format.ts`, used everywhere a Quantity feeds arithmetic, seems like the tidy fix.

## Environment

- Kubernetes: `v1.36.3+k3s1` (k3s, single node, bare metal)
- Node: 36 CPU / ~251 GiB, kubelet `--max-pods=1000`, podCIDR `/22`
- Radar version: happy to add this plus a diagnostics snapshot if useful — filing from the CLI, so I don't have the in-app snapshot to hand.

css521:
Fixes #1441.

## The bug

`ResourcesView.tsx` read allocatable pods with `parseInt`:

```ts
const max = parseInt(allocatable, 10)   // parseInt("1k", 10) === 1
```

A `resource.Quantity`'s canonical serialization collapses trailing zeros into an SI suffix, so a node started with `--max-pods=1000` reports `"1k"`, not `"1000"`. That made a node running 216 pods render as `216 / 1` with the bar pegged at **21600%**. CPU and memory in the same row were correct, which is what made it read like a real capacity emergency rather than a display defect.

Any count that is a multiple of 1000 hits it. `1020` and `990` do not, which is why this survived: the value has to be exactly divisible by 1000 before the API abbreviates it.

## The fix

The reporter suggested a shared `parseQuantityToNumber()` in `format.ts`, and the tree already agreed with them in two places — this was the third surface, and the only one with no parser at all:

| surface | had | 
| --- | --- |
| memory | `parseMemoryToBytes` (Ki…Ti, k…T) |
| extended resources / GPU counts | private `asCount`, whose comment already said *"3k is 3000"* |
| **pod counts in the Nodes list** | **`parseInt`** |

So `parseQuantityToNumber` is added to `format.ts`, and both the pods cell and `asCount` now go through it.

The parser covers every suffix form the API can emit rather than the ones memory happens to use:

- decimal SI including the sub-unit range — `n`, `u`, `m`, `k`, `M`, `G`, `T`, `P`, `E`
- binary SI up to `Ei`
- the decimal-exponent form

`"1E3"` is 1000 and `"1E"` is exa, so the exponent form is decided *before* the suffix table — otherwise `1E3` would be read as exa. An unrecognised suffix yields `0` rather than the bare number, which is deliberate and differs from `parseMemoryToBytes`: the pods cell already falls back to showing the count alone when it gets a non-positive max, so returning 0 makes an unreadable Quantity show as "unknown" instead of drawing a confidently wrong percentage.

`asCount` previously borrowed `parseMemoryToBytes`. That covered the suffixes an extended resource realistically carries, but it was the wrong contract to borrow, and it left the same class of bug live wherever a count had no parser. It now shares the count parser and picks up `P`/`E`/`Pi`/`Ei` and exponents as a side effect.

I did **not** touch the detail view (`NodeRenderer.tsx:140-141`). It renders the Quantity string as-is, so it already shows `1k` — as the issue says, at least not wrong.

## Verification

The reported case, as a test:

```ts
expect(parseQuantityToNumber('1k')).toBe(1000)
expect((216 / parseQuantityToNumber('1k')) * 100).toBeCloseTo(21.6)
```

which is the `216 / 1000` and `21.6%` the issue asks for. Nine test cases in total cover the reported bug, the near-misses that must stay unchanged (`1020`, `990`), each suffix family, the exponent/exa ambiguity, signs, unparseable input, and already-decoded numbers.

```
packages/k8s-ui   npm test          131 files, 2323 passed, 1 skipped
web               npm run tsc       0 errors
web               npm run lint      0 errors (302 pre-existing warnings)
web               npm run build     ok
```

Three things I checked rather than assumed, in case they save you time:

- **`packages/k8s-ui` `npm run tsc` reports 7 errors, all pre-existing.** They are in `src/components/trace/reachGraphModel.test.ts` and `reachInspector.test.ts`, which this PR does not touch, and the identical 7 appear on an unmodified checkout. CI does not run `tsc` for the `k8s-ui` package (only for `web`, which is clean), so they are not newly exposed either.
- **`web`'s `client.authRedirect.test.ts` is flaky under load, independently of this change.** "navigates to /auth/login exactly once for a burst of concurrent 401s" times out at exactly the 5000 ms default. Across full-suite runs I measured 2 failures in 5 on an unmodified checkout and 2 in 3 with this change; it passes every time when run alone. The test mocks `./config` and imports only `./client`, so it has no import path to any module here. My machine's load average was 12/31/40, which is the likeliest driver. Flagging it rather than filing it, since you may already know.
- **Prettier is not clean on these files, and was not before this change either.** All four fail `prettier --check` on an unmodified checkout, and no workflow references prettier, so I matched the surrounding style by hand rather than running `--write` — that would have reformatted whole files and buried a 160-line change in reformatting noise. Happy to run it if you would rather have it.

One thing I could not verify: I have no cluster with `--max-pods=1000`, so the fix is tested at the parser and not against a live node. The reporter's `kubectl` output (`{"capacityPods":"1k","allocatablePods":"1k"}` with 217 pods scheduled) is what the test encodes.

---

> [!NOTE]
> **Low Risk**
> Display-only quantity parsing for node/GPU counts; no API, auth, or data-path changes. Unrecognized suffixes now yield 0 instead of a guessed number.
> 
> **Overview**
> Fixes the Nodes list pods usage bar treating allocatable `"1k"` as `1` (`parseInt`), which pegged a 216-pod node at 21600%.
> 
> Adds shared `parseQuantityToNumber` in `format.ts` (SI, binary, millicounts, exponent form) and uses it for the pods cell and GPU/`asCount` helpers. Unparseable suffixes return 0 so the bar falls back instead of showing a wrong percent. Tests cover `"1k"` → 1000 and the nearby counts that must stay unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3efe40e7c76399fac1b8e229104c2907396b4d29. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/k8s-ui/src/components/resources/ResourcesView.tsx
+++ b/packages/k8s-ui/src/components/resources/ResourcesView.tsx
@@ -137,6 +137,7 @@ import {
 import { SEVERITY_BADGE, EVENT_TYPE_COLORS } from '../../utils/badge-colors'
 import { pluralize } from '../../utils/pluralize'
 import { getPodGpuCount, getNodeGpuCount } from '../../utils/extended-resources'
+import { parseQuantityToNumber } from '../../utils/format'
 import { type CustomColumnDef, type CustomColumnSource, customColumnKey, readCustomColumnValue, sanitizeCustomColumnDefs } from '../../utils/custom-columns'
 import { FreshnessControl, type FreshnessConnection } from '../ui/FreshnessControl'
 import { Tooltip } from '../ui/Tooltip'
@@ -7335,8 +7336,10 @@ function NodeCell({ resource, column, majorityNodeMinorVersion }: { resource: an
       const allocatable = resource.status?.allocatable?.pods
       const podCount = m?.podCount ?? 0
       if (!allocatable) return <span className="text-sm text-theme-text-tertiary font-mono">{podCount || '-'}</span>
-      const max = parseInt(allocatable, 10)
-      if (isNaN(max) || max <= 0) return <span className="text-sm text-theme-text-tertiary font-mono">{podCount || '-'}</span>
+      // Allocatable pods is a Quantity, so a 1000-pod node reports "1k" —
+      // parseInt read that as 1 and pegged the bar at 21600% (#1441).
+      const max = parseQuantityToNumber(allocatable)
+      if (!Number.isFinite(max) || max <= 0) return <span className="text-sm text-theme-text-tertiary font-mono">{podCount || '-'}</span>
       const pct = (podCount / max) * 100
       return <ResourceBar used={String(podCount)} total={String(max)} percent={pct} colorScheme="count" />
     }
--- a/packages/k8s-ui/src/components/resources/ResourcesView.tsx
+++ b/packages/k8s-ui/src/components/resources/ResourcesView.tsx
@@ -7337,7 +7337,7 @@ function NodeCell({ resource, column, majorityNodeMinorVersion }: { resource: an
       const podCount = m?.podCount ?? 0
       if (!allocatable) return <span className="text-sm text-theme-text-tertiary font-mono">{podCount || '-'}</span>
       // Allocatable pods is a Quantity, so a 1000-pod node reports "1k" —
-      // parseInt read that as 1 and pegged the bar at 21600% (#1441).
+      // a raw-integer read would see 1 and peg the bar at 21600%.
       const max = parseQuantityToNumber(allocatable)
       if (!Number.isFinite(max) || max <= 0) return <span className="text-sm text-theme-text-tertiary font-mono">{podCount || '-'}</span>
       const pct = (podCount / max) * 100
--- a/packages/k8s-ui/src/utils/extended-resources.ts
+++ b/packages/k8s-ui/src/utils/extended-resources.ts
@@ -2,7 +2,7 @@
 // the GPU views. Keep all effective-request math here — K8s semantics for
 // extended resources are subtle and must not be reimplemented per call site.
 
-import { parseMemoryToBytes } from './format'
+import { parseQuantityToNumber } from './format'
 
 const STANDARD_NODE_RESOURCES = new Set(['cpu', 'memory', 'pods', 'ephemeral-storage'])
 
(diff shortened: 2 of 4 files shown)


nadaverell (approved):
LGTM, thanks — nice writeup and good test coverage.

Pushed a small commit trimming the comments to match our convention (no ticket refs / change history in code comments — the PR body already covers it).

Review also turned up the same bug in the namespace quota bars; we'll fix that separately.

PR #1452, opened 2026-08-18
docs(reachability): trim the vantage paragraph, add a security model

roylibman:
Acts on Eyal's review of the reachability docs.

**1. The vantage paragraph was too much.** Five dense sentences doing four jobs. Now two: probes run from wherever Radar runs, and a failure attributable to the vantage leaves the verdict at `unknown` rather than condemning the path. Everything else it carried is already covered by the **Limits** section below, so it points there instead of restating it. Drops the "fail-toward-silence" coinage — internal vocabulary.

**2. Dig into the security benefits.** The page had one RBAC line. New `## Security model` section covers:

- Probes run as the impersonated user; if impersonation can't be established the probe fails rather than falling back to Radar's ServiceAccount
- The two separate permission gates (API-server proxy vs in-cluster Job)
- Turning the feature off is an RBAC change — no app-level flag to bypass
- What the probe pod is: **no ServiceAccount token mounted**, non-root, read-only rootfs, all capabilities dropped, 25s deadline, self-deleting, max 5 per run

`### What an agent gets` answers Eyal's actual question — an agent gets typed tools with fixed schemas, no kubeconfig and no shell, under the driver's identity, versus handing it credentials where the surface is every verb on every resource.

Claims were verified against the implementation rather than restated from the existing docs. That corrected the in-cluster verb list to `create jobs` + `list pods` + `get pods/log`, and surfaced the no-token-mounted detail.

The same edit is mirrored in skyhook-dev/radar-docs, which this file is hand-kept in sync with.

**Worth a look while reviewing:** `diagnose` is registered on `/mcp-readonly` too, and `inCluster: true` there still creates probe pods. RBAC and consent gate it, so it's defensible — but "read-only endpoint that can create pods" is a sentence someone will quote back at us. The docs avoid implying otherwise; flagging in case it deserves a change.

---

> [!NOTE]
> **Low Risk**
> Documentation-only edits to reachability.md; no runtime or RBAC behavior changes.
> 
> **Overview**
> **Reachability docs** are tightened at the top and expanded on security.
> 
> The opening **vantage** explanation is shortened from a long fail-toward-silence paragraph to two sentences that defer edge cases to **Limits**, and the standalone RBAC bullet under active probes is removed in favor of the new section.
> 
> A new **`## Security model`** documents impersonation (no fallback to Radar’s ServiceAccount), separate RBAC for API-server proxy vs in-cluster Jobs (`create jobs`, `list pods`, `get pods/log`), consent and SSAR gating, disabling probes via RBAC only, and probe pod hardening (no SA token, restricted PSS, caps, TTL). **`### What an agent gets`** contrasts typed MCP tools under the driver’s identity with kubeconfig/shell access and describes `/mcp-readonly` as not registering mutating tools.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit abb2fbc6ce4d249e34be99c6c6cf2bddd2cc3d7c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/reachability.md
+++ b/docs/reachability.md
@@ -15,7 +15,7 @@ The trace has two layers:
 
 The active layer can escalate the static verdict when probes give clear evidence of a real failure on a hop (every non-skipped probe failed → that hop counts toward broken; over half failed → counts toward degraded). It never softens a static verdict: a critical static finding outranks probe state, and an unverifiable path stays unverifiable.
 
-Probes run from the operator's current vantage (laptop or in-cluster). A failure caused by the vantage itself - NetworkPolicy blocking Radar's path, a service-mesh mTLS handshake without a client cert, an API-server-proxy probe that can't reach an internal address - is **never** allowed to set the headline. Only a real-traffic (data-path) result escalates the verdict; an API-server-proxy-only or otherwise vantage-attributable failure annotates the hop and localizes the symptom but leaves the verdict at *unknown* / *not confirmed* rather than condemning a path that real traffic may well reach. This is the fail-toward-silence rule: Radar would rather say "couldn't confirm from here" than false-condemn a healthy path. See "What it deliberately does NOT do" below for the boundaries of what probes can and cannot tell you.
+Probes run from wherever Radar is running - your laptop, or inside the cluster - and where they run from changes what they can prove. When a probe fails for a reason that belongs to the vantage rather than the workload, Radar marks the hop and leaves the verdict at *unknown* instead of reporting the path as broken; only a result from the real traffic path sets the headline. The **Limits** section below lists the cases where this happens and what to do about each one.
 
 ![A Service reached by both an Ingress and an HTTPRoute: parallel entry points on the left, the path graph, and the journey inspector on the right](images/reachability/reachability-parallel-entries.png)
 
@@ -89,9 +89,6 @@ Each row reports outcome (`ok` / `fail` / `skipped`), latency, the path it trave
 
 Every claim on the page carries its evidence: the verdict band states the live-check volume ("8 live checks from 2 vantages", with the DNS/TCP/TLS/HTTP breakdown on hover), a skipped route states WHY it was skipped from the exact vantage that skipped it, and node dots show each resource's own health - cluster state, never a test result (edges carry the test truth).
 
-**RBAC for active probes (from a laptop):** the user identity reading the trace must hold `get services/proxy` and `get pods/proxy` in the target namespace. In-cluster Radar uses the data path directly and doesn't need these. To disable the active layer entirely, deny those permissions to the role Radar uses.
-
-
 ![A Redis Service after an in-cluster TCP test: the probe capsule reports what it saw, and the coverage line states that only the transport was checked](images/reachability/reachability-non-http-service.png)
 
 *A non-HTTP Service (Redis on 6379) after an in-cluster test: the throwaway Pod connected over TCP, and the coverage line states the ceiling of that proof - "TCP connections only - application protocol not checked".*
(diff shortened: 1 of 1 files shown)


nadaverell (changes requested):
Claudex-led review for your consideration. I think Eyal should have the say on this ultimately.

=======================

The Security model section is the right addition — this page needed exactly it, and the probe-pod hardening list and the verified verb triple are the strongest parts. Requesting changes only because this doc is the reference readers will hold us to, and a handful of statements need to line up precisely with what ships today. Each one is a small wording fix.

1. **Scope the `/mcp-readonly` sentence.** "the mutating tools are never registered: … they don't exist on that endpoint" — as the PR description notes, `diagnose` is registered on the read-only mount too (deliberately: `internal/mcp/tools.go`, pinned by `tools_catalog_test.go`), and with `in_cluster: true` it creates the probe Job this same page describes. Suggest naming the write tools that are excluded (`apply_resource`, `patch_resource`, and the `manage_*` family — workload, rollout, cronjob, gitops, node) and stating plainly that `diagnose` remains available there, with the capped, credential-less probe Job as its one create. That's a stronger sentence than the absolute, because it's the one we can stand behind word-for-word.

2. **The consent sentence promises more than the mechanism.** "Nothing is created until someone clicks" — the dialog has a per-cluster don't-ask-again (persisted in localStorage), and the MCP path (`diagnose` with `in_cluster: true`) runs with no dialog at all; its gates are the explicit argument, the SSAR preflight, and the Cloud role. Suggest: in the UI, the first run asks and the dialog names the exact requests; via MCP, the explicit `in_cluster` argument plus the same RBAC preflight are the gate.

3. **Two identities, not one.** "A probe can only reach what that person could already reach with kubectl" is true for *creating* the Job, but the pod's network identity is the target namespace's default ServiceAccount — NetworkPolicy and mesh policy judge the pod, not the requesting user. The codebase already states this well (`internal/reachability/incluster.go`: "The throwaway probe pod has a DIFFERENT identity than the real client"; the consent dialog says the pod runs "not as you"). One added sentence here keeps the doc aligned with both.

4. **"A denial is reported as a denial … never scored as an unreachable service" doesn't match current rendering.** A proxy 403 comes back as a failed (non-skipped) probe, and `internal/trace/coverage.go` maps that to route outcome `unreachable` ("Unreachable via API server - real path not confirmed"); it's the headline *verdict* that stays `unknown`. Either narrow the sentence to the verdict — the true and still-valuable property — or make the 403 a distinct denial outcome in code first and keep the sentence. The quoted denial text also differs from the actual string ("Permission denied. Your identity lacks get services/proxy or get pods/proxy in this namespace.").

5. **The off-switch paragraph needs both halves adjusted.** In-cluster Radar does exercise the proxy verbs (`internal/trace/probes.go` gates the API-server-proxy probe on client presence, not vantage), so "needs neither proxy verb" isn't right; and removing the proxy verbs doesn't switch off the in-cluster lane — direct dials aren't RBAC-gated at all, and the probe Job is gated by `create jobs` / `list pods` / `get pods/log`. Suggest stating it per lane: remove the proxy verbs to switch off relayed probes; remove `create jobs` to switch off probe Jobs; an in-cluster Radar's direct dials are its own network reach.

6. **`inCluster` → `in_cluster`.** The tool argument is snake_case (`tools_diagnose.go`). The schemas are `additionalProperties: false`; the argument-repair middleware does absorb the camelCase spelling, but each repair is logged as a signal that a description or doc is teaching the wrong name (`toolparams.go`) — the doc should use the canonical one. Lines 154 and 190 carry the same pre-existing spelling — worth fixing all three in this pass since this section is the one agents will actually read.

7. **Reconcile the verb triple everywhere it appears.** The new table is the correct one (`create jobs`, `list pods`, `get pods/log`; the kubectl fallback fires when *any* of the three is denied — `runner.go`). Line 190 of this file and `docs/mcp.md` still say `create jobs` + `list`/`get` pods with a pod-create-denied fallback trigger (`docs/mcp.md` needs only the verb-list fix — its `in_cluster` spelling is already right). A least-privilege Role written from the older text grants `get pods` instead of `get pods/log` and hits the fallback for a reason the doc says can't happen.

8. **The Limits pointer now dangles for one case.** The deleted paragraph named the API-server-proxy-can't-reach-an-internal-address case; the new text says "The Limits section below lists the cases", but Limits covers mesh mTLS, NetworkPolicy, and DNS split-horizon only. Add the fourth bullet (it's shipped behavior — the internal-address dial guard, pinned by `TestIsInternalGuardError`) or trim the pointer.

9. **Scope "no raw-request … escape hatch" to `/mcp-readonly`.** On the full `/mcp`, `apply_resource` accepts arbitrary multi-document YAML and `patch_resource` arbitrary patches — RBAC-bounded, but that's the same bound as the kubeconfig alternative the paragraph contrasts against. The sentence is exactly right for the read-only endpoint; say it about that endpoint.

10. **"RBAC checked in the target namespace" → "RBAC required in the target namespace".** Only the Job row is preflighted (SSAR); the proxy verbs are enforced by the API server at request time, with no access review on the trace path. Relatedly, when impersonation can't be established the probe is *skipped* with a vantage reason rather than failing — "the probe fails rather than falling back" should soften to match.

None of this changes the thrust of the section — the underlying design holds up; the doc just needs to describe it exactly. Since the same edit mirrors to radar-docs, best to fix here first and re-sync.

nadaverell:
Closing this as superseded by #1476, which incorporated the reachability work with the corrected identity, RBAC, namespace, consent, and MCP semantics.

Eyal, if you would still like to review the final shape, please review the merged changes in #1476 or the coordinated radar-docs follow-up.

PR #1466, opened 2026-08-20
feat(desktop): keep the app running when the macOS window closes

### LINKED ISSUE #1453: Do not exit the app when closing the main window
## Problem statement

When closing the app window, the entire application exits.

## Proposed solution

Instead of exiting the app just close the window and leave the app running until explicitly requested to exit.

## Alternatives considered

Minimize the window to the dock.

## Additional context

I am using the macOS version of the app, v1.10.0. Thank you!

arlenvasconcelos:
## Description

Closing the window quit the whole process. On macOS it now hides the app instead: the Dock icon stays, a Dock click, Cmd+Tab, or Radar → Show All brings the window back, and the local server keeps serving so MCP clients stay connected. Quitting is explicit — Cmd+Q or Radar → Quit Radar.

Wails maps `HideWindowOnClose` to `[NSApp hide:]`, an *application*-level hide that the Dock icon survives, which is what lets AppKit restore it on reopen. Hiding through the window-scoped path (`runtime.WindowHide`) would order the window out instead and leave nothing to bring it back. Linux and Windows keep quitting on close: there the option hides the *window*, taking the taskbar entry with it, and Wails v2 ships no tray icon — so it would leave a process only `kill` could reach.

macOS also gains the native application menu (About, Hide, Hide Others, Show All, Quit) and File → Close Window on Cmd+W, which calls the same `[NSApp hide:]` the close button does. Quit and About move into the application menu there so nothing is bound twice; both stay in File and Help on the other platforms, where Wails renders menu roles as blank items.

### What running-while-hidden costs

Radar does not stop when the window goes away. A hidden Radar still serves its loopback HTTP and MCP endpoints under your kubeconfig identity, still holds watches open against the API server, and still holds the memory backing its caches. That is the point — MCP clients survive a window close — and it is also why quit is worth doing deliberately. Spelled out in `docs/configuration.md`.

### Deliberately not included

- **No `quitOnWindowClose` config key.** An earlier revision had one. Nothing in #1453 asked for an escape hatch, it was reachable only by hand-editing `~/.radar/config.json` — so the users most likely to want it would never find it — and it put a desktop-only, macOS-only field on the config struct the server and in-cluster binaries also load. One behavior, no toggle.
- **No Window menu.** Wails' stock `WindowMenu` role binds Full Screen to `enterFullScreenMode:`, but AppKit only offers `enterFullScreenMode:withOptions:` and Wails adds no one-argument responder, so nothing handles it. `WailsMenu` also sets `autoenablesItems` to `NO`, so the item does not grey itself out — it would ship looking enabled, holding the system Ctrl+Cmd+F accelerator, and doing nothing. Minimize and Zoom work, but the traffic-light buttons and system Cmd+M already reach both.
- **Multi-instance MCP port-file ownership.** A hidden Radar holding `~/.radar/mcp-port` while another instance starts or exits is a real problem, and this change makes "a Radar you forgot about" a normal state rather than a rare one. It is a cross-process ownership question that predates this PR — filed separately.

## Type of change

- [ ] Documentation update

## How has this been tested?

- [x] Added/updated unit tests — menu shape per platform, the darwin/non-darwin split
- [ ] Tested locally with minikube/kind

Unit tests cover the Go menu tree only. None of the lifecycle behavior here is reachable from Go tests or from browser visual tests — it is AppKit. Pending on a packaged `.app`:

- [ ] Red close button → window gone, Dock icon stays, process alive, endpoint still answers
- [ ] Cmd+W → identical
- [ ] Dock click / Cmd+Tab / Show All → same session, same context, no reload
- [ ] **Fullscreen → close**, both close button and Cmd+W (`[NSApp hide:]` against a window owning a fullscreen space is the least predictable case)
- [ ] Cmd+Q and Radar → Quit Radar → process exits, `~/.radar/mcp-port` removed, timeline flushed
- [ ] Linux/Windows: close still quits, File → Quit present, no blank role items, no Close Window item

## Checklist

## Related issues

Fixes #1453

---

> [!NOTE]
> **Medium Risk**
> macOS lifecycle changes keep cluster watches and MCP serving after the window is closed, which can surprise users who expect close to release kube access; Linux/Windows quit behavior is unchanged.
> 
> **Overview**
> On **macOS**, closing the window no longer exits Radar. Wails **`HideWindowOnClose`** is enabled only on Darwin so the red close button and dock behavior hide the app while the loopback server, cluster watches, and MCP stay up until an explicit quit (**Cmd+Q** / application menu).
> 
> **Linux and Windows** are unchanged: window close still quits, because hiding would leave a process with no dock or tray to restore the UI.
> 
> The native menubar is split by platform: macOS gets **`menu.AppMenu()`** (Quit, Hide, Show All, About from **`mac.AboutInfo`**), **File → Close Window** (**Cmd+W** → **`runtime.Hide`**), and duplicate **Quit** / **About Radar** entries are removed from File and Help. Non-mac builds keep **File → Quit** and **Help → About Radar**.
> 
> **`createMenu`** now takes **`goos`**; lifecycle comments in **`beforeClose`** document that macOS window close no longer triggers quit. Unit tests pin mac vs linux/windows menu labels and roles. **`docs/configuration.md`** documents hide-vs-quit behavior and that a hidden desktop app still holds API access until quit.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11adaff03d95070ae7ab8c7137265e6bf87c4fd3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
> **AI-assisted review disclaimer:** This review was prepared by AI after examining the issue, this PR, Radar's surrounding code, and the relevant Wails v2 macOS implementation. These are suggestions for the author's consideration—not instructions, and not findings that should be accepted automatically. Please triage each point against your own judgment and testing.

The core direction looks right: on macOS, closing the only window should not quit the application; Dock/⌘Tab restoration and explicit ⌘Q match the native platform model. Linux and Windows should continue quitting on close because Radar has no tray/recovery affordance there.

I would consider a few focused refinements before merging:

1. **Remove the `quitOnWindowClose` compatibility setting.** It adds a permanent public config field, helper, tests, documentation, and a conditional menu state solely to preserve the behavior this issue is fixing. There is no request for that escape hatch in #1453, and because it is only exposed in JSON, most users who might want it would not discover it anyway. A simpler contract is easier to understand and maintain: macOS always hides on close; Linux and Windows always quit.

2. **Derive the menu behavior from the platform instead of passing both `goos` and `hideOnClose`.** Once the compatibility setting is removed, those values encode the same predicate. Keeping both permits impossible combinations such as `darwin` with `hideOnClose=false` and creates a future drift risk. On macOS, File → Close Window / ⌘W should always use application-wide `runtime.Hide`, matching Wails' red-close-button path. It should never map ⌘W to Quit. `runtime.WindowHide` is not interchangeable here: it orders out only the window and can leave a running process with no reliable restore path.

3. **Reconsider adding Wails' stock `WindowMenu` in this PR.** In Wails v2.12.0 its Full Screen item is bound to `enterFullScreenMode:`, while AppKit's corresponding view API takes `enterFullScreenMode:withOptions:` and Wails does not appear to provide a one-argument responder. That may leave a permanently disabled Full Screen item. Since #1453 only requires close/hide/restore behavior, the focused choices are to validate the stock menu in the packaged app or omit `WindowMenu` here; building a custom replacement would unnecessarily expand this PR. The native `AppMenu` should remain because it provides the correct About, Hide, Show All, and Quit behavior.

4. **Update the documentation and PR description around the single resulting behavior.** Explain that close/⌘W hides Radar, Dock/⌘Tab restores the same session, and ⌘Q / Radar → Quit Radar terminates it. It is also worth stating why explicit quit matters: while hidden, Radar's loopback HTTP/MCP server and Kubernetes watches continue running under the user's Kubernetes identity. Remove the JSON opt-out text if the setting is removed.

5. **Validate the native lifecycle, not only the Go decision logic.** Before merge, exercise the packaged macOS app and confirm that the red close button and ⌘W hide without stopping the process or endpoint; Dock, ⌘Tab, and Show All restore the same usable session; and ⌘Q fully shuts it down. Test fullscreen explicitly rather than assuming AppKit will hide a fullscreen application. Browser visual tests cannot prove any of these Wails/AppKit behaviors.

I would keep unrelated multi-instance behavior—such as separate Radar processes competing for the MCP port-discovery file—out of this PR. That is a broader cross-process ownership problem and should only be considered separately if it proves worth addressing.

arlenskh:
> > **AI-assisted review disclaimer:** This review was prepared by AI after examining the issue, this PR, Radar's surrounding code, and the relevant Wails v2 macOS implementation. These are suggestions for the author's consideration—not instructions, and not findings that should be accepted automatically. Please triage each point against your own judgment and testing.
> 
> The core direction looks right: on macOS, closing the only window should not quit the application; Dock/⌘Tab restoration and explicit ⌘Q match the native platform model. Linux and Windows should continue quitting on close because Radar has no tray/recovery affordance there.
> 
> I would consider a few focused refinements before merging:
> 
> 1. **Remove the `quitOnWindowClose` compatibility setting.** It adds a permanent public config field, helper, tests, documentation, and a conditional menu state solely to preserve the behavior this issue is fixing. There is no request for that escape hatch in [Do not exit the app when closing the main window #1453](https://github.com/skyhook-io/radar/issues/1453), and because it is only exposed in JSON, most users who might want it would not discover it anyway. A simpler contract is easier to understand and maintain: macOS always hides on close; Linux and Windows always quit.
> 2. **Derive the menu behavior from the platform instead of passing both `goos` and `hideOnClose`.** Once the compatibility setting is removed, those values encode the same predicate. Keeping both permits impossible combinations such as `darwin` with `hideOnClose=false` and creates a future drift risk. On macOS, File → Close Window / ⌘W should always use application-wide `runtime.Hide`, matching Wails' red-close-button path. It should never map ⌘W to Quit. `runtime.WindowHide` is not interchangeable here: it orders out only the window and can leave a running process with no reliable restore path.
> 3. **Reconsider adding Wails' stock `WindowMenu` in this PR.** In Wails v2.12.0 its Full Screen item is bound to `enterFullScreenMode:`, while AppKit's corresponding view API takes `enterFullScreenMode:withOptions:` and Wails does not appear to provide a one-argument responder. That may leave a permanently disabled Full Screen item. Since [Do not exit the app when closing the main window #1453](https://github.com/skyhook-io/radar/issues/1453) only requires close/hide/restore behavior, the focused choices are to validate the stock menu in the packaged app or omit `WindowMenu` here; building a custom replacement would unnecessarily expand this PR. The native `AppMenu` should remain because it provides the correct About, Hide, Show All, and Quit behavior.
> 4. **Update the documentation and PR description around the single resulting behavior.** Explain that close/⌘W hides Radar, Dock/⌘Tab restores the same session, and ⌘Q / Radar → Quit Radar terminates it. It is also worth stating why explicit quit matters: while hidden, Radar's loopback HTTP/MCP server and Kubernetes watches continue running under the user's Kubernetes identity. Remove the JSON opt-out text if the setting is removed.
> 5. **Validate the native lifecycle, not only the Go decision logic.** Before merge, exercise the packaged macOS app and confirm that the red close button and ⌘W hide without stopping the process or endpoint; Dock, ⌘Tab, and Show All restore the same usable session; and ⌘Q fully shuts it down. Test fullscreen explicitly rather than assuming AppKit will hide a fullscreen application. Browser visual tests cannot prove any of these Wails/AppKit behaviors.
> 
> I would keep unrelated multi-instance behavior—such as separate Radar processes competing for the MCP port-discovery file—out of this PR. That is a broader cross-process ownership problem and should only be considered separately if it proves worth addressing.

Thanks — took all five. What changed since your review:

- **1. `quitOnWindowClose` removed** (`28aa8dc4`) — config field, helper, tests and docs all gone. macOS always hides on close; Linux/Windows always quit. No toggle.
- **2. Menu derived from the platform alone** (`8a7367e2`) — `createMenu(app, version, goos)`; the `hideOnClose` parameter is gone, so `darwin` + `hideOnClose=false` is no longer representable. On macOS, File → Close Window / ⌘W always calls `runtime.Hide`, matching the red-close-button path — never Quit, never `WindowHide`.
- **3. `WindowMenu` dropped** (`7311de70`) — you were right about `enterFullScreenMode:`; Wails also sets `autoenablesItems: NO`, so it would have shipped looking enabled while holding ⌃⌘F and doing nothing. Native `AppMenu` kept for About/Hide/Show All/Quit.
- **4. Docs and PR description rewritten** (`11adaff0`) — `docs/configuration.md` has a "Desktop Window Behavior" section covering close/⌘W → hide, Dock/⌘Tab/Show All → restore, ⌘Q → quit, and spells out that a hidden Radar keeps serving loopback HTTP/MCP under the user's kubeconfig identity and keeps its watches and caches open. JSON opt-out text removed.
- **5. Native validation — still open.** Go tests cover the menu tree only. The packaged `.app` checklist is in the PR description and unchecked, fullscreen → close included.

nadaverell (approved):

PR #1467, opened 2026-08-21
Reopen the Desktop app on its last used cluster

### LINKED ISSUE #1454: Restore app from the last cluster used
## Problem statement

Every time I open the app it goes to the same "default" cluster regardless of what cluster I was working with from the last session. With many clusters configured this adds a little friction to workflow, especially if the one I want is not the default one.

## Proposed solution

Whichever cluster was active in the last session (the last time the app exited) should be selected when the app is opened.

## Additional context

Using the macOS version of the app, v1.10.0. Thank you!

arlenvasconcelos:
## Summary

Radar Desktop now reopens on the cluster its window last used instead of following kubeconfig `current-context` on every launch. Terminal entrypoints retain kubectl semantics.

## What changed

- Add Desktop-scoped cluster memory, enabled by default and configurable in Settings → Connection. Disabling it takes effect on the next Desktop start and clears the remembered cluster.
- Record the initial durable context and successful durable context switches using exact kubeconfig file + in-file context identity.
- Fall back safely to kubeconfig `current-context` when the remembered context is gone, while reporting the reason in Diagnostics. An existing but unreachable remembered cluster remains selected and fails loudly.
- Keep temporary CAPI and otherwise unresolvable contexts out of the memory, and preserve a missed preference across retries or automatic reconnects.
- Keep Desktop state private from shared/authenticated/cloud-tunnel sessions and `/api/settings`.
- Refuse automatic context-memory writes when local settings cannot be read, preserving other preferences.
- Document Desktop versus CLI startup semantics and cover single-/multi-kubeconfig, collision, fallback, privacy, and settings-failure scenarios.

## Testing

- `make tsc`
- `make test`
- `make build`
- `make desktop-binary`
- Live-tested against `kind-radar-rollouts-demo` and `kind-radar-c3`: initial recording, in-app context switch, process restart restore while kubeconfig current-context remained unchanged, real resource loading, and frontend startup.
- Live-tested an existing but unreachable remembered context: Radar kept it selected and surfaced a network disconnect instead of silently switching clusters.
- Live-tested a deleted remembered context: Radar fell back to kubeconfig current-context, self-healed the saved identity, and recorded the reason in Diagnostics.
- Exercised the Settings toggle through the frontend: disabling restore persisted the setting, cleared a saved different context on restart, and followed kubeconfig current-context.

Closes #1454

---

> [!NOTE]
> **Medium Risk**
> Changes which cluster Desktop connects to at startup; mitigations (exact file+context match, no CLI coupling, no silent fallback to a different cluster) reduce but do not eliminate wrong-cluster risk if kubeconfig layout changes.
> 
> **Overview**
> **Radar Desktop** now starts on the cluster its window last used instead of always following kubeconfig `current-context`. **`kubectl radar` / `radar`** keep kubectl semantics: they ignore Desktop memory and never record UI switches.
> 
> Desktop opts in via `RestoreLastDesktopContext` (default **on** in `config.json`, toggle in **Settings → Connection**). The remembered cluster is stored as `lastDesktopContext` in `~/.radar/settings.json` with **kubeconfig file + in-file context name**, so multi-kubeconfig name collisions cannot silently point at a different cluster. Restore uses an **exact match only**; missing or stale entries fall back to `current-context` with **Diagnostics** warnings, while an existing but unreachable cluster stays selected and fails visibly.
> 
> Recording is limited to **local Desktop** (disabled for auth-enabled or cloud-tunnel processes), skips **CAPI temp** contexts, uses **`UpdateChecked`** so corrupt settings are not overwritten, and **`/api/settings` strips** the remembered cluster from GET/PUT responses. K8s init gains **`PreferredContext`** / `pickInitialContext` preference resolution wired from `InitializeK8s` and context-switch callbacks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3487422279cdc51f95fae0de8ac33f43533a367d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell (changes requested):
this I guess makes sense for desktop but may be counter intuitive for CLI-driven radar (or 'kubectl radar', or --diagnose invocations), where it is probably more expected to respect the kubectl context.

Posting a more detailed AI-assisted review in a sec but I think the change should be limited to desktop, maybe with a setting that users can opt in/out of this (choose between remembering last context and defaulting to current kubeconfig context). Claude/codex found a couple more things to note on handling deleted context etc.

nadaverell:
> **AI-assisted product review for consideration.** This was generated by Codex, independently cross-reviewed by Claude, and then triaged rather than accepted wholesale. Please use your own judgment—there is no expectation to automatically accept every suggestion below. This review is about the product behavior and user expectations, not code correctness.

The premise from #1454 is right: the Desktop app should reopen on the cluster the user was last working in. That is the natural behavior for a stateful GUI launched from the Dock, and the PR has made several thoughtful choices around recording only successful switches, surviving force-quits, avoiding CAPI temporary kubeconfigs, and not rewriting kubeconfig.

The main concern is scope. The same remembered state currently also changes the default for `kubectl radar`, `radar`, and `radar diagnose --standalone`. For a kubectl plugin or terminal-driven diagnostic, the expected contract is different: an explicit `--context` should win, otherwise the kubeconfig's `current-context` should win. With the proposed behavior, `kubectl config use-context staging && kubectl radar` can instead open a cluster selected in Radar days earlier. Since Radar includes mutating actions such as apply, scale, restart, drain, rollback, and sync, landing on an unexpected cluster is a safety problem rather than merely a small surprise.

Suggested product shape for consideration:

1. **Remember by default in Desktop only.** `cmd/desktop` should restore its last successfully selected cluster. `kubectl radar`, the `radar` CLI launch, and `diagnose --standalone` should remain deterministic on kubeconfig `current-context` unless the user explicitly opts the CLI into remembered behavior. Keep the remembered value Desktop-scoped, or otherwise enforce a hard separation so a Desktop switch cannot steer a later terminal command.

2. **Make explicit targeting strict and non-persistent.** `--context` should affect that invocation only. An unknown context should exit with a clear error (ideally with close matches), not silently connect to `current-context` and remember the fallback. Keep `"context"` in `config.json` as a useful durable pin: while set, always start there, but do not use it to overwrite remembered Desktop state.

3. **Do not guess when remembered identity is ambiguous.** Restore only when Radar can establish that it found the same context/cluster. If the saved source is gone or identity cannot be established, open kubeconfig `current-context` and explain the fallback. A same-named context from another file is not sufficient evidence because connecting to the wrong cluster is worse than losing the convenience.

4. **Handle missing and unreachable memories differently.** Missing/renamed/ambiguous memory can safely fall back to `current-context` with visible notice. If the exact remembered cluster still exists but is unreachable (for example, VPN is down), keep the current fail-loud behavior rather than silently redirecting; explain that Radar reopened the last-used cluster and offer the context switcher.

5. **Make the changed startup behavior legible.** At minimum, surface exceptional fallback and restored-but-unreachable states in the UI rather than only logs. A small one-time notice on the first successful restore after this feature ships (for example, “Reopened on prod-eu — where you left off”) would teach the new behavior; it should not become recurring toast noise. If an opt-out remains, expose “Startup cluster: Last used / kubeconfig current context” in the Desktop Connection settings rather than requiring Desktop users to edit JSON. Turning remembrance off should clear/reset the latent selection instead of reviving a months-old value when re-enabled.

6. **Apply the single-user rule consistently.** In addition to auth and Cloud modes, a deliberately non-loopback/shared listener should not persist or expose one viewer's remembered cluster.

I don't think this necessarily needs to be split into multiple PRs—the important split is in the product semantics, not the contribution process. The concise target is: **stateful Desktop, deterministic CLI, strict explicit overrides, and no silent wrong-cluster guesses.**

nadaverell (approved):
Approved after end-to-end code review, full repository QA, Desktop compilation, and live two-cluster testing. Verified initial persistence, in-app switching, restart restore without mutating kubeconfig current-context, fail-loud unreachable behavior, missing-context fallback with Diagnostics, and Settings opt-out clearing on restart.

PR #1507, opened 2026-08-26
fix(ui): clear the error boundary on navigation

hisco:
## The problem

The error boundary around the main content area latched. Once a view threw, `hasError` stayed set, and navigating re-renders that same boundary instance rather than remounting it - so the fallback stayed on screen. Clicking another view in the nav rail changed the URL and the document title but left the error panel up, so the click read as doing nothing.

Recovery was possible but undiscoverable: navigate first, then press **Try Again**, and the already-changed route renders. Pressing **Try Again** without navigating does nothing at all - it re-renders the same failed subtree against the same data and throws again immediately. So a crash in one view read as a crash of the whole app, and the healthy deep routes looked unreachable.

This was found while investigating a real incident: a removed `/api/dashboard` field made the home dashboard's topology card dereference `undefined`, and the boundary replaced the entire content area. The missing field has since been fixed. The dead end it exposed had not been.

## The change

`ErrorBoundary` takes a `resetKey` and clears a caught error when it changes. `App` passes `location.pathname`, so any navigation recovers.

Keyed on the full path rather than the current view: `mainView` is only the first path segment, so a crash under `/resources` would otherwise still trap the user while they switch kinds.

**`resetKey` rather than a `key`.** Remounting on every route change would also discard the children's state while nothing is wrong - `ResourcesView` in particular deliberately persists across kind changes and reports them through `onKindChange`. Resetting only clears the error, so a healthy tree keeps its identity and the fix costs nothing when there is no error.

A lint check requires every `ErrorBoundary` to carry a reset path, since a boundary without one is a dead end by construction rather than by accident. It follows the existing `check-no-native-title.mjs` pattern and runs as part of `npm run lint`.

## Verification

`npm run tsc` clean, `npm run lint` 0 errors, 485 tests pass.

The guards were checked against a revert, not just against the fix: removing `getDerivedStateFromProps` fails 3 of the 4 new tests, and restoring the bare `<ErrorBoundary>` fails the lint check with the file and line.

## Not covered here

This makes a crashed view recoverable; it does not make one card's failure stop at that card. The boundary still spans all 15 views, so a single card can still blank the content area - it is just no longer a dead end. Narrowing it is a larger change and wants per-card fallbacks that visibly say they failed rather than rendering empty, which is worth doing separately.

https://claude.ai/code/session_01JJGtvmJpie4iq6h8AiUd6t

---

> [!NOTE]
> **Low Risk**
> Localized UX recovery around an existing error boundary; no auth, data, or API changes.
> 
> **Overview**
> Fixes a **latched** main-content `ErrorBoundary`: after one view threw, the fallback stayed up because navigation only re-rendered the same boundary instance instead of remounting it.
> 
> **`ErrorBoundary`** now requires a **`resetKey`** and uses **`getDerivedStateFromProps`** to clear `hasError` when that key changes, without remounting children on every route change (so healthy views keep state). **`App`** passes **`location.pathname + location.search`** so resets follow full routes—including kind switches under `/resources` and selection in the query (`?resource=`, `?release=`).
> 
> Adds **`ErrorBoundary.test.tsx`** covering reset on path change, kind segment change, query change, and no reset when the key is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7b9c520aa896170159d1b33d912875e103aeec2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hisco:
## Manual verification in a browser

The unit tests cover the reset logic, not React actually performing it, so I drove this by hand against a local kind cluster. The crash was injected into `HomeView` behind a URL flag, reproducing the original failure exactly - including the `nodeCount` message from the incident.

Each step below was run twice: once with `main`'s boundary, once with this branch. Nothing else differed.

### Without the fix

| Step | Result |
|---|---|
| Load the crashing view | Error panel: `Cannot read properties of undefined (reading 'nodeCount')`. Nav rail and header intact. |
| Click **Topology** in the nav rail | URL became `/topology?group=namespace`, tab title became `Topology · Radar`, **error panel still on screen** |
| Click **Try Again**, having navigated | Topology rendered |
| Click **Try Again** x3 without navigating | Still errored each time; console errors went 2 → 8 |

So the click on Topology looked inert, and recovery needed a second click on a button that does nothing on its own. That is the behaviour this PR removes.

### With the fix

| Step | Result |
|---|---|
| Load the crashing view | Error panel, as above |
| Click **Topology** - one click, no **Try Again** | Error cleared, topology rendered with live data (Service 4, Deployment 4, Rollout 2, DaemonSet) |
| Click **Resources**, then **Home** | Both clean, error did not return |
| Client-side nav changing **only the query**, same path | Error cleared - the case a path-only key would have missed |

### The `resetKey` vs `key` decision, checked rather than argued

Captured a specific `<table>` node inside the boundary, performed a navigation that changes the reset key, and compared node identity afterwards: same object, still attached.

That is the concrete reason this uses `resetKey` and not `key`. A `key` would have discarded that table - and the list state and in-flight fetches behind it - on every query change, including all the changes where nothing is wrong.

### Correction to the original description

The first version of this description said a page reload was the only way out. That is not what happens, and I have corrected it above. Navigating and then pressing **Try Again** did recover. The real defect is that the navigation click appears to do nothing, and that **Try Again** alone is inert on a persistently failing view.

### Not covered

The boundary still spans all 15 views, so one card can still blank the content area - it is no longer a dead end, but the blast radius is unchanged. The fallback also still shows a raw exception string to operators and offers no route to the existing bug-report flow. Both are worth their own change.

PR #1519, opened 2026-08-28
fix(k8s): fence stale permission probes behind a generation counter

### LINKED ISSUE #1169: Permission-cache invalidation lacks a generation check — stale in-flight probe can overwrite the new cluster's result
`CheckResourcePermissions` computes without holding the mutex and publishes its result unconditionally. `InvalidateResourcePermissionsCache` only clears the pointer — there is no generation counter — so during a context switch an old in-flight probe can finish after invalidation and republish the previous cluster's result with a fresh TTL. Anything that reads the cached result inside that window (typed informer wiring via `Scopes`, the dynamic cache's `ScopeCandidates`/`Namespace` snapshot) can be wired against the previous cluster's namespaces until the next re-probe.

This predates the multi-namespace work (`Scopes`/`Namespace` had the same exposure); the fanout only widens what a stale result can affect. Mitigations today: switch-scoped contexts fail the old probe's remaining calls fast, and error results carry a short TTL — but the publish itself is still unconditional.

Proposed fix: a generation counter incremented by `InvalidateResourcePermissionsCache`; probes capture it before starting and publish only if unchanged. Needs its own PR — it changes core probe caching used by every consumer — plus a context-switch race test where an old result completes after invalidation.

Surfaced by adversarial review on #1160.

arlenvasconcelos:
## Description

`CheckResourcePermissions` deliberately runs its probe pass **without** holding `resourcePermsMu`, so a context switch is never blocked behind network calls. But the publish at the end was unconditional, and `InvalidateResourcePermissionsCache` only cleared the cached pointer — nothing recorded that an invalidation had happened.

So a probe still in flight when the cluster changed would finish and republish the **previous** cluster's per-kind scopes with a fresh 60s TTL. Anything reading the cache in that window could be wired against the old cluster's namespaces: typed informer wiring via `Scopes` (`internal/k8s/cache.go`), the dynamic CRD cache's `ScopeCandidates`/`Namespace` snapshot (`internal/k8s/dynamic_cache.go`), and every `GetCachedPermissionResult()` consumer in `internal/server`, `internal/mcp` and audit. Existing mitigations (switch-scoped contexts failing the old probe's remaining calls fast, short TTL on error results) shrink the window but don't close it.

**Fix** — a generation counter guarded by the existing mutex, as proposed in the issue:

- `resourcePermsGeneration uint64` alongside `cachedPermResult`, under `resourcePermsMu`. No new locks.
- `InvalidateResourcePermissionsCache` increments it while clearing the pointer, so an invalidation is now observable by in-flight probes.
- `CheckResourcePermissions` captures the generation under the same `RLock` as its cache-miss check (strictly before probing starts) and, at publish time, stores the result only if the generation is unchanged. A superseded probe is dropped with a log line and returns `nil`.
- Both non-test callers already handle a `nil` result; `InitResourceCache` now fails initialization explicitly instead of dereferencing it.
- `ResetTestState` routes through `InvalidateResourcePermissionsCache` so tests get the generation bump too.

Non-race paths behave exactly as before. No API changes.

**Out of scope (noted, not changed):** `CheckCapabilities` / `CheckNamespaceCapabilities` and the per-user caches have the same unconditional-publish shape. A stale publish there survives at most the 60s TTL and only affects UI control gating — the apiserver still enforces real RBAC on writes — so the same fence pattern applies if it ever bites, but the issue scopes this PR to the resource-permission probe cache.

## Type of change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

## How has this been tested?

New regression test `internal/k8s/capabilities_generation_test.go` — `TestStaleProbeResultDroppedAfterInvalidation`. A gated fake dynamic client holds the probe goroutines mid-flight (the old cluster's probe), the test invalidates the cache (the context switch), releases the probe, and asserts the stale publish is dropped. It then swaps in an all-denied "new cluster" fake and asserts the fresh probe's publish still lands — the fence must not wedge the cache shut.

- Reproduces the race before the fix: `capabilities_generation_test.go:97: stale in-flight probe republished the previous cluster's result after invalidation (Pods=true)`
- After the fix: `go test -race -run TestStaleProbeResultDroppedAfterInvalidation -count=5 ./internal/k8s/` passes
- Full package: `go test -race ./internal/k8s/` — ok (31.4s), including the pre-existing `TestConcurrentInvalidateDuringPermissionCheck` deadlock guard
- `go vet ./internal/k8s/` and `go build ./...` clean

## Checklist

## Related issues

Fixes #1169

---

> [!NOTE]
> **Medium Risk**
> Touches cluster-switch informer wiring and permission cache publishing; incorrect behavior could mis-scope watches, but the change closes a known stale-cache race with a targeted generation fence and regression test.
> 
> **Overview**
> Fixes a **kubeconfig context-switch race** where an in-flight `CheckResourcePermissions` probe could still **publish the previous cluster’s scopes** into the 60s cache after `InvalidateResourcePermissionsCache`, wiring informers and dynamic CRD fallbacks against the wrong cluster.
> 
> A **`resourcePermsGeneration`** counter (under the existing `resourcePermsMu`) is captured before probes run and checked at publish time; **`InvalidateResourcePermissionsCache`** clears the cache and bumps the generation so superseded probes are **dropped** (`nil` + log) instead of cached. **`InitResourceCache`** now fails init explicitly when the probe is superseded, and **`ResetTestState`** invalidates via the same path so tests fence in-flight probes too.
> 
> Adds **`TestStaleProbeResultDroppedAfterInvalidation`** with a gated fake dynamic client to assert stale publishes are dropped while a post-invalidation probe still caches normally.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2a04ad7a5d9a60fae9e3542e3509bd8fdd82bf7c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/k8s/cache.go
+++ b/internal/k8s/cache.go
@@ -343,6 +343,10 @@ func InitResourceCache(ctx context.Context) error {
 			initErr = ctx.Err()
 			return
 		}
+		if permResult == nil {
+			initErr = fmt.Errorf("resource permission probe superseded during initialization")
+			return
+		}
 
 		// scopes drives which informers are created and at what scope.
 		// k8score routes each kind through the matching factory (cluster-wide
--- a/internal/k8s/capabilities.go
+++ b/internal/k8s/capabilities.go
@@ -711,6 +711,12 @@ var (
 	resourcePermsExpiry   time.Time
 	resourcePermsTTL      = 60 * time.Second
 	resourcePermsErrorTTL = 5 * time.Second // Short TTL when API errors caused fail-closed results
+	// resourcePermsGeneration (guarded by resourcePermsMu) fences in-flight
+	// probes against invalidation: probes run without the lock, so a context
+	// switch can invalidate mid-probe and the old cluster's result would
+	// otherwise republish with a fresh TTL. Probes capture the generation
+	// before starting and publish only if it is unchanged.
+	resourcePermsGeneration uint64
 )
 
 // resourceProbe describes one typed-resource probe target. The probe issues
@@ -940,6 +946,7 @@ func CheckResourcePermissions(ctx context.Context) *PermissionCheckResult {
 		resourcePermsMu.RUnlock()
 		return result
 	}
+	observedGeneration := resourcePermsGeneration
 	resourcePermsMu.RUnlock()
 
 	// Compute probes WITHOUT holding the write lock — concurrent callers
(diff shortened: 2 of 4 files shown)


arlenvasconcelos:
Superseded by #1522, which implements the same generation fence plus a client-generation guard for context switches. Closing in favour of that one.

PR #1525, opened 2026-08-28
fix(copy): stream pod file downloads and save them natively on desktop

### LINKED ISSUE #1511: Radar fails to download file from pod
## Describe the bug

Radar app&desktop fails to download a large file ~250MB from the pod. 
In smaller files download works fine.

## To reproduce

1. on Resources tab choose a pod
2. Click on 'Browse files'
3. File explorer opens
4. Go to the folder where the csv file is
5. Click on 'Download file'
6. File is not downloaded

## Expected behavior

The csv file should be downloaded.

## Logs

```
2026/08/27 13:06:56 [copy] Failed to read file from tar <namespace>/<pod>: unexpected EOF
2026/08/27 13:06:56 "GET http://localhost:9280/api/pods/<namespace>/<pod>/files/download?container=<container>&path=%2Foutput%2Fdata_2026-08-27T10%3A02%3A03.511991310Z.csv HTTP/1.1" from 127.0.0.1:50126 - 500 73B in 32.211367041s
```

## Additional context

If debug logs are needed i can provide them

arlenvasconcelos:
Fixes #1511

Downloading a large file from a pod failed with `unexpected EOF` and a 500 after ~32s. Fixing that surfaced a second, independent failure in the desktop app, and a third in how failures were reported at all.

## 1. The stream was buffered, and drops were not survivable

`handlePodFileDownload` read the entire tar stream into a `bytes.Buffer` before extracting it. Exec transports routinely kill long transfers (LBs, tunnels, apiserver proxies), and a drop can arrive as a *clean* close — client-go's WebSocket demux treats `CloseNormalClosure` as plain `io.EOF`, so `StreamWithContext` returns nil while stdout is truncated. That is exactly what the reported log shows: the error came from tar extraction, not from the exec.

This is the long-standing Kubernetes problem that broke `kubectl cp` for large files (kubernetes/kubernetes#60140). kubectl's answer is the `--retries` resume: re-exec `tar cf - <src> | tail -c+<offset>` and continue from the byte where the stream died.

`resumingTarStream` does the same here, and the response is now streamed to the client instead of held in memory twice. A premature clean EOF also triggers a resume, since it is indistinguishable from real completion at that layer.

The resume execs via `/bin/sh` rather than a bare `sh`, matching the file listing paths — direct exec through the container runtime can fail to resolve binaries that are reachable through the shell.

## 2. The desktop app capped downloads at ~75MB

Independent of the above, and the reason this still looked broken after the server fix.

The webview cannot save from a blob URL, so downloads were routed back to the server through `/desktop/save-file`: read the whole file into a blob, base64-encode it (+33%), POST it as JSON — against a 100MB body cap. A 250MB file needed over a gigabyte of webview memory and was rejected regardless. There is a second ceiling too: V8's max string length puts a hard stop around 384MB.

The desktop app does not show a save dialog; it writes to `~/Downloads`. The bytes were already on the server, so they now go straight there:

- `SetSaveStreamFunc` installs a streaming save hook. The desktop app copies the reader to `~/Downloads` and removes the partial file if the copy fails.
- `deliverPodFile` serves `save=native` by handing the extracted file to that hook and replying with the saved path — constant memory, no size ceiling.
- The hook is installed **only** by the desktop app, so `save=native` does nothing in server or in-cluster mode. There is a test pinning that.
- `downloadFromUrl` replaces the blob-level helper, falling back to the old upload path for endpoints that do not support native save.

## 3. Failures were invisible

Both file browsers reported download errors with `console.error` only — the spinner stopped and nothing appeared on screen. They now raise a toast.

That exposed a further problem: toasts sat at z-index 50 while the file browser modals cover the viewport at z-100 with a 60%-black blurred backdrop, so the toast arrived behind frosted glass. Toasts moved to 200 — above modals and the omnibar/search popovers (121, 130), below tooltips (9999).

## Testing

10 new unit tests across `deliverPodFile` (default streaming, `save=native` inert without the hook, correct bytes excluding tar padding, traversal filename rejected, save failure surfaced) and `saveStream` (writes, collisions, partial-file cleanup on error).

Verified against a live EKS cluster:

| Case | Result |
|---|---|
| 250MB, browser | byte-identical (sha256), 7s |
| 1GB, desktop native save | byte-identical, 27s, 75-byte response |
| `save=native` in server mode | inert — streams the file as normal |
| Error toast | legible above the modal backdrop |
| Success toast | shows the saved path and "Show in Finder" |

## Not included

`/api/images/file` (the image filesystem browser) still routes desktop downloads through the base64 path and keeps the ~75MB ceiling. It lives in `internal/images`, which has no access to the server's save hook — threading it through is a separate change. That browser does gain the error toast here.

---

> [!NOTE]
> **Medium Risk**
> Changes pod exec download/resume behavior and adds a GET side effect that writes to local disk when `save=native` is used, though it is gated to loopback origins and only when the desktop hook is installed.
> 
> **Overview**
> Fixes large pod file downloads that failed with truncated tar streams and desktop size limits.
> 
> **Server:** Pod file download no longer buffers the full `tar` exec output. A **`resumingTarStream`** re-runs `tar … | tail -c+<offset>` (kubectl `cp --retries` style) when the exec transport drops or silently truncates, then streams the extracted file with **`io.CopyN`** instead of `ReadAll`. **`deliverPodFile`** adds optional **`save=native`**: when the desktop app registers **`SetSaveStreamFunc`**, bytes go straight to disk and the handler returns JSON `{path}`; cross-origin requests to that branch are rejected via **`localOriginOK`**, and filenames go through shared **`sanitizeFilename`**.
> 
> **Desktop:** **`saveStream`** copies an `io.Reader` into `~/Downloads` (shared collision logic with **`saveFile`**), deleting partial files on failure; wired at startup with **`SetSaveStreamFunc`**.
> 
> **UI:** Pod/image file browsers use **`useFileDownload`** / **`downloadFromUrl`** (`save=native` on desktop) with success/error toasts; toast **z-index** raised so messages appear above file modals.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3de7e5cdfb116030186b62afbc4b81e2c51357f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/desktop/app.go
+++ b/cmd/desktop/app.go
@@ -127,12 +127,19 @@ func (a *DesktopApp) saveStream(defaultFilename string, src io.Reader) (string, 
 	if err != nil {
 		return "", err
 	}
-	if _, err := io.Copy(f, src); err != nil {
-		f.Close()
-		os.Remove(path)
-		return "", err
+	// Buffered writes commonly report failure at close (ENOSPC, EIO on a
+	// network mount), so the close error matters as much as the copy error —
+	// either one has to take the partial file with it.
+	_, copyErr := io.Copy(f, src)
+	closeErr := f.Close()
+	if copyErr == nil {
+		copyErr = closeErr
 	}
-	return path, f.Close()
+	if copyErr != nil {
+		os.Remove(path)
+		return "", copyErr
+	}
+	return path, nil
 }
 
 // domReady is called when the webview DOM is ready.
--- a/cmd/desktop/app.go
+++ b/cmd/desktop/app.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"path/filepath"
@@ -61,6 +62,7 @@ func (a *DesktopApp) startup(ctx context.Context) {
 	a.ctx = ctx
 	startNativeMouseMonitor(ctx)
 	a.srv.SetSaveFileFunc(a.saveFile)
+	a.srv.SetSaveStreamFunc(a.saveStream)
 
 	// The OS titlebar must track the active kubeconfig context for the same
 	// reason the in-page selector does: a fleet UI showing the wrong cluster
@@ -70,10 +72,10 @@ func (a *DesktopApp) startup(ctx context.Context) {
 	})
 }
 
-// saveFile writes a file to the user's Downloads folder.
+// downloadPath resolves a collision-free path in the user's Downloads folder.
 // We write directly to ~/Downloads instead of showing a native save dialog
 // because Wails' SaveFileDialog is immediately dismissed by the webview on macOS.
-func (a *DesktopApp) saveFile(defaultFilename string, data []byte) (string, error) {
+func downloadPath(defaultFilename string) (string, error) {
 	home, err := os.UserHomeDir()
 	if err != nil {
 		return "", err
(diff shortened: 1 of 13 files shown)


hisco:
Heads up: #1521 merged for #1511 a few hours before I saw this one, and neither of us knew the other was working it. That is a process miss on our side, not yours. Your diagnosis was right; it came down to timing.

Your PR also caught a couple of things mine missed. Let's sync on what's worth pulling across before either of us moves on it, rather than me guessing here.

PR #1558, opened 2026-08-31
Expose exact-GVR observation coverage

nadaverell:
## Status — DRAFT

Unit, race, build, and full-privilege live API verification are green. A real restricted-RBAC/apiserver run is still required before this truth contract is ready for review.

## Why this exists

The GPU, batch, and AI/ML breadth work exposed a shared observation prerequisite: “Radar found zero” is not equivalent to “Radar continuously observed the complete authorized scope and found zero.” This P0.2a foundation records whether an installed exact GVR is watched, syncing, deferred, denied, unsupported, namespace-partial, or stale. It exposes the truth contract; later consumers must propagate it rather than presenting false-complete aggregate results.

Controller-native contexts and drilldowns are separate verticals. This PR does not add a GPU view or semantic controller integration.

## Summary

- add a read-only, exact-GVR dynamic-cache observation contract for unwatched, deferred, syncing, watched, denied, and unsupported resources
- retain warmup, small-eager, and on-demand origins plus observation start, scope, namespace partiality, and bounded-fanout state
- retain discovery attempt, last-successful refresh, error, partial, and stale state through the existing source-coverage vocabulary
- add CRD observations to the API-resources envelope and filter namespace identities to the authenticated caller's RBAC ceiling

## Scope-integrity fixes

- Non-authoritative namespace enumeration remains incomplete even when it yields no explicit fallback candidates. A resulting GVR-global denial is exposed as partial/truncated (`access_denied_partial_probe`), never as authoritative proof that no authorized namespace exists.
- A denial from a request for one preferred namespace remains request-scoped and cannot replace the retained observation for the GVR as a whole.
- If cluster-wide count probing is denied and any namespace fallback fails transiently, Radar defers the discovery decision unless another fallback succeeds. Only conclusive authorization failures become denial evidence.

## Remaining validation gate

Before marking this ready, run Radar with an identity that is denied cluster-wide list access but has mixed namespace grants and denials. Verify that the API reports namespace-partial observation honestly, filters namespace identities to the caller, and that a preferred-namespace denial cannot replace the GVR-global observation. Those cases are covered by unit tests today, but have not been exercised against a real API server. #1559 and user-visible aggregate coverage claims remain gated on this run.

## Dependency and merge map

- This PR targets `main` and is the base of [#1559](https://github.com/skyhook-io/radar/pull/1559).
- Merge order: merge **#1558**, then rebase/retarget and merge **#1559**.
- #1559 makes the 37 curated identities continuously warm when installed; #1558 makes the resulting watch/readiness/partiality state observable.
- The exact-identity work [#1557](https://github.com/skyhook-io/radar/pull/1557) and [#1560](https://github.com/skyhook-io/radar/pull/1560), and the JobSet vertical, are independent tracks.

## Review focus

- whether each observation state is authoritative for the exact GVR and scope it names
- namespace-candidate authority, truncation, and denial propagation under restricted RBAC
- auth-versus-transient precedence across cluster-wide and namespace-fallback probes
- distinction between installed-but-denied, unsupported list/watch verbs, deferred, syncing, and not discovered
- discovery freshness and recovery transitions without converting missing evidence into absence
- authenticated response filtering: never expose namespace identities above the caller's RBAC ceiling

## Deliberate boundaries

This is an enabling contract only. It does not add a runtime-kind catalog, force watches, or change Issues, Search, Timeline, Topology, resource lists, or MCP to consume coverage. Unsupported means an installed exact GVR lacks required list/watch support; it does not mean an undiscovered API is absent.

## Verification

- `make test`
- `make build`
- `(cd pkg && go test -race ./k8score)`
- `go test ./internal/k8s ./internal/server`
- `make tsc`
- `git diff --check origin/main...HEAD`
- independent read-only review of the final scope-integrity and probe-error semantics: clean
- live full-privilege run against `kind-radar-gpu-ecosystem-demo`: `/api/health`, `/api/api-resources`, and `/api/capabilities` returned successfully; installed Kueue v1beta2, JobSet v1alpha2, and Ray v1 GVRs reported continuously watched; the snapshot contained 42 watched and 6 list/watch-unsupported resources
- restricted-RBAC live run: pending, as described above

Visual testing was skipped because there is no rendered UI change.

Part of RAD-418.

### DIFF
--- a/internal/k8s/capabilities.go
+++ b/internal/k8s/capabilities.go
@@ -68,35 +68,37 @@ type ResourcePermissions struct {
 //     namespace-scoped on the same cluster, and a namespace-scoped kind may
 //     be readable in several explicitly named namespaces.
 type PermissionCheckResult struct {
-	Perms           *ResourcePermissions
-	NamespaceScoped bool   // True if at least one resource type ended up namespace-scoped
-	Namespace       string // The fallback namespace used for namespace-scoped probes
-	Scopes          map[string]k8score.ResourceScope
-	ScopeNamespaces map[string][]string // Per-kind namespace-scoped informer fanout; nil/empty means use Scopes[key].Namespace
-	ScopeCandidates []string            // The candidate namespaces the probe walked — the dynamic CRD cache probes these per-GVR as its namespace fallbacks
+	Perms                    *ResourcePermissions
+	NamespaceScoped          bool   // True if at least one resource type ended up namespace-scoped
+	Namespace                string // The fallback namespace used for namespace-scoped probes
+	Scopes                   map[string]k8score.ResourceScope
+	ScopeNamespaces          map[string][]string // Per-kind namespace-scoped informer fanout; nil/empty means use Scopes[key].Namespace
+	ScopeCandidates          []string            // The candidate namespaces the probe walked — the dynamic CRD cache probes these per-GVR as its namespace fallbacks
+	ScopeCandidatesTruncated bool                // Additional candidates were omitted by MaxScopeCandidates
 }
 
 // Capabilities represents the features available based on RBAC permissions
 type Capabilities struct {
-	Exec           bool                     `json:"exec"`                    // Can create pods/exec (terminal feature)
-	LocalTerminal  bool                     `json:"localTerminal"`           // Local terminal available (not in-cluster, not disabled)
-	Logs           bool                     `json:"logs"`                    // Can get pods/log (log viewer)
-	PortForward    bool                     `json:"portForward"`             // Can create pods/portforward
-	Secrets        bool                     `json:"secrets"`                 // Can list secrets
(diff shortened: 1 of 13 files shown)


nadaverell:
@cursor review: Please focus on the final observation contract and authenticated namespace projection: synced is initial informer synchronization only, never continuous watch health; incomplete scope probes must not become definitive denials; preferred-namespace denial must not overwrite shared GVR evidence; empty viewer intersections must expose neither hidden namespaces nor watch metadata.

nadaverell:
Addressed all three Qodo findings in 11d023ef: empty namespace visibility now returns a fresh observation without retained probe evidence; api-resources has the connection guard and a disconnected-route regression; syncing metadata identifies the oldest unsynced informer. Full server tests and k8score race suite passed. The prior real mixed-RBAC validation remains documented in the PR; no fixture remains.

Issue #1629, opened 2026-09-04
GitOps tree can't show a Service's port info — health/sync always wins the subtitle


kcaashish:
## Problem

`GitOpsTreeGraph.tsx`'s `getSubtitle()` checks `node.sync || node.health` before falling back to `node.info`:

```tsx
if (node.sync || node.health) {
  return [node.sync, node.health].filter(Boolean).join(' • ')
}
if (node.info?.[0]?.value) return node.info[0].value
```

`pkg/topology/builder.go` gives every Service node `Status: StatusHealthy` unconditionally, and `pkg/gitops/tree/graph.go`'s `healthFromTopology` turns that into a non-empty `health` string for every Service in the tree. So `node.sync || node.health` is always true for a Service node, and `node.info` — the compact `"ClusterIP :80"` / `"ClusterIP :80 +1 more"` line computed by `infoFromTopology`'s `"Service"` case (#1604) — is never actually reachable in this view. A Service node in the GitOps resource tree always shows just its health ("Healthy") instead of what it exposes.

This isn't specific to Services — any kind whose health is always non-empty in this tree hits the same fallback-never-wins problem, since `info` is only shown when both `sync` and `health` are empty.

Verified against the `podinfo` Flux fixture in `scripts/gitops-demo`: the API (`/api/gitops/tree/...`) correctly returns `info: [{name: "Service", value: "ClusterIP :9898 +1 more"}]`, but the tree UI shows only `"Healthy"`.

## Goal

Make `info` visible for resources that have both a health/sync state and useful info content, without disturbing the primary sync/health status line the rest of the tree relies on.

## Suggested shape

- Append `info` after health/sync rather than treating it as a fallback, e.g. `"Synced • Healthy • ClusterIP :80 +1 more"`, or
- Surface `info` in a secondary location (tooltip, expanded row) instead of competing for the single subtitle line — similar to how the topology graph now shows the full port list on hover.

## Implementation notes

`getSubtitle` lives in `packages/k8s-ui/src/components/gitops/tree/GitOpsTreeGraph.tsx` (~line 700). `infoFromTopology`'s `"Service"` case is in `pkg/gitops/tree/graph.go` (~line 140). `node.info` is read in exactly one place in the whole frontend, so a fix is contained to `getSubtitle`.

PR #1630, opened 2026-09-04
fix(topology): expose all declared Service ports not just the first

### LINKED ISSUE #1604: Show all declared Service ports in topology without cluttering nodes
## Problem

Topology Service nodes currently display only the first entry in `spec.ports`. Both topology builders serialize a single `port` value from `spec.ports[0]`, so a Service such as `http:80` and `https:443` misleadingly appears to expose only `:80`.

## Goal

Make every declared Service port discoverable from the topology while keeping dense graphs compact and readable.

## Suggested shape

- Preserve a compact node subtitle for Services with several ports (for example, show a bounded summary rather than an unbounded list).
- Make the full declared set available on hover or another lightweight inspection affordance. Include port name when present, protocol, and Service port; showing targetPort is useful where it fits.
- Cover both resource and traffic topology builders.
- Keep the one-port presentation unsurprising, and avoid making node height grow with the number of ports.

## Implementation notes

The relevant Service-node construction currently lives twice in `pkg/topology/builder.go`, each taking `svc.Spec.Ports[0].Port`; the shared subtitle rendering is in `packages/k8s-ui/src/components/topology/K8sResourceNode.tsx`. Add focused coverage for the topology-node data and the compact/full display behavior.

kcaashish:
## Problem

Topology Service nodes displayed only `spec.ports[0]` — both the resources
and traffic topology builders (`pkg/topology/builder.go`) serialized a
single port, so a Service exposing e.g. `http:80` and `https:443` looked
like it only had `:80`. The GitOps resource tree's Service info line had
the same problem, reading the same trimmed field.

## Fix

- `pkg/topology/builder.go`: both builders now serialize every declared
  port via a shared `serviceTopologyPorts` helper into `Data["ports"]`
  (name, port, targetPort, protocol, appProtocol) instead of a single
  `Data["port"]`. `name`/`appProtocol` are omitted when unset — Port,
  TargetPort, and Protocol are always populated by apiserver defaulting
  by the time an informer sees them, so those stay unconditional.
- `pkg/gitops/tree/graph.go`: the GitOps tree's Service info line
  (`"ClusterIP :80"`) read the same removed `Data["port"]` field and
  silently dropped the port entirely once it was gone. Updated to the
  new `Data["ports"]` shape with the same first-port-plus-count summary.
- `packages/k8s-ui/.../K8sResourceNode.tsx`: the topology graph node
  subtitle now shows `ClusterIP :80` unchanged for a single port,
  `ClusterIP :80 +1 more` for multiple, and the full list on hover
  (reusing the existing `Tooltip` already used elsewhere in this file).
  Formatting (hiding `targetPort` when it matches `port`) mirrors
  `ServicePortCards` on the resource detail page, so a Service reads the
  same whether glanced at in the graph or opened in full.

## Tests

- `pkg/topology/builder_test.go`: pipeline coverage for both the
  resources and traffic builders exposing all ports, plus a focused unit
  test for the port-transform edge cases (unnamed port, `appProtocol`).
- `pkg/gitops/tree/graph_test.go`: new coverage for the Service info line
  (single- and multi-port) — previously only `Rollout` was covered here.
- `packages/k8s-ui/.../K8sResourceNode.test.ts`: subtitle test for
  0/1/many ports.
- Verified live against a real cluster (`kube-dns`, 3 ports) and the
  `podinfo` Flux fixture (`scripts/gitops-demo`, 2 ports) — both the
  topology graph and the GitOps tree API now report every declared port.

Fixes #1604.

## Related

While verifying the GitOps tree fix, found its Service info line is
actually unreachable in the UI today for an unrelated, pre-existing
reason (health/sync status always wins over the info line, and Service
nodes always carry a health status). Filed as #1629 — not fixed here.

---

> [!NOTE]
> **Low Risk**
> Display and topology payload shape change for Services only; backward-incompatible removal of `Data["port"]` but scoped to UI/GitOps summaries with test coverage.
> 
> **Overview**
> **Service topology nodes** no longer serialize only `spec.ports[0]`. Both resource and traffic builders now populate `Data["ports"]` via `serviceTopologyPorts` (port, targetPort, protocol, optional name/appProtocol), replacing the single `Data["port"]` field.
> 
> **GitOps tree** Service info lines use the same `ports` array and show `ClusterIP :80` or `ClusterIP :80 +N more` when multiple ports exist.
> 
> **Topology graph UI** (`K8sResourceNode`) subtitles follow that first-port-plus-count pattern; multi-port Services get a hover tooltip listing every port (aligned with `ServicePortCards` formatting). Tests cover builder serialization, GitOps info, and subtitle behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 728707df3aaba852da3ef5c87590814203d450b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/k8s-ui/src/components/topology/K8sResourceNode.test.ts
+++ b/packages/k8s-ui/src/components/topology/K8sResourceNode.test.ts
@@ -9,6 +9,15 @@ describe('baseSubtitle', () => {
     expect(baseSubtitle('ServiceMonitor', { endpointCount: 1 })).toBe('1 scrape endpoint')
   })
 
+  it('describes a Service by its first port, capping additional ports to a count', () => {
+    expect(baseSubtitle('Service', { type: 'ClusterIP', ports: [] })).toBe('ClusterIP')
+    expect(baseSubtitle('Service', { type: 'ClusterIP', ports: [{ port: 80 }] })).toBe('ClusterIP :80')
+    expect(baseSubtitle('Service', {
+      type: 'ClusterIP',
+      ports: [{ port: 80 }, { port: 443 }, { port: 9153 }],
+    })).toBe('ClusterIP :80 +2 more')
+  })
+
   it('describes a Calico staged deletion as a removal, not as selecting everything', () => {
     // A staged deletion carries no selector, because the Calico API forbids one.
     // Falling through to the empty-selector default would label the broadest
--- a/packages/k8s-ui/src/components/topology/K8sResourceNode.tsx
+++ b/packages/k8s-ui/src/components/topology/K8sResourceNode.tsx
@@ -168,6 +168,34 @@ function getIssueTooltip(issue: string | undefined): React.ReactNode {
   );
 }
 
+// Full port list for the Service subtitle's hover affordance — the subtitle
+// itself only ever shows the first port + a count, so this is the only place
+// a multi-port Service's other ports are visible without opening the detail
+// page. Formatting (hide targetPort when it matches port) mirrors
+// ServicePortCards in ServiceRenderer.tsx so a Service's ports read the same
+// whether glanced at in the graph or opened in the full resource view.
+function servicePortsTooltip(ports: ServicePortEntry[]): React.ReactNode | null {
+  if (ports.length < 2) return null;
+  return (
+    <div className="max-w-xs space-y-0.5">
+      {ports.map((p, i) => {
+        const target =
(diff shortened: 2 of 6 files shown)


nadaverell (approved):
LGTM, thanks for the fix!

PR #1646, opened 2026-09-06
fix(helm): persist and recover chart sources

StefanSa:
### Problem

Helm release storage retains chart metadata but not reliable original chart-source provenance. After restart, moving Radar between machines, or for older releases, Radar can therefore lose the repository/OCI source required to reconstruct the exact installed chart.

This is especially important for safe Values Preview/Apply because complete chart reconstruction must not depend on machine-local repository aliases.

### Fix

- Persist per-release chart-source provenance.
- For classic Helm repositories, retain both repository name and canonical URL.
- For OCI, retain the exact chart source without storing credentials.
- Recover existing releases from configured sources when unambiguous.
- Never guess between multiple matching sources.
- Support cross-machine recovery.
- Keep manual Helm repository and OCI recovery available.
- Add explicit opt-in ArtifactHub-assisted discovery for legacy unresolved releases.
- Treat ArtifactHub only as discovery metadata; independently verify the exact chart name/version/package/source before association.
- Improve Track Chart Source form clarity and recovery UX.
- Preserve fail-closed complete-chart reconstruction.

### Safety

- No credentials are persisted in release provenance.
- ArtifactHub lookup is explicit/opt-in and sends only the chart name.
- Ambiguous or unverifiable sources remain unresolved.
- Exact installed versions are required.
- Incomplete dependency trees remain fail-closed.

### Tests

- `go test ./internal/helm -count=1`
- Classic repository and OCI provenance/recovery, including cross-machine recovery and ambiguity handling.
- ArtifactHub HTTP/OCI discovery, privacy, independent exact-version verification, and dependency validation.
- Focused frontend source-recovery tests and frontend typecheck.
- Real existing-release Preview/recovery validation without unintended Helm revision changes.

### Dependency

Depends on #1631, which provides the fail-closed complete-chart reconstruction used by this recovery path.

### DIFF
--- a/internal/helm/artifacthub_recovery.go
+++ b/internal/helm/artifacthub_recovery.go
@@ -1,0 +1,395 @@
+package helm
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"slices"
+	"strings"
+	"time"
+
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/release"
+	"helm.sh/helm/v3/pkg/repo"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	artifactHubRecoveryPageSize   = 50
+	artifactHubRecoveryMaxResults = 250
+	artifactHubRecoveryTimeout    = 20 * time.Second
+)
+
+type artifactHubRecoverySearch func(context.Context, string, int, int, bool, bool, string) (*ArtifactHubSearchResult, error)
+type artifactHubRecoveryVerifier func(context.Context, *action.Configuration, *release.Release, ChartSourceCandidate) (ChartSourceCandidate, error)
+
+type artifactHubVerificationError struct {
+	kind string
+	err  error
+}
+
+func (e *artifactHubVerificationError) Error() string { return e.err.Error() }
+func (e *artifactHubVerificationError) Unwrap() error { return e.err }
+
+// DiscoverArtifactHubSources performs the optional, user-triggered final source
+// discovery step. ArtifactHub only supplies possible locations; every returned
+// candidate has independently yielded the exact complete installed chart.
+func (c *Client) DiscoverArtifactHubSources(ctx context.Context, namespace, name string) ([]ChartSourceCandidate, error) {
+	cfg, err := c.getActionConfig(namespace)
+	if err != nil {
+		return nil, err
+	}
+	return c.discoverArtifactHubSourcesWith(ctx, cfg, name, SearchArtifactHubContext, c.verifyArtifactHubSource)
+}
+
+func (c *Client) DiscoverArtifactHubSourcesAsUser(ctx context.Context, namespace, name, username string, groups []string) ([]ChartSourceCandidate, error) {
+	cfg, err := c.getActionConfigForUser(namespace, username, groups)
+	if err != nil {
+		return nil, err
+	}
+	return c.discoverArtifactHubSourcesWith(ctx, cfg, name, SearchArtifactHubContext, c.verifyArtifactHubSource)
+}
+
+func (c *Client) discoverArtifactHubSourcesWith(ctx context.Context, cfg *action.Configuration, releaseName string, search artifactHubRecoverySearch, verify artifactHubRecoveryVerifier) ([]ChartSourceCandidate, error) {
+	ctx, cancel := context.WithTimeout(ctx, artifactHubRecoveryTimeout)
+	defer cancel()
+
+	rel, err := action.NewGet(cfg).Run(releaseName)
+	if err != nil {
+		return nil, fmt.Errorf("get release for source discovery: %w", err)
+	}
+	if rel.Chart == nil || rel.Chart.Metadata == nil || rel.Chart.Metadata.Name == "" || rel.Chart.Metadata.Version == "" {
+		return nil, fmt.Errorf("release has no usable chart name and version")
+	}
+
+	// Recorded provenance remains authoritative while it is usable. If it has
+	// become unavailable, an explicit ArtifactHub request may suggest a
+	// replacement, but discovery never persists that replacement itself.
+	var unavailableRecorded *ChartSourceCandidate
+	if recorded, ok := chartSourceFromRelease(rel); ok && (recorded.Type == "oci" || recorded.URL != "") {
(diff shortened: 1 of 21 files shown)


hisco:
Thanks for this, and sorry it sat. I owed you an answer on the ArtifactHub question from #1631 before you put more work in, and it took me far too long. It's at the bottom.

The premise is right, and you got there from the same place we did. The comment we left on `storedChartMatchesReconstructed` when fixing up #1631 ends with "Closing it needs recorded provenance of where the release was installed from, not a better fingerprint of a chart Helm has already altered." That is this PR. We want it in Radar.

Helm release labels are the right mechanism too. Provenance lives in the cluster so it survives a restart and works from another machine, and `mergeCustomLabels` carries it forward when someone runs `helm upgrade` by hand. No second datastore. Good call.

**One correction on the diagnosis.** `configuredRepositoryForSource` returning nil is the check working: the recorded URL and the local alias disagree, so the match is refused. But `loadTargetChart` then sets `repositoryName = source.Reference` and resolves through that same alias anyway, so Radar pulls from the repo it just rejected. That's the failure this PR exists to prevent, and it matters more here than most places because #1631 is what happens when the wrong chart gets applied.

**I think it makes sense to split this up.** It's five changes together, which is 3.5k lines a review round, and that's a large part of why I took so long. Three PRs, roughly in this order:

- **OCI install.** `installWith`, `installWithProgressUsing` and `GetChartDetail` can't handle an `oci://` source at all on main, which your change fixes. It isn't in the description and has no tests. It's also the most self-contained of the five, so it's the easiest one to start with.
- **Chart source tracking**, including the new `POST /repositories` endpoint, since that's what it serves. The list below.
- **ArtifactHub discovery**, with the search error handling that goes with it. Read the last section before you start it. Three things about the shape need to change, so I'd hold off porting the current version across.

Rebase each onto current `main` as you open it. Two notes on that further down.

### Chart source PR

1. The alias fallback above. When `configuredRepositoryForSource` refuses the match, I think it should fail closed and tell the user to add the recorded repository, rather than fall back to the alias.
2. Chunk labels never shrink. Replacing a long source ref with a shorter one leaves the old `-1` behind, and the stitched value decodes to nothing, so provenance is silently lost. It fails closed, which is the right direction. Clearing the unused indexes needs a `"null"` value, which is the literal Helm treats as a delete, but only on the upgrade path and only for indexes the previous release actually had. `mergeCustomLabels` runs on upgrade only, so writing tombstones unconditionally would land sixteen literal `null` labels on the Secret for a fresh install. Your `mergeChartSourceLabels` already computes the right set for the direct-write path.
3. `recoverLegacyClassicSource` and its tests can go. `chartSourceLabels` rejects a repository candidate with an empty URL, so the name-only shape it recovers from is never written, and the tests build it by hand. I checked whether the shrink bug in item 2 could produce that shape by accident, across every shrink case I could construct, and it always ends as provenance lost instead. Deleting it also removes the problem that `checkForUpgrade` and `batchCheckUpgrades` are GET routes that write to release Secrets.
4. `ensureClassicRepository` holds `c.mu` across `DownloadIndexFile`. That's the same lock `buildActionConfig` takes on every Helm request, and the getter's default timeout is 120s, so one slow repo can block every Helm operation in the process for two minutes. The lock only needs to cover the `repositories.yaml` read-modify-write.
5. `handleAddRepository` is gated on `requireHelmWrite` alone, but it reaches `SetSource` whenever it gets a namespace and release name, and `PUT /releases/{ns}/{name}/source` also requires `requireCloudRole(auth.RoleMember)`. Same gate.
6. The label prefix should be `radarhq.io/`. You had no way to know this one. We migrated, and both `radar.skyhook.io/` and `skyhook.io/` are legacy read-only with a TODO to drop them this quarter, see `pkg/k8score/node_debug.go` and `pkg/gitops/operations.go`. Worth doing before it ships, since these land on release Secrets in live clusters and changing the prefix afterwards means a migration.
7. The dialog can't tell "recorded" from "recorded and reachable". `sourceCandidatesWith` returns the recorded source whenever it's a repository with a URL, without checking the repo exists on this machine. So on a new laptop `candidates.length` is 1, the ArtifactHub button never renders, and the Select next to the dead source fails with "selected source does not publish X version Y". That's the moved-machine case, which is the one the feature is for. The backend already handles it, keeping it as `unavailableRecorded` and continuing.
8. `AddOCISource` and `ensureClassicRepository` run before `preInstallCheck`, so an install we reject still leaves machine-wide config behind.
9. `UpgradeInfo.sourceCandidates` is populated by the backend and read by nothing.
10. `web/scripts/test-helm-repository-click.mjs` is in no npm script and no workflow. Either alongside `check-no-native-title.mjs` in `lint`, or drop it.

### Two rebase notes

Your branch carries its own copy of #1631, from before ac6c3213. We changed `validateChartDependencyBodies` there to stop consuming one dependency body per declaration: Helm vendors a single directory per chart name, so a chart declaring the same subchart twice under two aliases legitimately ships fewer bodies than declarations, and consuming rejects it as incomplete. Main's version is the one to keep.

CI is green now, both jobs. The runs had been sitting at `action_required` since the day you opened it, waiting on a maintainer to approve them, which was on us. The branch still conflicts with main though, so it needs the rebase either way.

### ArtifactHub

Yes to the feature. Radar already installs from ArtifactHub-named URLs, so the concept is fine. Three things about the current shape need changing.

**Verify on selection, not on search.** I measured the fan-out: one click is 29 exact-name matches for `nginx`, 59 for `redis`, 57 for `postgresql`. Each means an `index.yaml` download and then a chart tarball, serially, against a 20 second budget. Eight redis index files alone took me 3.7 seconds, and three of those failed fast, so the real per-repo cost is higher. On any common chart name this runs out of time, and when it does `ctx.Err()` returns an error that discards every candidate already verified. There's no cache, so clicking again repeats all of it. Searching, showing the candidates with publisher, URL and ArtifactHub's official and verified badges, and verifying only the one the user picks brings it to one index and one tarball. Your dialog already renders that list when there's more than one result, it just needs to be the input to verification rather than the output. That also puts the trust signals in front of a person, the way the chart browser already does. The search currently passes `official=false, verified=false` and drops them.

**Compare templates, not just name and version.** Name and version is the same weak signal that caused the original bug, and plenty of unrelated repos publish `nginx` `1.2.3`. `storedChartMatchesReconstructed` on main already does that comparison against the copy Helm stored in the release. Without it the dialog says "verified" and means "maybe".

**Don't send repository credentials to a host the index chose.** `loadExactHTTPChart` attaches the matched repo's basic auth to the chart download, and that URL comes from the repository index, which can point at any host. Your redirect handler already does the hostname check. The initial request needs the same one.

Keep the gating as it is: refuse when recorded provenance still works, refuse when a configured source already matches, and only then search. Moving verification later shouldn't move the search earlier.

The tests are good, by the way. The ambiguity cases, the auth versus infrastructure split, the safe-redirect check, fail-closed on an incomplete umbrella. That's usually the part we have to ask for.

PR #1674, opened 2026-09-07
Make informer seen/recreate-diff tracking API-group aware

nadaverell:
## Summary

Radar's shared informer cache tracks "have I already recorded this resource" and matches delete→recreate pairs (for a clean recreate diff) using a key of just `(cluster, kind, namespace, name)` — no API group. A CRD that shares a Kind name with a built-in or another CRD (e.g. a Volcano/Kueue-style `Job` outside the core `batch` group) can therefore have its seen-state or recreate-diff confused with an unrelated resource of the same name. Separately, a dynamic resource's delete event could reach Radar's timeline callback still wrapped in client-go's `DeletedFinalStateUnknown` tombstone instead of the actual object, silently dropping that delete from the timeline.

## What changed

- **Group-aware seen/recreate tracking.** The resource's API group is now part of the seen-tracking key. `EventStore.MarkResourceSeen` / `IsResourceSeen` / `ClearResourceSeen` (and the underlying `SeenResourceKey`) gain an `apiGroup` parameter, implemented consistently across all three store backends (in-memory, SQLite, Postgres). `internal/k8s/cache.go` resolves the exact `apiVersion` for both dynamic (unstructured) and typed informer objects — via a new `timelineAPIVersion` helper — before recording, so the group is always known at the point of use.
- **Fixed a dynamic-delete tombstone bug.** `pkg/k8score/dynamic_cache.go`'s `enqueueDynamicChange` was passing client-go's `DeletedFinalStateUnknown` wrapper straight to `OnChange` callbacks instead of unwrapping it, even though it had already resolved the real object's identity. Callbacks (including the timeline recorder) expect the resource object itself.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` — both Go modules (`.` and `pkg/`), full pass
- `gofmt -l` on all changed files — clean

### DIFF
--- a/internal/k8s/cache.go
+++ b/internal/k8s/cache.go
@@ -686,8 +686,15 @@ func recordToTimelineStore(clusterContext, kind, namespace, name, uid, op string
 		return
 	}
 
+	obj := newObj
+	if obj == nil {
+		obj = oldObj
+	}
+	apiVersion := timelineAPIVersion(kind, obj)
+	apiGroup := GroupFromAPIVersion(apiVersion)
+
 	if op == "add" {
-		if store.IsResourceSeen(clusterContext, kind, namespace, name) {
+		if store.IsResourceSeen(clusterContext, apiGroup, kind, namespace, name) {
 			timeline.RecordDrop(kind, namespace, name, timeline.DropReasonAlreadySeen, op, clusterContext)
 			if DebugEvents {
 				log.Printf("[DEBUG] Already seen, skipping: %s/%s/%s", kind, namespace, name)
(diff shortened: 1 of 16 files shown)


nadaverell:
Closing as a duplicate — this recreates the same fix already better-covered by the in-flight #1549 → #1553 → #1560 stack (including served-version tolerance, SQLite legacy-key migration, and typed-builtin group canonicalization, which this PR didn't handle). Splitting out the two genuinely novel pieces (dynamic-cache delete-tombstone bug, Postgres store support) separately.

PR #1678, opened 2026-09-07
Add shared Prometheus seams for availability, bounding and selectors

nadaverell:
## Summary

Three shared seams in `internal/prometheus` that the next metrics work (Prometheus nudge in investigations, vitals in `diagnose`, metrics authorization) builds on. No user-visible behavior change.

## What changed

- **Availability seam.** `prometheus.Availability(ctx)` (and the `Client` method) answers `connected` / `configured_failed` / `absent` with `Address` and `Err`, on top of `EnsureConnected` and bounded by the caller's context. It starts no work beyond what `EnsureConnected` already does. Discovery now returns an unexported error with the same message as `ErrPrometheusNotFound` (and `errors.Is`-compatible with it) when it enumerated a candidate it could not reach, so an unreachable install is reported as `configured_failed`, not `absent`. A timeout during a probe or rediscovery keeps the endpoint the call started with.
- **Bounding seam.** `maxPromResponseBytes` and `summarizeLargeResult` move out of the MCP tool as `prometheus.MaxResponseBytes` / `prometheus.SummarizeLargeResult`, byte-for-byte the same behavior and the same `RADAR_MCP_PROM_MAX_RESPONSE_BYTES` override. The MCP tool calls the moved functions; existing MCP tests pass unchanged.
- **Selector inventory.** `prometheus.ExtractSelectors(query)` tokenizes a PromQL query into its vector selectors (`{metric, matchers: [{label, op, value}]}`) and fails closed: string literals are skipped, every identifier not followed by `(` and not a keyword/aggregator/number/duration is a selector, `{...}` alone is a selector with an empty metric, and anything the tokenizer cannot account for returns `unknown=true`. It tracks operand state so grammar-legal keyword metric names (`by`, `or`, `sum`, ...) are recorded rather than dropped. `query_prometheus` now returns `selectors` (always an array) and `selectorsUnknown`.

## Testing

```
go test ./...                                  PASS (exit 0)
make tsc                                       PASS (exit 0)
make build                                     PASS ("Build complete: ./radar")
cd packages/k8s-ui && npm test                 PASS (172 files, 3549 passed, 1 skipped)
cd web && npm test                             2 failures in src/api/client.authRedirect.test.ts (5s timeouts)
git diff --check                               clean
```

The `web` failures are pre-existing: the same file fails on `main` (b54b3117) in this environment with 4 timeouts, and this PR touches no frontend file.

Visual test skipped: nothing rendered changes.

Cross-model review (Codex, two rounds, 9 findings): 8 fixed (keyword-named metrics, CR-terminated comments, postfix `by` clause operand state, numeric underscores, strict number/range literals, in-cluster unreachable classified as absent, timeout losing a known endpoint in two paths); 1 skipped: a Reset racing a cached probe can report the superseded address as connected, which is inherent to `EnsureConnected` and shared by every caller, and a consumer of this seam tolerates one stale answer.

## Scope and tradeoffs

- Tokenizer over a PromQL parser dependency, per the plan's default. Invalid PromQL may be over-reported (more selectors, never fewer), which only ever makes a chart "broader".
- The unreachable-install distinction is carried by an error value with an identical message so `internal/opencost` and `internal/server` callers matching `ErrPrometheusNotFound` are unaffected.

---

> [!NOTE]
> **Low Risk**
> Mostly additive MCP fields and internal refactors; discovery error wrapping preserves existing `ErrPrometheusNotFound` matching while the tokenizer intentionally over-reports scope when uncertain.
> 
> **Overview**
> Adds reusable Prometheus plumbing in `internal/prometheus` and surfaces **query scope metadata** on the MCP `query_prometheus` tool.
> 
> **`query_prometheus`** now always includes `selectors` (metric + label matchers per vector selector) and `selectorsUnknown` when PromQL cannot be tokenized safely (e.g. Grafana `$__interval`); docs updated accordingly. Oversized-result handling and the `RADAR_MCP_PROM_MAX_RESPONSE_BYTES` cap move to `prometheus.MaxResponseBytes` / `SummarizeLargeResult` with behavior unchanged.
> 
> **New `prometheus.Availability(ctx)`** classifies reachability as `connected`, `configured_failed`, or `absent` (with address/error), respecting caller context timeouts and preserving a known endpoint when probes or rediscovery outlive the context. Discovery returns a wrapped `errPrometheusUnreachable` (still `errors.Is` → `ErrPrometheusNotFound`) when candidates exist but cannot be reached, so unreachable installs classify as configured-failed rather than absent.
> 
> **`prometheus.ExtractSelectors`** is a fail-closed PromQL tokenizer (extensive unit tests); invalid or templated queries yield empty selectors + `selectorsUnknown`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fe0eb879b4fd280732348fc284eb2ede8a826c32. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
bugbot run

PR #1679, opened 2026-09-07
Capture workload vitals inside every diagnose

nadaverell:
## Summary

`diagnose` now returns a `metrics` field alongside the rest of the bundle: CPU, memory and restarts for Pod, Deployment, StatefulSet, DaemonSet and Argo Rollout targets, captured inside the same call over the pods the workload's selector matches.

The reason it is captured rather than left to the agent is that an agent which has to ask for resource data can silently never ask. A healthy verdict reached without looking at memory is indistinguishable, afterwards, from one where memory was fine — and the Findings pane cannot chart what the bundle does not carry, because the evidence ledger is built from the tool results the agent received. Capturing makes "did anyone check?" answerable, and gives the chart, the agent and the ledger the same numbers over the same window and the same pods.

## What changed

**The vitals branch** (`internal/mcp/tools_diagnose_metrics.go`). Runs concurrently with the rest of diagnose — launched before the bundle is assembled, joined immediately before the response is returned — and the three categories fan out within it, so all of this is one Prometheus round trip wide. Gated in order: a per-kind `get` SubjectAccessReview on the diagnosed resource (diagnose otherwise only checks namespace access before reading the shared cache), then availability. A denial omits the field without contacting Prometheus; an absent Prometheus omits it; a configured but unreachable one sets `error`.

**Shape.** `metrics: {window: {start, end, step}, pods, partial?, omittedPods?, series: [{category, unit, query, observedSamples, series}], error?}`. Each entry carries the PromQL that produced it, so an empty `series` with no `error` reads as "the window holds no samples", not "nothing was asked".

**Bounds.** Pod set cut to the first 50 by name (`partial` + `omittedPods` when cut). Step chosen for at most 60 points per series inside an 8 KiB sample budget, with a 24 KiB backstop over the whole field including query text. Values rounded to four significant digits. A 3s budget covers the branch including the availability probe. On a time or size breach, series are dropped and `error` names the limit; a single failing category is left out and named while the others stand.

**Flat stretches are compressed.** The interior of an evenly spaced run of equal values is dropped, keeping the points that open and close it. A workload sitting still costs two points per category instead of sixty. Each category reports `observedSamples`, the count Prometheus actually returned before compression — without it a two-point series is ambiguous between an hour of agreeing samples and two lonely ones, and `window`/`step` cannot settle it, because a range query returns nothing for an evaluation with no data. Values are compared after rounding, so movement below the four significant digits the field already reports counts as flat; a run must be evenly spaced to compress, so a gap in the series survives.

**One switch.** `DiagnoseMetricsEnabled` in `internal/mcp` is the single place the decision lives, so a later release can change the default by flipping it: the field is omitted and the rest of the bundle and `query_prometheus` are untouched.

**`pkg/prom`.** `BuildPodSetQuery` / `BuildPodSetQueryNoContainerFilter` aggregate one category over an explicit pod list as `sum(...)` — one series per category regardless of pod count, never one per pod. #1694 supersedes their call site with the ownership join; see below.

## What a reviewer should focus on

**Membership and the query form are both transitional here, and #1694 replaces both.** Two separate layers:

| | this PR | #1694 |
|---|---|---|
| which pods count as the workload's | the workload's **label selector** | **controller ownership** |
| how the query says it | enumerate every pod name | `kube_pod_owner` join on the workload |

On membership, this PR is an improvement over the `pod=~"<name>-.*"` prefix the older chart queries use — a prefix makes a Deployment named `api` absorb `api-worker`'s pods. It is not yet correct: a selector also matches bare pods, a sibling controller's pods during a Rollout `workloadRef` migration, and anything else sharing its labels. Ownership is what kube-state-metrics itself records, so it is also what makes history and the present agree on membership.

On the query form, enumeration is exact but does not scale: 525 B of query text at one pod, 2.2 KB at ten, 9.9 KB at fifty across the three categories — and `pkg/prom` issues queries by GET, so a wide workload puts ~3.3 KB into a URL per category, which some ingresses and auth proxies will refuse. The 50-pod cap is what bounds that. The ownership join names the workload once and is constant-length at any pod count.

No release happens between this PR and #1694, so neither the selector membership nor the enumerated query reaches a user. Reviewing this on the understanding that both are stepping stones is the right frame.

**Restarts use `sum(round(increase(...[1h])))`, not the `changes()` the older per-pod and workload queries use.** `changes()` counts sampled value transitions, so restarts landing between two scrapes — what crash-loop backoff produces in its first minute at 10s, 20s, 40s — collapse into one for the whole hour. `increase()` reads the counter delta and handles resets; its extrapolation yields fractions, so `round()` settles each pod's value before the sum. The older builders keep `changes()` because `RestartEventLane` (`RestartChart.tsx`) emits a marker on every positive delta between consecutive rolling counts and depends on integer semantics; changing them needs verification against a crash-looping cluster and is left for a follow-up. Both call sites now carry a comment saying which is which.

## Testing

`go test ./...` (both modules), `make tsc`, `make build`, `git diff --check` — all pass.

New tests cover the three series over the exact pod set within the byte budget against a step-faithful fake; `since` and `kind=pod`; the 55-pod set capped to 50 with lowered resolution; Prometheus absent, unreachable, and read-gate denied; the time and size budget breaches; a single failing category and a failed container-filter fallback; rounding and error bounding; the compression rule (full collapse, step boundaries, gap boundaries preserved, varying series untouched, NaN and ±Inf never collapsed, 0/1/2-point inputs); the end-to-end byte reduction; and the switch, including that turning it off leaves the rest of the bundle intact.

The fake Prometheus gains a `rangeDelay` so the time budget can be breached deterministically (slept outside the lock so concurrent callers are not serialized) and a `rangeBodyFunc` that renders the range body from the request parameters, which is what makes a step-faithful response possible. Fixtures that measure point counts or serialized size now vary their values above the reported precision — a flat fixture compresses to two points and would stop testing what it was for.

**Verified against a live GKE cluster with a real Prometheus**, driving the MCP `diagnose` tool directly:

- A crash-looping Deployment (874 restarts) and a three-pod StatefulSet returned all three categories with correct units, whole-number restarts, and the exact pod set in the query.
- Compression on a Deployment blocked by a missing Secret: the field went from 7.7 KB (50.4% of the bundle) to 1.2 KB (13.6%), all three categories collapsing to two points with `observedSamples: 56`. The crash-looping workload barely moved — its CPU and restarts genuinely vary — which is the intended shape.
- `observedSamples` earns its place there: the crash-looping workload's memory series has six samples, not fifty-six, because the pod keeps dying. Three retained points with `observedSamples: 6` is a sparse series; three with `observedSamples: 56` is a flat one.
- Latency: diagnose without the metrics branch returns in ~130 ms, with it in ~650–840 ms, and a single Prometheus range query takes ~620 ms. The branch is concurrent, but it is the critical path — nothing else diagnose does takes long enough to hide it. `docs/mcp.md` states this so the default can be revisited against numbers.

Visual test: skipped, Go only. The frontend adapter that turns this into a chart card follows on Track B.

The caps and failure paths — the 50-pod cut, the byte budgets, the 3s breach, an unreachable Prometheus — are covered against the fake rather than live, since reaching them means degrading a real cluster.

## Tradeoffs

- The 8 KiB budget covers samples and envelope only; the 24 KiB backstop is what bounds query text. Values are rounded to four significant digits so three series fit.
- Metrics from a partial pod set are marked `partial` with the omitted count, so the agent does not read them as the whole workload.
- Collecting on every diagnose costs one Prometheus round trip on the critical path. That is the price of the guarantee above; the switch is there for when it stops being worth paying.


nadaverell:
Retargeted to `main` so build and test CI run (workflows only trigger for PRs against main). This branch stacks on #1678 and should merge after it; the diff shrinks to this PR's own changes once #1678 lands.

PR #1680, opened 2026-09-07
Gate metrics access on the caller's RBAC

nadaverell:
## Summary

With auth enabled, every metrics surface ran with Radar's own Prometheus access rather than the caller's. A user limited to one namespace could read another namespace's charts, and with raw PromQL could read anything the Prometheus backend holds — including `kube-system` and node-level series. This gates each surface on the Kubernetes permission that corresponds to what it exposes. Local no-auth mode is unchanged.

The response-bounding and availability helpers this builds on are already on `main` (#1678).

## What changed

Two tiers. A curated surface is gated on the same read grant as the object it charts. A surface whose scope cannot be determined by inspection — raw PromQL, metric discovery, alert rules, the cluster aggregate — is gated on being able to list pods in every namespace, because a PromQL expression cannot be namespace-filtered by reading it.

| Surface | Gate |
|---|---|
| `GET /prometheus/resources/{kind}/{ns}/{name}` | `get` on that resource (group/resource derived from the kind) |
| `GET /prometheus/resources/Node/{name}`, and Node on the namespaced route | `get` on the Node, always cluster-scoped |
| `GET /prometheus/namespace/{ns}` | `list pods` in that namespace |
| `GET /prometheus/cluster`, `GET /prometheus/query`, MCP `query_prometheus` / `discover_metrics` / `get_prometheus_rules` | any-namespace `list pods`; 403 or tool error `cluster-wide metrics access requires permission to list pods across all namespaces` |
| `GET /prometheus/hpa/{ns}/{name}` (new) | `get` on the HPA; returns the current and desired replica series |

- `internal/prometheus/auth.go` — `canReadClusterWideMetrics` and `canReadMetricsResource`. An unmapped kind fails closed, and a cluster-scoped kind forces `namespace=""` so a namespaced RoleBinding cannot pass a Node read.
- `internal/prometheus/handlers.go` — the gate on every route; `handleRawQuery` now applies `MaxResponseBytes` / `SummarizeLargeResult` and returns `truncated` + `summary` the way the MCP tool already did, so an oversized result is never served as a partial one; `handleHPAMetrics` shares the range handling of resource metrics.
- `internal/server/server.go` — the inline gate closure becomes `Server.prometheusAuthGate`, which makes the two checks testable in isolation.
- `pkg/prom/queries.go` — `BuildHPAReplicasQueries` builds the two kube-state-metrics series server-side, with label sanitizing, instead of the client assembling PromQL.
- `web/src/api/client.ts`, `HPACharts.tsx` — the HPA chart reads the curated endpoint instead of raw PromQL. A 403 renders an explicit access banner and stops polling rather than showing an empty chart, which would have read as "this workload never scaled".
- `docs/mcp.md` and `CLAUDE.md` describe the two tiers.

**Worth a reviewer's attention:** the gate authorizes the *requested* kind, not the object the query resolves to. See the first tradeoff below.

## Testing

- `go test ./...` (both modules), `make tsc`, `make build` — pass.
- `packages/k8s-ui` — 174 files, 3,582 passed. `web` — 85 files, 912 passed.
- `internal/prometheus/handlers_auth_test.go` — every row of the matrix: no-gate passthrough, an authorized user (asserting the exact tuple the gate is asked for), and a namespace-scoped user (403 with the right message, Prometheus never queried). Plus the HPA endpoint's range/step handling and raw-query truncation.
- `internal/mcp/tools_prometheus_auth_test.go` — the three unbounded tools for no user, a namespace-scoped user (denied, no Prometheus probe), a cluster-wide user (allowed, one memoized SAR), and a SAR error (fails closed, not cached).
- `internal/server/prometheus_auth_test.go` — `prometheusAuthGate` against a fake apiserver: no-auth passthrough, a namespace-scoped user failing the all-namespaces review and cross-namespace/Node reads, an SAR-allowed namespace outside the allow-list refused, and an SAR error failing closed without caching.

Visual test: skipped. The 403 banner needs auth-enabled mode with a namespace-restricted user, which no available cluster is configured for; the banner is covered by `HPACharts.render.test.tsx` instead.

`web/src/api/client.authRedirect.test.ts` is flaky in this repo (a timing-sensitive 401 single-flight test, unrelated to this change). It passed in the run above and has failed in others; a rerun clears it.

## Tradeoffs

- **The gate authorizes the requested kind, but the query does not resolve one.** `buildWorkloadQuery` takes only namespace and name and matches `pod=~"<name>-.*"`, discarding the kind — so a user who can `get` Deployments but not StatefulSets in a namespace can request `Deployment/ns/db` and receive StatefulSet `db`'s pods. This is a real gap in the curated tier while it stands. It is closed by #1694 later in this series, which resolves a workload's pods by controller ownership and returns nothing for a kind/name pair that owns none; the gate itself does not need to change, because the query can then only return pods the requested object actually owns.
- **Object-scoped RBAC (`resourceNames`) is not consulted.** The gate carries group, resource, namespace and verb, the same shape every other read gate in Radar uses. It fails closed, so such a user sees less rather than more.
- **Radar queries Prometheus with its own credentials**; only the authorization is per user. Where several tenants share one Prometheus, that backend needs its own isolation.
- **Hosted consumers need an in-cluster Radar at or above this version for the HPA chart.** `@skyhook-io/radar-app` renders `HPACharts` inside `RadarApp`, and against an older agent the new `/prometheus/hpa` route returns 404, so the chart hides itself. There is deliberately no fallback to the raw endpoint, which an older agent would serve ungated. Release note, not a code path.
- Per-resource denials return the same `forbidden` body that PVC and rightsizing already use; only the cluster-wide tier carries the message naming the grant it needs.
- `usePromQLRange` was removed rather than deprecated: it was not part of the library's public surface in `web/src/index.ts` and had one caller.

---

> [!NOTE]
> **Medium Risk**
> Changes authorization for all Prometheus and related MCP entry points; mis-gating could block legitimate operators or leave a gap, though behavior is fail-closed with extensive test coverage.
> 
> **Overview**
> With auth enabled, Prometheus REST routes and MCP tools now enforce **two authorization tiers** instead of running every query with Radar’s backend credentials.
> 
> **Curated surfaces** (per-resource charts, namespace aggregates, new `GET /prometheus/hpa/{ns}/{name}`) require the same Kubernetes read as the object they expose—e.g. `get` on the workload/HPA/Node, or `list pods` in that namespace. **Unbounded surfaces**—`GET /prometheus/cluster`, `GET /prometheus/query`, and MCP `query_prometheus` / `discover_metrics` / `get_prometheus_rules`—require an any-namespace **`list pods`** SubjectAccessReview and return a message naming that grant on denial. Local no-auth mode still passes through.
> 
> `Server.prometheusAuthGate` centralizes SAR plus namespace allow-list checks (same semantics as the main resource API). Raw PromQL responses now use **`MaxResponseBytes` / `SummarizeLargeResult`** (`truncated` + `summary`) like the MCP tool. The HPA detail chart calls the curated HPA endpoint instead of client-built PromQL, shows an explicit **access banner** on 403, and stops polling denials.
> 
> Docs (`CLAUDE.md`, `docs/mcp.md`) and broad handler/MCP/server tests document and lock in the matrix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f15397fc5f8f098821d0cf3973163cfd0ae8b14a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell:
Retargeted to `main` so build and test CI run (workflows only trigger for PRs against main). This branch stacks on #1678 and should merge after it; the diff shrinks to this PR's own changes once #1678 lands.

PR #1682, opened 2026-09-08
Copy or download logs with a shared scope and format

### LINKED ISSUE #1617: Copy logs to clipboard
Often times, when debugging, what we really want is to copy/paste the logs into some chat.
Currently that is very difficult. Selecting the text is cumbersome and downloading the logs is again not very user friendly when all you want is to move fast.

Would really be great if there was a button next to the download to copy to clipboard directly.

roylibman:
Closes #1617.

## Summary

Getting logs out of Radar meant either dragging a selection across a virtualized viewer — which fights back the moment the list scrolls — or downloading a file, which is heavy when the destination is a Slack message. Neither carried your filters across: after narrowing to errors, the only things you could actually take with you were the whole buffer or whatever text you managed to highlight.

The log viewer's download menu is now an export popover. You choose how many lines and in what format, then copy to the clipboard or save a file — and both actions obey both choices. Filtering to errors and pasting them into an incident channel is one flow; so is saving exactly those errors as CSV for a ticket.

## What changed

**The export popover** (`LogCore.tsx`) has two rows — **Lines** (`Filtered N` / `All M`) and **Format** (Text / JSON / CSV) — above **Copy** and **Download**. Scope and format are orthogonal to destination, so they sit above both actions rather than being baked into four separate menu commands. The Lines row collapses to a plain count when the active filters exclude nothing, since a choice between two identical numbers is a choice about nothing.

**A live preview** renders the first line or two of the real payload. It is what makes the rest of the rule legible without a caption: everything an export carries beyond scope and format — timestamps, ANSI colour, level and search filters — already has a control on the toolbar, so instead of duplicating those as checkboxes (and creating a second source of truth for state the toolbar already owns), the popover shows where they land. Toggling timestamps changes the preview. It also gives JSON and CSV somewhere to show their shape: CSV previews its header beside a row, JSON a single compact object.

**Serialization moved to `utils/log-export.ts`**, which now backs copy, download and the preview from one place — the preview cannot drift from what you receive, and the two viewers no longer carry near-duplicate format builders. `LogCore`'s `onDownload` takes a finished `{content, mime, extension}` payload instead of a format string, leaving `LogsViewer` and `WorkloadLogsViewer` owning only filenames and delivery. The public props of both viewers are unchanged; the exported `DownloadFormat` type is kept as an alias.

**What an export contains** is worth calling out, since these differ from what the screen shows:

- **Timestamps are raw RFC3339**, not the viewer's display format. A pasted `14:23:07` — or `12s ago` — cannot be correlated against a dashboard or another service's logs an hour later. The toolbar toggle still decides whether they appear at all, and JSON and CSV keep the timestamp column regardless, where it is a named field a reader can ignore rather than the visual noise the toggle exists to suppress.
- **Pod attribution carries the container** (`[pod/container]`) in workload views, so multi-container workloads stay attributable. In JSON and CSV the pod is always present when the entry has one — there it is a field, not a rendering choice, and a display preference should not be able to make an export unattributable. CSV columns follow the data, so a row can never carry a field the header does not name.
- **CSV cells that open with `=`, `+`, `-` or `@` are prefixed with an apostrophe.** Excel and Sheets evaluate those as formulas and quoting does not stop it; every cell here holds whatever a pod chose to print.
- **ANSI escapes are stripped** from every format, so a colourized log (nginx, most Go and Rust loggers) no longer lands in a `.txt` as escape codes.
- **Content is otherwise raw** — full JSON, and stack continuations the viewer collapsed.

**Copy works over plain HTTP.** `navigator.clipboard` only exists in a secure context, and Radar is routinely served from a LAN address or a jumpbox, so copying falls back to the synchronous legacy path there. The per-line copy button and the empty-state recovery command take the same route.

**Scope is measured against the real buffer.** The workload viewer pod-filters before `LogCore` sees anything, so it also passes the unfiltered buffer as `allEntries`. Without that, `All N` would offer only the selected pods' lines, and the Lines row would vanish entirely when a pod filter was the only filter in play — stating a count as though nothing had been excluded.

## Testing

`make tsc` and the full `@skyhook-io/k8s-ui` suite (3,576 tests) pass. `utils/log-export.test.ts` adds 27 tests pinning the decisions above — raw timestamps, attribution fallbacks, ANSI stripping, CSV quoting, formula neutralization and header/column agreement across mixed entries, the JSON timestamp and pod columns, empty-input shapes — plus one that asserts the preview is a subset of the payload it previews, so the two cannot drift.

Exercised live on a two-pod Deployment and a single-container Pod, capturing the actual clipboard and blob payloads rather than reading the confirmation toasts: text copy carries raw nanosecond RFC3339 and `[pod/container]` with no ANSI; the download blob is byte-identical to it and arrives as `text/plain;charset=utf-8`; the JSON copy parses to an array with `timestamp, pod, container, content`; CSV emits a well-formed 4-column table in the workload viewer and a 3-column one for a single pod, where no entry carries a pod. Also checked the scope row reading `Filtered 0 / All 43` under a level filter with Copy disabled at zero, scope resetting to `Filtered` on reopen while the format preference persists, and Escape dismissing the popover without closing the drawer behind it.

**Reviewer focus.** `LogCore`'s `onDownload` signature changed, and `LogCore` is exported from the package — nothing consumes it directly today (the two viewers are the only call sites, and Radar Hub uses `LogsViewer`/`WorkloadLogsViewer`, whose props including `overrideDownload` are unchanged), but it warrants a major bump on the next `@skyhook-io/k8s-ui` publish.

## Notes

Scope resets to `Filtered` every time the popover opens rather than persisting: it describes the filters in play at the moment of asking, not a lasting intent, and a sticky `All` would silently widen a later export. Format does persist, since that is a standing preference.

Clearing the log buffer is untouched by this PR. Whether that button should confirm, or offer an undo on its toast, is a real question — the buffer holds streamed lines a refresh cannot re-fetch — but it is a separate decision about a separate control and does not belong in an export change.

The export popover dismisses on Escape in the capture phase and stops immediate propagation, because the host drawer also listens for Escape on `window` — a bubble-phase handler would dismiss the popover and close the drawer behind it in one keystroke.


nadaverell:
> **Superseded** — the design described below (a caption above the download items) was replaced by the scope + format popover. Kept for history; see the PR description for current behaviour.

Pushed review fixes as `2db1d0e9`.

**Copied timestamps are now raw RFC3339.** Copy was emitting `formatLogTimestamp(ts, tsFormat)`, so the default produced a bare `14:23:07` and the `relative` format produced `12s ago` — neither of which correlates against anything once it lands in a chat. That contradicted the PR's own "raw beats screen-literal" principle, and disagreed with the Download items directly below it, which have always written `l.timestamp`. The `showTimestamps` toggle stays as the on/off gate. Copy also now carries `[pod/container]` rather than `[pod]`, matching the TXT download, so multi-container workloads stay attributable.

**The formatting moved out of the component** into `formatEntriesForCopy` in `log-format.ts`, with 9 tests pinning the timestamp / attribution / ANSI / empty-field decisions — same shape as `rbac-blast-radius.ts`. It was untestable inline.

**The menu now states the download scope.** Copy honours the level and search filters, the downloads don't, and nothing said so — "filter to errors → Download TXT" silently hands back the whole buffer. Went with a caption above the download group rather than per-item tooltips: the failure mode is not knowing to ask, and a 500ms hover delay arrives too late for someone who has already filtered and is reaching to click. Menu widened to `w-72` so the caption sits on one line in the viewer's mono font.

**Clear confirms only when it needs to.** It was asking even when the dialog's own copy said "Refresh to load them again" — i.e. when the action was recoverable. It now asks once the buffer has held streamed lines (including after the stream is stopped), which are the ones a refresh genuinely cannot bring back, and the message says that. Window-level shortcuts are also suppressed while the dialog is open — `s` was toggling streaming behind it.

The copy item is disabled when nothing is visible instead of reporting an error toast after the click.

One note on the description: the "Also in this PR" paragraph still says the clear confirm is "skipped when the buffer is empty", which `122ec078` had already removed — worth a quick edit.

**Verified:** `make tsc` clean, full `k8s-ui` suite (3558 tests) green, and a live visual test against `kind-radar-kyverno-demo` — menu layout, the 46-line copy with its success toast, and the clear dialog.

nadaverell:
Disposition of the automated findings, all of which were raised against earlier commits on this branch.

**Addressed**

- *Clear confirm ignores hidden buffer lines* — correct, and the sharpest finding here. `WorkloadLogsViewer` passes pod-filtered entries while `onClear` wipes the whole buffer, so gating on `entries.length` meant deselecting every pod wiped a full buffer with no dialog. The clear-confirmation is no longer part of this PR at all — `Clear` matches `main`. The underlying confusion is fixed regardless: `LogCore` now takes an optional `allEntries` so anything reasoning about "the buffer" reads the buffer, which is also what makes the export's `All N` mean all.
- *Redundant menu comment* — gone.

**Not changing, with reasons**

- *Copy includes hidden stack lines* — deliberate. Stack continuations the viewer collapsed are included in full; the export is for a reader elsewhere, not a screenshot of the pane.
- *Copied pod names differ from the viewer* — deliberate, same reason. The viewer truncates pod names to fit a column; a truncated pod name is useless in a paste, so the export carries the full name (and the container alongside it).
- *Fallback copy loses keyboard focus* — not accurate against the code. `utils/clipboard.ts` restores the previously-focused element in its `finally`, guarded on the textarea still owning focus and using `preventScroll`.
- *Inconsistent menu shadow* — `shadow-lg` is what the timestamp-format and structured-mode menus in this same toolbar already use, and the count is unchanged from `main`. Switching only this one to the theme token would make it the inconsistent surface, not a consistent one. Worth doing across all three in a separate pass if the rule should hold here.

**Known tradeoff, flagged rather than hidden:** the CSV formula guard prefixes cells opening `=` `+` `-` `@` with an apostrophe. On real logs this fires on things like `====` banner separators — benign content that ends up as `'====` in the raw file. Spreadsheets consume that apostrophe on import, so the path CSV exists for renders correctly, and Text and JSON are untouched for anyone who needs byte fidelity.

nadaverell (approved):

PR #1710, opened 2026-09-09
Recognize cert-manager Secrets in the unused-resource audit

### LINKED ISSUE #1638: Audit reports cert-manager TLS Secrets as unused when consumed through reflector and Istio Gateway
## Describe the bug

The cluster audit reports cert-manager-managed TLS Secrets as unused even though they are actively used by an Istio Gateway through reflector-based cross-namespace synchronization.

## To reproduce

1. Deploy Radar 1.12.1 in a Kubernetes cluster with cert-manager, reflector, and Istio.
2. Create a `Certificate` whose `spec.secretName` is `example-api-tls` in an application namespace.
3. Configure the generated TLS Secret for reflector synchronization to `istio-system`.
4. Configure an Istio Gateway in `istio-system` with `spec.servers[].tls.credentialName: example-api-tls`.
5. Open the cluster audit and inspect the `Unused ConfigMap or Secret` findings.

## Expected behavior

The source TLS Secret should not be reported as unused when it is managed by cert-manager and configured for reflector synchronization to a namespace where an Istio Gateway consumes the reflected Secret.

At minimum, cert-manager-managed TLS Secrets should be recognized from their standard metadata regardless of whether the certificate identifier is represented as a label or annotation. Ideally, the audit should also account for the reflector source-to-copy relationship when evaluating cross-namespace Gateway usage.

## Actual behavior

The source Secret is reported as an unused Secret. The Certificate is Ready, the reflected Secret exists in `istio-system`, and the Istio Gateway uses that reflected Secret.

## Diagnostics (optional)

- Radar image: `ghcr.io/skyhook-io/radar:1.12.1`
- Kubernetes: EKS cluster
- Components involved: cert-manager, reflector, Istio Gateway
- No Secret values are included in this report.

## Additional context

The same audit run correctly identifies a separate, unreferenced Opaque Secret as unused, so disabling the entire audit check or excluding the application namespace would hide legitimate findings. No cluster Secret was deleted while investigating this issue.

siddiqueirshad:
Fixes #1638.

The unused-resource audit checked `cert-manager.io/certificate-name` only as a label, but cert-manager records it as an annotation. Recognize both forms and register Certificate output Secrets and keystore password Secret references through the existing controller-reference collector. This prevents the reported TLS source Secret from being flagged when its reflected copy is consumed in another namespace, including when Radar cannot see that namespace.

The metadata exemption is conservative: a Secret retaining certificate-name metadata after its Certificate is deleted remains excluded from orphan findings. Generic Reflector detection is deferred; it needs explicit source relationships and a way to distinguish missing consumers from incomplete visibility.

Validation: both Go audit test packages and the full application build passed. A live cert-manager test on radar-test-management verified issued TLS Secrets, metadata exemptions, Certificate output and keystore references, unrelated orphan controls, and orphan detection after a Certificate reference was removed. A namespace-only reader denied Certificate access still excluded annotated cert-manager Secrets; annotation-free output and keystore password Secrets remained warnings under that restricted visibility. The Checks page loaded without console errors. Generic Reflector/Istio behavior was not tested.

### DIFF
--- a/internal/audit/crd_config_refs.go
+++ b/internal/audit/crd_config_refs.go
@@ -144,6 +144,8 @@ func dynamicConfigRefHandlerFor(gvr schema.GroupVersionResource) dynamicConfigRe
 		}
 	case "cert-manager.io":
 		switch gvr.Resource {
+		case "certificates":
+			return certManagerCertificateConfigRefs
 		case "issuers":
 			return certManagerIssuerConfigRefs
 		case "clusterissuers":
@@ -313,6 +315,15 @@ func istioGatewayConfigRefs(u *unstructured.Unstructured) []bp.ConfigObjectRef {
 			addSecret(&refs, ns, name)
 		}
 	}
+	return refs
+}
+
+func certManagerCertificateConfigRefs(u *unstructured.Unstructured) []bp.ConfigObjectRef {
+	var refs []bp.ConfigObjectRef
+	ns := u.GetNamespace()
+	addSecret(&refs, ns, stringAt(u.Object, "spec", "secretName"))
+	addSecret(&refs, ns, stringAt(u.Object, "spec", "keystores", "pkcs12", "passwordSecretRef", "name"))
+	addSecret(&refs, ns, stringAt(u.Object, "spec", "keystores", "jks", "passwordSecretRef", "name"))
 	return refs
 }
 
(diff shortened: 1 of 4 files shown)


nadaverell (approved):
Thanks for the fix. The cert-manager changes address the reported case, including when Radar cannot see the destination namespace. I've prepared a small patch to keep this PR focused on those changes. Generic reflector support needs a separate pass so missing permissions don't turn into misleading unused-Secret warnings.

PR #1742, opened 2026-09-13
feat(resources): import multi-resource YAML from a file or drop target

### LINKED ISSUE #1580: Import multi-resource YAML from a file or drop target
## Problem

Radar already supports multi-document YAML apply, but users must paste manifests into the editor. There is no file-picker or drag-and-drop path, which makes applying a rendered manifest bundle unnecessarily awkward.

## Scope

Add a file-picker and drag-and-drop entry point that loads YAML into the existing resource-create/apply dialog. This should reuse the existing multi-document apply, schema, dry-run preview, conflict, and error-handling pipeline; do not create a parallel apply path.

## Acceptance criteria

- Users can select or drop one YAML/YML file from the existing create/apply UI.
- The file content is loaded into the existing editor and follows the normal preview/apply flow.
- A multi-document YAML file works as it does when pasted manually.
- Clear validation/error feedback for unreadable or empty files.
- Cap the apply request body before reading it: the current handler uses an unbounded `io.ReadAll`, which should not become a footgun once file upload is promoted.
- Add focused tests for file loading / drop handling.

## Pointers

- Existing apply UI and editor: `packages/k8s-ui/src/components/shared/CreateResourceDialog.tsx`
- Server apply and preview routes: `internal/server/server.go`, `internal/server/yaml_preview.go`

Please comment before starting so we can coordinate scope and UX.

kcaashish:
Radar already applied multi-document YAML, but only from text pasted into the editor. This adds a file picker and a drop target that feed that same editor, plus the request bound the issue asked for.

## What changed

An **Import YAML file** button beside a short hint, and the editor surface itself as the drop target — no extra panel to discover.

The file becomes editor text and nothing else. Client validation, schema loading, dry-run review, conflict handling and apply are untouched, so a multi-document file behaves exactly as a pasted one already did. There is no second apply path.

| | |
|---|---|
| Accepts | one `.yaml` / `.yml`, up to 6 MiB |
| Rejects | empty, unreadable, multi-file, wrong extension — through the banner the dialog already uses for apply failures |
| Decides on | the extension, never the MIME type |

**The extension, not the MIME type.** Browsers report nothing dependable for YAML: macOS Chrome commonly sends `""`, Windows sends assorted `octet-stream` variants. A `.yaml` reported as `application/octet-stream` is accepted; a `.txt` claiming `text/yaml` is refused.

**The drag overlay promises only what a drag can know.** `dragover` exposes an item's `kind` and the item count but withholds file names, so it highlights for any file drag and calls out a multi-file one, leaving the extension check for the drop. Detection reads `dataTransfer.types` rather than `items`, because Safari keeps `items` empty for the whole drag — the overlay never appeared there otherwise.

**Replacing the editor asks first only when there is work to lose.** The dialog opens prefilled in both entry points — a kind skeleton from `ResourcesView`, duplicate YAML from `WorkloadView` — so a non-empty editor is the normal state, not a sign the user typed anything. The confirmation is gated on the content differing from what the dialog itself put there: dropping onto an untouched skeleton, or into an empty editor, loads with no prompt.

## Server bounds

`handleApplyResource` read its body with an unbounded `io.ReadAll`. The limit is now one number — 6 MiB of YAML — enforced on every route, and stated on each in the terms that route actually carries:

  - **apply** caps the body directly, because there the body *is* the YAML.
  - **preview** checks the decoded `yaml` field. It never bounded the document at all before; its envelope cap only refused large files as a side effect,
  so the point it began failing moved with however much escaping a file happened to need.
  - **preview's envelope** is sized for worst-case escaping — twice the content limit, plus the wrapper's own bytes. (12.06 MiB) That is the ceiling the grammar allows: JSON doubles `"` and `\`, and YAML 1.2 admits no raw C0 control character in content except tab and newline, which also double. A ConfigMap holding minified JSON is close to half quotes, so a smaller envelope strands real files.
  - **apply** also caps at **100 documents**, the limit preview already enforced, and validates the request fully before opening a cluster client.

Oversize returns **413**, not 400: the cap fires before anything parses the body, so a 400 would blame content nobody has read yet. The convention is recorded alongside the existing status codes.

## Correcting myself on the document cap

I said on the issue that capping documents on apply "would reject content that pastes fine today." That was wrong. `YAMLReview: true` is hardcoded, so preview always runs first and already caps at 100 — nothing pasted in the UI is affected, only direct API callers. That is why it is included here rather than left open.

## Verification

Frontend: 3,668 tests, 23 covering the import module — extension over MIME, empty, oversize, unreadable, multi-file, multi-document passthrough, drag classification including Safari's shape, and confirm gating. Backend: handler tests for the 413 and the document cap, plus tests pinning apply's limits equal to preview's.

Against a live kind cluster: a dropped two-document file moves the editor status bar from "1 cluster schema active" to "2", and Review returns "2 accepted · 0 rejected · 0 not previewable" with admission-injected fields diffed — the existing pipeline, unchanged. A 7 MiB file is refused in 3 ms with the editor untouched.

## Known limitation

Loading a large file blocks the main thread while the editor parses the buffer — in Chrome roughly 90 ms at 128 KiB, 460 ms at 1 MiB, 2.4 s at 5.5 MiB. The import covers that by naming the file it is loading, painted before the editor starts and retired only once frames arrive on time again, so the status describes when the editor is usable rather than when the content was handed over.

**Safari is materially slower**, and at multi-megabyte sizes the pause is long enough to be a real problem. The cost is `YamlEditor` parsing the whole buffer two to three times per change (`parseFallbackYamlDiagnostics` → `parseYamlDocumentIdentities`, plus `parseAllDocuments`), which affects paste and the resource YAML editor equally — it is not introduced here. Happy to lower the import cap to whatever keeps the worst case tolerable, or open a separate issue for the parse cost, whichever you prefer.

There is also a ceiling above Radar's. The Kubernetes API server refuses request bodies over 3 MiB, and etcd's practical per-object limit is lower still — and preview's dry-run goes through the apiserver like any other write. So 6 MiB describes a *file*, not an object: a multi-document bundle reaches it fine, while a single object approaching it is refused by the cluster whatever Radar allows. That error surfaces unchanged rather than being masked.

## Type of change

- [x] Bug fix (non-breaking: preview's oversize status, Safari drag detection)

## How has this been tested?

## Related issues

Closes #1580.

---

> [!NOTE]
> **Medium Risk**
> Changes cluster YAML apply/preview HTTP handlers (body limits, status codes, document cap) that direct API callers may hit; UI import only affects the create dialog but loads full file content into the editor on the main thread.
> 
> **Overview**
> Adds **Import YAML file** and drag-and-drop on the create-resource dialog so multi-document manifests land in the same editor and flow through the existing preview/apply pipeline—no new apply path. Client checks enforce one `.yaml`/`.yml` up to **6 MiB** (extension-based), with a replace confirmation only when the buffer differs from the dialog’s initial skeleton, plus loading UI while the editor ingests large text.
> 
> On the server, **apply** and **preview** now cap request bodies **before** read (`readBoundedTextBody` / bounded JSON decode) and return **413** for oversize payloads instead of mislabeling size as malformed **400**. Shared constants align **6 MiB** YAML content, a preview envelope sized for worst-case JSON escaping, and a **100-document** cap on apply (matching preview). Apply validates document count and `reviewedVersions` indices before opening a cluster client. **CLAUDE.md** documents the **413** convention.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 197e57ac24dbcf606f8d596f4e9c3fbd72214c53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kcaashish:
Previews: 
image

image

file-a543e23508e34eefdae15f55f5088a4e

hisco:
Thanks, merging as-is. Keeping 6 MiB.

The parse cost is worth its own PR rather than a smaller cap. It's the two or three parses per change you pointed at, so it's more than moving a number, and a lower import cap would only hide it on one path.

hisco (approved):

PR #1746, opened 2026-09-13
Say when rightsizing could not read a workload's pods

### LINKED ISSUE #1726: Rightsizing reports an empty pod inventory when the cache could not answer
Split out of #1694 (merged as `3c4c4b78`), which widened this path rather than introducing it.

## What happens

`internal/prometheus/rightsizing.go` calls `k8s.WorkloadPods` and discards the error:

```go
pods, err := k8s.WorkloadPods(cache, kind, namespace, name)
if err != nil {
    // Rightsizing deliberately ignores the denial: its recommendation
    // comes from seven days of kube-state-metrics ownership, which does
    // not need the pod list, and the current-pod fallback simply finds
    // nothing. Nothing in the response distinguishes that from a workload
    // with no pods, which is a gap worth closing when the response gains
    // a field for it.
    return workload, nil
}
```

The comment is accurate and the reasoning holds for the *recommendation*, which comes from KSM history. What it silently loses is everything derived from the live pod list: `podNames`, and `currentPodOOM` — so a recommendation can omit "this workload's pods have been OOM-killed" and look identical to one where they have not.

## Why #1694 made it worse

Before that PR, `WorkloadPods` failed only on a denied lister. It now also returns:

- `ErrWorkloadCacheWarming` — an informer still filling its initial sync
- `ErrWorkloadAccessDenied` — a namespace an informer does not cover (`KindCoversNamespace` false)

Both take the same silent route. The warming case is the uncomfortable one: it is transient and common right after startup, so a rightsizing recommendation fetched in that window quietly drops OOM evidence and then silently starts including it a few seconds later.

## What it needs

A field on the response saying the live inventory could not be read, so the UI can qualify the recommendation instead of presenting a partial one as whole. `RightsizingStrip` and `RightsizingScanView` both consume this shape.

Worth deciding at the same time: whether a warming cache should be retried rather than reported, since unlike a denial it resolves on its own.

nadaverell:
Closes #1726.

Rightsizing read a workload's pods for two things — the pod names, and whether those pods have been OOM-killed — then discarded the error when the read failed. An empty answer was indistinguishable from a workload whose pods are healthy, so a recommendation could silently drop "these pods have been OOM-killed" and still present itself as complete.

## Availability is decided per workload

The question is never "can Radar list pods" but "can Radar list *this workload's* pods". Three different failures answer no, and they are not alike:

- **A warming informer** clears on its own within seconds, so reporting it would attach a caveat the next refresh removes. Both paths wait for it, bounded, and use the real pods. The single-workload read retries `WorkloadPods`; the scan waits once for the whole scan rather than per workload, and only for the informers it will actually read — a scan holding no Deployments never waits on the deferred ReplicaSet informer.
- **A denial** never clears and is reported.
- **A namespace the informer does not watch** is the subtle one. A synced, non-nil lister still answers only for its own namespaces, which is how Radar runs whenever cluster-wide list is denied. A successful empty list from an unwatched namespace proves nothing, so coverage is checked per workload — pods for every kind, and ReplicaSets as well for a Deployment, which reaches its pods only through them. A StatefulSet in the same scan is unaffected and stays unflagged.

Rows carry `liveInventoryUnavailable` — *unavailable* rather than *denied*, because an exhausted warming wait lands there too and is not a permission decision.

## What the reader sees

The caveat qualifies whichever explanation applies rather than standing in for it: a row can both be unable to check its pods and have a limit conflict, and hiding either behind the other loses something. It survives the staged-reduction path too — the one suggestion that asks you to reduce in steps is exactly where an unchecked pod matters most.

When there is no retained ownership *and* the pods cannot be listed, the response says so directly instead of reporting an absence of samples. The selection falls back to the current pods and then matches nothing by construction, so "no samples available" would send the reader hunting for missing metrics rather than an unreadable inventory. That reason is set server-side, so every consumer gets it rather than one UI.

## What does not change

The recommendation itself. Its safety gate is `OOMEvidenceAvailable`, computed from the Prometheus termination window, which never depended on the live pod read. `liveInventoryUnavailable` means `currentPodOOM` is **unknown** rather than false — a qualification, not a suppression.

A cancelled request stops rather than recording an unreadable inventory and continuing into Prometheus on a dead context.

## Testing

**Unit.** Go tests drive the scan against a real `ResourceCache` over a fake clientset: pods readable, pod lister denied, ReplicaSet ownership unreadable (Deployments flagged, StatefulSets not), and a namespace-scoped `ResourceScopes` cache where one workload sits inside the watched namespace and one outside. A further test pins that a DaemonSet-only scan does not spend the warming budget on ReplicaSets. Frontend tests cover the caveat composing with a suppression reason and surviving a staged reduction.

`go build ./...`, `go vet`, `go test ./internal/prometheus/` (including `-race`), `tsc --noEmit`, and the rightsizing suites pass.

**Live, against a GKE cluster with real Prometheus and kube-state-metrics.** Binaries built from this branch and from `main` were run against the same cluster and the same Prometheus, and their responses compared:

- Single workload (`Deployment staging/billing`): identical — `ownerCoverage: ksm_history`, 2 rows, `fit=oversized`, matching OOM-evidence flags. `liveInventoryUnavailable` is absent, which is correct for a readable inventory.
- Cluster-wide scan: 26 workloads discovered and evaluated on both, 25 with data. Across **52 rows, every decision field is identical** — `fit`, `recommendationReason`, `hpaManaged`, `hpaEvidenceAvailable`, `oomEvidenceAvailable`, `currentPodOOM`, `windowOomEvidence`, `limitConflict`, `confidence`, `reductionLimited`, `recommendedRequest`. The only payload differences were measured values (observed, peak, throttle ratio, sample counts) drifting between two runs seconds apart.

That exercises the per-workload coverage check, the ReplicaSet ownership path and the warming waits against real informers.

**Live, under a genuinely denied pod inventory.** A throwaway ServiceAccount was granted everything rightsizing needs except `pods` — `kubectl auth can-i list pods` returned `no`, `deployments` and `replicasets` returned `yes` — and both builds were run under it against the same cluster:

| | `main` | this branch |
|---|---|---|
| Single workload, `currentPodOOM` | `false` | `false` |
| Single workload, `liveInventoryUnavailable` | *field absent* | **`true`** |
| Cluster scan | 100 workloads · 316 rows · **0 flagged** | 100 workloads · 316 rows · **316 flagged** |

On `main` the response is indistinguishable from a workload whose pods have simply never been OOM-killed. The recommendation is unchanged between the two builds — `fit=oversized`, same `ownerCoverage: ksm_history`, same row and workload counts — which is the point: the pod read was never what the recommendation rested on, only what the OOM check needed. The ServiceAccount, ClusterRole, ClusterRoleBinding and namespace were deleted afterwards and verified gone.

Visual-test skipped: the UI delta is copy inside existing components, no new surface.

## Follow-up

The same class of bug exists on the HPA paths, where it is worse: `HPAEvidenceAvailable` gates whether a reduction is suppressed at all, so an uncovered namespace can let one through on a workload whose autoscaler was never seen. It is pre-existing and outside these lines, tracked in #1749.

Both rightsizing components predate the current Prettier config and are not formatted to it on `main`; this change matches each file's existing style rather than reformatting ~1,000 unrelated lines.

https://claude.ai/code/session_0125QDX24TNuod7FKLwTDhvW


nadaverell:
On the remaining cross-repo item ("Cloud users miss inventory caveat") — accurate, but not actionable here, so it is deliberately not addressed in this PR.

Radar Hub builds its embedded `RadarApp` from the published `@skyhook-io/radar-app` package, so *any* frontend change in this repo reaches cloud users only after that package is published and Hub's lockfile moves. That is the standing condition for every UI change, not something this branch introduces, and publishing is gated on explicit approval naming the version.

The backend field is additive and optional (`liveInventoryDenied?`), so an older embedded build ignores it and renders exactly as it does today. Nothing breaks in the interim; the caveat simply arrives with the next package bump.

Also worth noting for anyone reading the checks: **Cursor Bugbot reported `NEUTRAL` (skipped)** on both pushes here, so Qodo was the only automated reviewer on this PR.

PR #1793, opened 2026-09-17
desktop: capture ANTHROPIC_* and CLAUDE_CODE_* env vars via prefix

feiming:
## Summary
- The desktop app's login-shell enrichment only forwarded a fixed, enumerated list of known env vars, so any Claude/Anthropic var set via `~/.clauderc` or similar (beyond the hardcoded list) was silently dropped.
- Switches `getShellEnv` to capture all `ANTHROPIC_*` and `CLAUDE_CODE_*` vars by prefix instead of enumerating each one, so new vars are picked up automatically.
- Also switches shell env parsing from positional lines to `KEY=VALUE` pairs parsed via `env | grep -E`, which is more robust and naturally supports prefix wildcards.

## Test plan
- [ ] Verify desktop app launches and forwards `ANTHROPIC_*`/`CLAUDE_CODE_*` env vars set in the login shell into the app's environment
- [ ] Confirm previously-supported env vars still pass through correctly

---

> [!NOTE]
> **Medium Risk**
> Changes how the desktop process inherits shell environment (including API-key-style vars) and replaces env parsing logic, so regressions could affect kube/cloud/Claude tooling at launch.
> 
> **Overview**
> **Expands desktop login-shell env enrichment** so Claude/Anthropic settings from `~/.zshrc` and similar are no longer dropped when the app is launched from the GUI.
> 
> Adds **`AWS_REGION`** to the fixed capture list and introduces **prefix capture** for `ANTHROPIC_*` and `CLAUDE_CODE_*`, applying every captured name (not only the hardcoded list) when the process env is empty. **`getShellEnv`** is reworked into a **two-pass** flow: discover variable names from `env` (prefix matches included, with a regex that ignores bash function exports), then re-query the shell for each value using **control-byte record separators** so multiline values and noisy rc-file output are less likely to corrupt parsing. Shell invocation is centralized in **`runLoginShell`**, and **KUBECONFIG** skip/warning diagnostics now compare against the **pre-enrichment** process value.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f89299f026a6949576c31e8e163865ddbbaf1c44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell (approved):
LGTM - thanks for the PR! made one small change to avoid duplicate env capture one after the other, and added tests. but the core was solid already - great work.

PR #1795, opened 2026-09-17
Feature/kamalesh hpa fixes

### LINKED ISSUE #1756: HPA evidence is rejected while any unrelated deferred informer is warming
Surfaced by a cross-model review of #1755, and deliberately left out of it: fixing this changes full-access behaviour, which that PR was required not to touch.

## The problem

Both rightsizing paths gate HPA evidence on `cache.IsDeferredSynced()`:

- `internal/prometheus/rightsizing.go` — `loadHPAManagedResources`
- `internal/prometheus/rightsizing_scan.go` — `enrichScanHPA`

That flag is **global to the deferred phase**. It reports false while *any* deferred informer is still warming, and stays false permanently if any of them failed — even when the HPA informer itself is fully synced, cluster-wide, and perfectly readable.

## Why it matters

`HPAEvidenceAvailable` gates whether a reduction may be suggested at all. A false negative here means Radar withholds a correct, useful recommendation and tells the operator it "could not verify whether autoscaling depends on this request" — when it could have, and the obstacle was an unrelated kind.

This is the mirror of the bug #1755 fixed. That one let reductions through on unseen autoscalers; this one blocks valid reductions for an unrelated reason. Both come from treating a coarse signal as an answer about one specific read.

## Shape of the fix

`ResourceCache.InformerSynced(key)` already answers the precise question — and its doc comment calls out this exact hazard. Gating on the HPA informer's own sync state rather than the whole deferred phase is the change; #1755 does the same for pods and ReplicaSets in the scan, so there is a pattern to follow.

The other three callers of `IsDeferredSynced()` were checked and are **not** affected — `internal/server/dashboard.go` sets a `DeferredLoading` UI flag, and `pkg/k8score/cache.go` uses it twice to label a sync *phase* (`SyncPhaseDeferred`). All three report progress; none gates whether data is trustworthy. So the change is confined to the two rightsizing call sites.

## Before you start

**`InformerSynced` is safe to gate on here.** It returns `(synced, known)`, and a key it has never heard of comes back `known: false` — which would be a hole if deferred kinds were registered late. They are not: `cache.go` marks an existing entry with `informerStatuses[i].Deferred = true` when the deferred phase begins, so a watched HPA informer is `known` from the start and `synced` tracks the informer's own `HasSynced`. Decide deliberately what an *unknown* key should mean, though — `requireSynced` in `internal/k8s/workload_pods.go` treats unknown as fine, because reaching a lister at all implies the kind is watched; the same reasoning applies here, with `HorizontalPodAutoscalers() == nil` already covering the not-watched case.

**`deferredFailed` is never cleared.** Three `Store(true)` sites, no `Store(false)`. One deferred informer that times out therefore suppresses HPA evidence for the entire process lifetime, not for a warm-up window — which is why this is worth fixing rather than waiting out. The timeout path logs "These resources will return 403 to API consumers", so affected clusters are identifiable from logs.

**This loosens a safety gate.** #1755 tightened it; this widens when a reduction may be recommended. On a cluster with a failed deferred sync the change flips rightsizing from *never* suggesting reductions to suggesting them, which is a large, user-visible behavioural change — worth its own before/after evidence rather than a "looks right" review.

## Gotchas in the tests

Both of these cost time during #1755:

- **`IsDeferredSynced()` is false until the deferred phase closes its channel**, even in a test cache with no deferred types. A test that builds a `ResourceCache` and immediately calls `enrichScanHPA` gets an early return and silently asserts nothing. Poll until it reports true before exercising the code.
- **`scanScopeNamespaces(nil)` returns an empty, non-nil slice**, which routes `enrichScanHPA` into its per-namespace branch over zero namespaces — so passing `nil` scopes marks nothing and the test passes vacuously. The cluster-wide branch needs an explicit nil entry: `map[string][]string{"Deployment": nil}`.

`scopeTestCache` in `internal/prometheus/podscope_test.go` builds a real `ResourceCache` over a fake clientset, and accepts `ResourceScopes` for namespace-scoped informers — #1755's HPA tests use it and are the closest model to copy.

Kamalesh-Seervi:
## Description

Fixes #1756

## Problem

Both rightsizing paths gated HPA evidence on `cache.IsDeferredSynced()`:

- `internal/prometheus/rightsizing.go` — `loadHPAManagedResources`
- `internal/prometheus/rightsizing_scan.go` — `enrichScanHPA`

That flag is global to the deferred phase. It reports false while *any* deferred
informer is warming, and — because `deferredFailed` has three `Store(true)` sites
and no `Store(false)` — stays false for the rest of the process lifetime once one
of them times out. The HPA informer itself can be fully synced, cluster-wide and
perfectly readable the whole time.

`HPAEvidenceAvailable` gates whether a reduction may be suggested at all, so a
false negative means Radar withholds a correct recommendation and tells the
operator it "could not verify whether autoscaling depends on this request" — when
it could have, and the obstacle was an unrelated kind.

This is the mirror of the bug #1755 fixed. That one let reductions through on
unseen autoscalers; this one blocked valid reductions for an unrelated reason.

## Change

Both call sites now gate on `waitForInformerSynced(ctx, cache, "horizontalpodautoscalers")`,
which reads the HPA informer's own `HasSynced` via `ResourceCache.InformerSynced`
and waits out a warming cache within the existing `warmingRetryBudget`. This is
the same helper #1755 introduced for pods and ReplicaSets in the scan.

`loadHPAManagedResources` and `enrichScanHPA` take a `ctx` so they can share that
helper; both callers already had one.

**Unknown informer keys are treated as fine.** `InformerSynced` returns
`known: false` for a kind this cache never started, and `waitForInformerSynced`
reports that as synced — the same reasoning `requireSynced` in
`internal/k8s/workload_pods.go` uses. The not-watched case is already covered by
the `HorizontalPodAutoscalers() == nil` check, which now runs first so a
disabled kind returns immediately instead of entering the wait.

The other three `IsDeferredSynced()` callers are unaffected: `internal/server/dashboard.go`
sets a `DeferredLoading` UI flag and `pkg/k8score/cache.go` uses it twice to label
a sync *phase*. All three report progress; none gates whether data is trustworthy.

## Behavioural impact

This **loosens** a safety gate. On a cluster with a failed deferred sync,
rightsizing goes from never suggesting reductions to suggesting them — a
large, user-visible change, so it gets before/after evidence rather than a
"looks right" review.

Coverage semantics are untouched: `KindCoversNamespace` still decides per
workload whether an empty HPA list is an answer or a miss, so a namespace-scoped
informer still cannot claim it ruled out an autoscaler it never looked for.

## Tests

`stalledDeferredCache` builds a real `ResourceCache` where HPAs are deferred and
synced while an unrelated deferred kind (Jobs, via `DebugSyncDelays`) never
starts — the shape of a cluster whose deferred sync timed out. It asserts
`IsDeferredSynced()` is false, so the test cannot pass vacuously.

- `TestEnrichScanHPAIgnoresUnrelatedDeferredInformers` — scan path
- `TestLoadHPAManagedResourcesIgnoresUnrelatedDeferredInformers` — single-workload path
- `TestLoadHPAManagedResourcesReportsAnUnwatchedKind` — pins the unknown-key decision

Both regression tests **fail** against the old `IsDeferredSynced()` gate
(verified by temporarily restoring it) and pass with the fix.

Also removed the now-unnecessary `IsDeferredSynced()` poll from
`TestEnrichScanHPAHonoursCoverageEndToEnd` — the gate it was waiting on is gone.

## Verification

- `go build ./...`
- `go vet ./internal/prometheus/`
- `go test ./internal/prometheus/... ./internal/k8s/... -count=1`

No UI surface changed, so no visual test.

## Type of change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

## How has this been tested?

Describe the tests you ran to verify your changes.

- [ ] Tested locally with minikube/kind
- [ ] Added/updated unit tests

## Checklist

- [ ] My code follows the project's coding standards
- [ ] I have performed a self-review of my code
- [ ] I have added comments where necessary
- [ ] My changes generate no new warnings

## Related issues

Fixes #(issue number)

---

> [!NOTE]
> **Medium Risk**
> Changes when rightsizing may suggest CPU/memory reductions by altering HPA evidence gating; behavior improves when unrelated deferred informers stall but could surface reductions that were previously withheld.
> 
> **Overview**
> Rightsizing was treating HPA autoscaler evidence as unavailable whenever the **entire deferred informer phase** had not finished, even if the HPA informer itself was already synced. That blocked valid downsize guidance (`hpa_evidence_unavailable`) when an unrelated deferred kind was still warming or permanently stuck.
> 
> **`loadHPAManagedResources`** and **`enrichScanHPA`** now take a **`context`** and gate on **`waitForInformerSynced`** for the HorizontalPodAutoscaler informer only, not **`IsDeferredSynced()`**. Namespace coverage checks are unchanged.
> 
> Tests add a **`stalledDeferredCache`** (synced HPA + stuck Jobs deferred informer) and assert both the scan and single-workload paths still set **`hpaAvailable`** and **`hpaManaged`** when HPAs are readable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32174ba9ea30013b0c908ef29d2b6b0e94320feb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nadaverell (approved):
LGTM, merging - thanks!

PR #1808, opened 2026-09-17
Logs: show why the stream failed, not just that it failed

hisco:
## What a reader sees today

Open the Logs tab on a pod whose container has never started, and Radar says:

> Log stream connection failed

Nothing is wrong with the connection. The kubelet has no logs for a container
that never ran, and the pod log handler already says exactly that: it sends an
SSE `error` event carrying a reason like `waiting to start: trying and failing
to pull image`.

The client parsed that reason, wrote it to the browser console, and then put a
hardcoded string on screen instead. So an image the cluster cannot pull reads
as a broken Radar connection, and the reader goes looking at the network.

## The change

`handleSSEError` returns the reason it was already parsing, and `useLogStream`
shows it.

The old wording stays as the fallback for the two cases where it is accurate:

- a genuinely dropped connection, where EventSource fires a bare event with no
  body at all
- a refusal before the handshake, such as a denied namespace, which is an
  ordinary non-2xx response whose body EventSource never exposes

## Where to look hardest

The listener wiring moved out of the hook into `attachLogStreamListeners`. That
is most of the diff, and it is a move rather than a rewrite.

The reason for moving it is that the orderings are what actually matter here
and none of them were reachable from a test while they lived inside the hook:

- current Radar flushes `connected` before it opens the log stream, so the
  refusal arrives after the handshake
- older Radar never sends `connected` on that path, so the refusal is the first
  event
- the browser fires the same `error` listener again on the normal close that
  follows a clean `end`, which must stay silent
- a container switch supersedes an EventSource that can still fire late

Each of those is now a test. Reverting the one-line fix fails two of them.

## Compatibility

`handleSSEError` is exported from the package. Adding a return value to a
function that returned `void`, and making its third parameter optional, are both
source-compatible: existing call sites compile unchanged, and the new signature
is still assignable to the old type. Checked with `tsc --strict`.

## Tests

`packages/k8s-ui`: 3793 pass. `web`: 1381 pass. No new type errors.

---

> [!NOTE]
> **Low Risk**
> User-facing logs error copy and a refactor of SSE listener wiring; behavior is covered by new unit tests with no auth or data-handling changes.
> 
> **Overview**
> When pod log streaming fails with a server-provided reason (e.g. container waiting on **ImagePullBackOff**), the Logs UI now shows that message instead of the generic **"Log stream connection failed"**.
> 
> **`handleSSEError`** now **returns** the user-facing string (parsed from SSE `error` / `message` JSON, with the transport fallback when the event has no body). **`useLogStream`** sets **`streamError`** from that return value rather than always using the fallback label.
> 
> SSE listener wiring is **extracted** to **`attachLogStreamListeners`** so event **ordering** can be unit-tested: failure before/after **`connected`**, silent **`error`** after a clean **`end`**, and ignored late errors from superseded streams. New tests cover **`handleSSEError`** and the listener module.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df05ccb11a9a3f23afef1f3a72c2ac0c760c84e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hisco:

PR #1809, opened 2026-09-17
Logs: say when the stream is no longer live

hisco:
Stacked on #1808. Base it there so the diff stays to this change; retarget to
`main` once that merges.

## What a reader sees today

Open a pod's logs, watch them stream, then have the pod go away. Nothing on
screen changes except that the lines stop and the Stop control disappears.

That looks exactly like a workload that went quiet. It is the wrong conclusion
to reach while debugging one, and there is no way to tell the two apart.

## Reproduced

Against a kind cluster, with a pod printing a line a second:

1. open the pod's Logs tab, confirm lines arriving
2. `kubectl delete pod <name> --grace-period=0 --force`
3. the last 25 lines stay frozen on screen, no message, nothing in the toolbar

## Two paths, neither surfaced

**A failure after lines had arrived** was dropped. Both viewers gated the error
on an empty buffer:

```
errorMessage={fetchError || (entries.length === 0 ? streamError : null)}
```

**An ordinary end was never reported at all.** This is the one the repro above
hits, and it is why my first attempt at this did not work: `handlePodLogsStream`
answers a terminated pod with `event: end`, not `event: error`. `endedRef` then
makes the error listener return early, which is correct, so there was nothing
for the viewer to show even in principle. The hook now carries the end through
to the viewer.

## Where it goes

With lines on screen, above them rather than instead of them. Replacing them
would throw away the last thing the workload said, which is usually the reason
the reader opened the tab. With an empty buffer it keeps the body, since there
is nothing else to show.

The server's end reason is often the bare marker `stream ended`, which just
repeats the headline, so reasons that add nothing are dropped rather than
printed twice. A real one ("container exited with code 137") is kept.

## Where to look hardest

`placeStreamNotice` decides tone and placement and is the whole of the logic;
everything else is wiring. Note that a failure outranks an end when both are
set, because the error listener also fires on the close that follows a clean
end, and a real failure must not be downgraded to "ended".

## Tests

`packages/k8s-ui`: 3802 pass, 13 of them new. No new type errors.
Verified in the running app on a kind cluster, before and after.

---

> [!NOTE]
> **Low Risk**
> UI-only logging viewer changes with dedicated unit tests; no auth, API, or data-path changes beyond displaying existing stream end/error state.
> 
> **Overview**
> When a log stream stops—clean server `end` or a real failure—the UI now tells the user instead of leaving frozen lines that look like a quiet workload.
> 
> The stream hook and SSE listeners expose **`streamEnded`** (reason from the `end` event) separately from **`streamError`**. New **`placeStreamNotice`** picks copy and placement: failures beat ends when both are set; redundant end reasons like `stream ended` are dropped; with an empty buffer the message fills the log body, and with lines already shown it appears as a **banner above the logs** via a new `notice` prop on `LogCore`. Pod and workload viewers no longer hide stream failures once lines have arrived.
> 
> Vitest coverage was added for notice placement and headlines.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0c72fd6b35b524387a11334ec69a91068b2d4ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hisco:

PR #1814, opened 2026-09-18
Logs: tell the truth about which run you are reading

hisco:
Stacked on #1809, which is stacked on #1808. Draft until both land; retarget
down the chain as they merge.

Three findings from a product review of the logs surface, all walked on a live
kind cluster with a real crashlooping pod.

## 1. The Previous checkbox showed the current run

The worst of the three, because it does not show too little, it shows the wrong
thing confidently.

Ticking Previous fetched the previous run, and then the auto-stream re-armed,
cleared the buffer and refilled it from the container running now. The checkbox
stayed ticked. An operator investigating a crash read the live run believing it
was the one that died.

Proved by giving each run a unique id and toggling the box:

```
without the fix   unticked: RUN-eb9c53c4     ticked: RUN-eb9c53c4   <- same run
with the fix      unticked: RUN-da859e57     ticked: (previous run)
```

My first attempt at evidence used a fixture that printed identical text on
every run, so it could not have distinguished the two. That is what the run ids
are for.

Following a container and reading a run that already exited are mutually
exclusive, so Previous now suppresses the auto-stream.

## 2. The kubelet's apology reached the reader as a log line

When the kubelet has already collected a container's log file, the apiserver
answers **200** and the body is its own sentence:

```
unable to retrieve container logs for containerd://6e2396b58f81ea5e…
```

Rendered in the log pane, that reads as something the workload printed. It is
now recognised and answered as **410** with a plain sentence instead.

Matching on the prefix is safe here: logs are fetched with `Timestamps: true`,
so a genuine line always begins with an RFC3339 stamp and cannot collide. If
the kubelet ever rewords it we fall back to today's behaviour, no worse.

## 3. Red meant two different things

Red was doing duty for both "Radar failed" and "your pod is in a bad state". A
log file the node no longer holds is the second. Spending red on ordinary pod
states is how a reader learns to ignore it, so a 404 or 410 now renders neutral
and red keeps its meaning.

Also drops the live-stream notice while a previous run is on screen, where it
described something that was not happening.

## Tests

`packages/k8s-ui` 3802 pass, `web` lint clean, `go test ./internal/server` green,
new unit test for the sentinel match. Verified in the running app against a real
CrashLoopBackOff pod, and the Previous fix was confirmed by reverting it and
watching the wrong run come back.


hisco:

